PR #22, opened 2018-05-11
misc/log: print foo, not nil, in (log "foo") output

dotemacs:
This is really just resolving the issue https://github.com/clojure-emacs/nREPL/issues/18.
The patch was approved, but never merged. Packaging it here as a PR since it's low hanging fruit.

The patch was created by @rlbdv 2.5 years ago, but it never got applied. If you look at the issue linked above, it'll just take you to [this discussion on Clojure JIRA](https://dev.clojure.org/jira/browse/NREPL-84).

The original description of the patch:

> Previously (log "foo") would just convert any first argument that
> wasn't a Throwable to nil.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+* `clojure.tools.nrepl.misc/log` now logs all the arguments passed in,
+  without enforcing the conversion to Throwable even when the argument
+  is not an instance of Throwable.
+
 [`0.3.0`](https://github.com/clojure-emacs/nREPL/milestone/2?closed=1):
 
 * Materially identical to `[org.clojure/tools.nrepl "0.2.13"]`, but released under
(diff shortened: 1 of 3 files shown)


dotemacs:
I see that the build fails on TravisCI but it isn't this change, but rather the setup of the tests with `eastwood`...

bbatsov:
Are there any tests for the log function? Maybe you should add one covering the new behaviour?

bbatsov:
P.S. You should also update the changelog.

dotemacs:
Sure, will do, thanks

dotemacs:
Hi, I've added a test and updated the changelog.

In the test I've added this macro: 

https://github.com/dotemacs/nREPL/blob/196992ea9a2605629c9a5f6ed4edb0b649e7f9fa/src/test/clojure/clojure/tools/nrepl/misc_test.clj#L6-L16

just to capture the output of stderr. I know that macros are shunned usually, especially for such small use case, but this is almost the same as `with-out-str` only changed to that it acts on **\*err\*** instead. 

So should I leave it as is, or just wrap the `log` with the body of the macro in the test?

dotemacs:
Since the function in #23 re-implements the `log` function, this is now not relevant, so closing it.

PR #23, opened 2018-05-11
Linting fix

dotemacs:
Amended formatting of the code in the project in order to keep Eastwood happy. 
All of the `eastwood` warnings have been resolved.

All of the `cljfmt` issues resolved aswell.


dotemacs:
From 21 down to 7 warnings.

Will continue a bit later...

dotemacs:
There are only two `eastwood` warnings left:

```
* src/main/clojure/clojure/tools/nrepl/misc.clj:11:11: redefd-vars: Var log def'd 2 times at line:col locations: clojure/tools/nrepl/misc.clj:8:13 clojure/tools/nrepl/misc.clj:11:11
* src/main/clojure/clojure/tools/nrepl/misc.clj:8:13: bad-arglists: Function on var log defined taking # args [2 :or-more] but :arglists metadata has # args [0 :or-more]
```

Which is related to this function: https://github.com/clojure-emacs/nREPL/blob/380e161dae0f67c07645e2b4eb615797665ca989/src/main/clojure/clojure/tools/nrepl/misc.clj#L6-L17

From my conversation with the `eastwood` maintainers:

> So is there a way to say: in this namespace, disable warnings for :redefd-vars, in a configuration file ?

> The only way I can think of is to do a separate Eastwood run on that namespace, disabling warnings completely for the namespace.
> Then for your run on "most of your namespaces", exclude that namespace.
> So 2 separate Eastwood runs.
> Sorry, I know Eastwood could really use a facility for fine-grained disabling of individual warnings, like other lint tools have (e.g. via comments you add to the code with a special linter-recognized syntax), but never got around to implementing that.

Would you like to do that, have two configs:
* one for the whole project except `misc.clj`
* and another for just `misc.clj` ?

Or do you have some other suggestion?
Thanks

PR #24, opened 2018-05-17
ns clean up

dotemacs:
Went through all the ns declarations and tried to follow the
recommended best practices for ns requires, imports...


arrdem:
LGTM

bbatsov:
👍

Issue #15, opened 2017-11-06
Some messages to an nREPL server never result in a response that includes status :done


cemerick:
According to the documentation:

"Once a handler has completely processed a message, a response
containing a `:status` of `:done` must be sent."

I am assuming this includes messages that have resulted in an error. Some messages to an nREPL server never respond with such a status possibly leaving some clients waiting for follow up messages forever.

So far I have found two instances of this problem.

1. The interruptible-eval middleware may result in a :no-code error that does not include the :done status.

2. The session middleware may result in a :unknown-session error that does not include the :done status.

I'm including a patch for the above two which are just one word fixes. There might be other situations in which the problem arises.

(from https://dev.clojure.org/jira/browse/NREPL-76)

bbatsov:
Fixed by #25.

PR #25, opened 2018-05-17
Fixed some messages not resulting in status :done

dotemacs:
Fix for :status :done, issue https://github.com/nrepl/nREPL/issues/15

The fix for it was created over 3 years ago, I'm just applying it here.

The original patch was submitted by
@Dirklectisch on 30/Mar/15 10:26 AM

His description of the problem:

According to the documentation:

"Once a handler has completely processed a message, a response
containing a `:status` of `:done` must be sent."

I am assuming this includes messages that have resulted in an
error. Some messages to an nREPL server never respond with such a
status possibly leaving some clients waiting for follow up messages
forever.

So far I have found two instances of this problem.

1. The interruptible-eval middleware may result in a :no-code error
that does not include the :done status.

2. The session middleware may result in a :unknown-session error that
does not include the :done status.

I'm including a patch for the above two which are just one word
fixes. There might be other situations in which the problem arises.

### DIFF
--- a/src/clojure/clojure/tools/nrepl/middleware/interruptible_eval.clj
+++ b/src/clojure/clojure/tools/nrepl/middleware/interruptible_eval.clj
@@ -210,7 +210,7 @@
       (case op
         "eval"
         (if-not (:code msg)
-          (t/send transport (response-for msg :status #{:error :no-code}))
+          (t/send transport (response-for msg :status #{:error :no-code :done}))
           (queue-eval session executor
                       (fn []
                         (alter-meta! session assoc
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+* Bugfix for `clojure.tools.nrepl.middleware.session` for `:unknown-session`
+  error and `clojure.tools.nrepl.middleware.interruptible-eval` for `:no-code`
+  error, the correct response of `:status :done` is now being returned.
+
 [`0.3.0`](https://github.com/nrepl/nREPL/milestone/2?closed=1):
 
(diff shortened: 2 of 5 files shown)


src/clojure/clojure/tools/nrepl/middleware/interruptible_eval.clj
```diff
@@ -211,7 +211,7 @@
       (case op
         "eval"
         (if-not (:code msg)
-          (t/send transport (response-for msg :status #{:error :no-code}))
+          (t/send transport (response-for msg :status #{:error :no-code :done}))
```
bbatsov: Can we add a test for this change as well?

bbatsov:
The fix looks good to me, but it needs a changelog entry.

src/clojure/clojure/tools/nrepl/middleware/interruptible_eval.clj
dotemacs: Test added.

dotemacs:
Changelog entry added.

Thanks

Dirklectisch:
It's been three years! 👍

dotemacs:
I've taken the liberty of adding an unrelated change to this PR.

The badge in the README is showing as red due to the "incorrect" indentation.
So I took the liberty of adding it here:
https://github.com/nrepl/nREPL/pull/25/files#diff-1034d7bb623f15b45e70f92ae2f345bb
If this ruins the protocol of only sticking with the related changes into PRs, I'll omit this last commit.

bbatsov:
👍 

@Dirklectisch Now we're back in business and we'd love to get more contributors. We also accept pull requests finally! :-)

PR #34, opened 2018-06-14
Make sure that we assign zero when port is nil

arichiardi:
This is required when the `:port` key is passed in but the value is `nil`:

```clojure
cider-nrepl.main> (let [{:keys [port bind] :or {port 0}}
                        {:bind nil
                         :port nil}]
                    port)
nil
```


bbatsov:
This change should have some unit test.

arichiardi:
Agree, it would take some refactoring though as the options are not unit tested at all at the moment.

I tested in the REPL and by using `nrepl-server` from `cider-nrepl` with no parameters. It works. I still feel that you are right :smile:

bbatsov:
@arichiardi Tests are always the right thing to do.

arichiardi:
Totally but no additional time for the refactoring now :smile:

PR #36, opened 2018-07-23
Fix #16.

### LINKED ISSUE #16: Thread used to evaluate forms changes over time
See this thread for reference: https://groups.google.com/forum/#!topic/clojure/UfTQvYNf1-s

When a session is left idle for a little bit of time, the thread id evaluating forms changes. This causes any thread local variables in use to reset to their initial values. This causes strange behavior when doing interactive programming that involves use of thread locals.

The solution is for the same thread to always be used for a given session for as long as the session is open.

(from https://dev.clojure.org/jira/browse/NREPL-80)

blueberry:
I have tried this and it seems that now the main REPL thread stays the same while the futures started from it are being executed in the newly spawned threads from the pool. 

The change is trivial. Maybe in the future the whole function could be replaced by one of the Java's thread pool executor factory methods that does the same. It seems that one of many factory methods does this, but I opted to keeping the change minimal at first until we are sure that this is really the solution in more exotic cases...

### DIFF
--- a/src/clojure/nrepl/middleware/interruptible_eval.clj
+++ b/src/clojure/nrepl/middleware/interruptible_eval.clj
@@ -143,10 +143,6 @@
           (.setDaemon true)
           (.setContextClassLoader cl))))))
 
-(def ^{:private true} jdk6? (try
-                              (class? (Class/forName "java.util.ServiceLoader"))
-                              (catch ClassNotFoundException e false)))
-
 ;; this is essentially the same as Executors.newCachedThreadPool, except
 ;; for the JDK 5/6 fix described below
 (defn- configure-executor
@@ -159,7 +155,7 @@
   (let [^ThreadFactory thread-factory (or thread-factory (configure-thread-factory))]
     ;; ThreadPoolExecutor in JDK5 *will not run* submitted jobs if the core pool size is zero and
     ;; the queue has not yet rejected a job (see http://kirkwylie.blogspot.com/2008/10/java5-vs-java6-threadpoolexecutor.html)
-    (ThreadPoolExecutor. (if jdk6? 0 1) Integer/MAX_VALUE
+    (ThreadPoolExecutor. 1 Integer/MAX_VALUE
                          (long 30000) TimeUnit/MILLISECONDS
                          ^BlockingQueue queue
                          thread-factory)))
--- a/src/clojure/nrepl/middleware/interruptible_eval.clj
+++ b/src/clojure/nrepl/middleware/interruptible_eval.clj
@@ -143,18 +143,14 @@
           (.setDaemon true)
           (.setContextClassLoader cl))))))
 
-;; this is essentially the same as Executors.newCachedThreadPool, except
-;; for the JDK 5/6 fix described below
 (defn- configure-executor
   "Returns a ThreadPoolExecutor, configured (by default) to
-   have no core threads, use an unbounded queue, create only daemon threads,
+   have 1 core thread, use an unbounded queue, create only daemon threads,
    and allow unused threads to expire after 30s."
   [& {:keys [keep-alive queue thread-factory]
       :or {keep-alive 30000
            queue (SynchronousQueue.)}}]
   (let [^ThreadFactory thread-factory (or thread-factory (configure-thread-factory))]
-    ;; ThreadPoolExecutor in JDK5 *will not run* submitted jobs if the core pool size is zero and
-    ;; the queue has not yet rejected a job (see http://kirkwylie.blogspot.com/2008/10/java5-vs-java6-threadpoolexecutor.html)
     (ThreadPoolExecutor. 1 Integer/MAX_VALUE
                          (long 30000) TimeUnit/MILLISECONDS
                          ^BlockingQueue queue


bbatsov:
>  futures started from it are being executed in the newly spawned threads from the pool.

I don't get what futures have to do with this thread pool.

blueberry:
I have tried with two separate projects at the same time, and did not get any obstacles. Both projects work as intended.

I tried do play with futures because in the issue #16 discussion it was mentioned that nrepl wraps all evals into futures that are executed by the pool that we changed. Since the problem was that the actual thread from the pool was different after 30 sec inactivity, I wanted to check whether the spawned future from might "steal" the main thread for itself. Fortunately, it doesn't; so far, the main REPL thread stays the same...

bbatsov:
> I have tried with two separate projects at the same time, and did not get any obstacles. Both projects work as intended.

I meant several connections to the same nREPL server. 

> I tried do play with futures because in the issue #16 discussion it was mentioned that nrepl wraps all evals into futures that are executed by the pool that we changed.

Reading the code a bit I'm not sure that's actually the case. Seems originally nREPL used agents for the evaluation, but this proved problematic and the implementation was switched to mini-agents implementation based on this ThreadPool. 

The most relevant part of the implementation is probably:

```clojure
(defn- run-next
  [session executor f]
  #(try
     (f)
     (finally
       (run-next* session executor))))

(defn queue-eval
  "Queues the function for the given session."
  [session ^Executor executor f]
  (let [qa (-> session prep-session meta :queue)]
    (loop []
      (let [q @qa]
        (if-not (compare-and-set! qa q (conj q f))
          (recur)
          (when (empty? q)
            (.execute executor (run-next session executor f))))))))

(defn interruptible-eval
  "Evaluation middleware that supports interrupts.  Returns a handler that supports
   \"eval\" and \"interrupt\" :op-erations that delegates to the given handler
   otherwise."
  [h & configuration]
  (let [executor (:executor configuration @default-executor)]
    (fn [{:keys [op session interrupt-id id transport] :as msg}]
      (case op
        "eval"
        (if-not (:code msg)
          (t/send transport (response-for msg :status #{:error :no-code :done}))
          (queue-eval session executor
                      (fn []
                        (alter-meta! session assoc
                                     :thread (Thread/currentThread)
                                     :eval-msg msg)
                        (binding [*msg* msg]
                          (evaluate @session msg)
                          (t/send transport (response-for msg :status :done))
                          (alter-meta! session dissoc :thread :eval-msg)))))
```

bbatsov:
Basically it seems to me your change probably fixes things for 1 session as you end up with a thread that will never be killed, but what happens if you need more than 1?

bbatsov:
Or I might be misinterpreting the code and it's actually one Executor per session, which would mean your code is what we need.

blueberry:
Here is what happens with 2 concurrent connections to the same session:

The main thread is consistently the same in both sessions whenever there is no competition between them.

When there is, for example by calling Thread/sleep in one session, the other will be given a new thread for the time being (what else?) but as soon as the default thread is available, it will be preferred and used again.

blueberry:
No, read my last comment. The same executor is used in both sessions, but I think it still solves the problem in most cases, and is a good solution, since the trouble with changing threads mostly arises in interactive development, when the developer expects the session to be called in the same thread. If there is a single session, this is always true. If there are multiple sessions, it is not in cases when there is competing work, but I am not sure how realistic is that the developer will: 
1. Issue long-running evaluations from both threads.
2. Expect that they be in the same thread and not blocking

blueberry:
About your suggestion of a separate executor for each session: In that case, is the thread pool even needed? Each session should get one fixed thread for the main REPL flow and that's it?

blueberry:
Heh, now I misinterpreted you too :)

I have been connecting to the same session!

If i start two completely different sessions in two different terminals, then both get their own thread pool.

So, it seems that this is a good solution after all.

blueberry:
Hm, now, rereading the whole thread, I am not longer sure about the terminology. I am not sure what is considered session, and what is connection, where those are referring to the same thing, and when they are not.

So, as a clarifying point, in my last message, I am referring to the session as a thing you get by starting `lein nrepl` from a separate terminal. Connection is what I get when i run `M-x cider-connect` and choose the session localhost:id...

bbatsov:
> Hm, now, rereading the whole thread, I am not longer sure about the terminology. I am not sure what is considered session, and what is connection, where those are referring to the same thing, and when they are not.

Yeah, this should be documented somewhere, once we're certain about it that is. Here's how I understand it:

* one connection is mapped to one session by default (session in the sense of a session middleware session, basically the same REPL state with different thread bindings). Requests get dispatched to sessions and this allows you to have different REPLs attached to each session (e.g. one for user evaluation and one evaluating things like completion code or whatever). Sessions can be created with the `clone` op, I'm not sure whether if you open a few different connections to the same server this creates multiple sessions (I assume it does, but this has to be checked). 
* in each session you have one evaluation thread available, which means you can't evaluate things concurrently in the same session. 

Here's a bit of info on the subject in the manual http://nrepl.readthedocs.io/en/latest/design/#sessions

There's a lot of room for improving the "Design" section.

@SevereOverfl0w and @vspinu can probably contribute to the explanation of sessions more, as they played a lot with them. None of us was part of the creation of nREPL and we're all discovering things as we go. I try to create good documentation when this happens, so things would be happening easier down the road.

bbatsov:
> The same executor is used in both sessions, but I think it still solves the problem in most cases, and is a good solution, since the trouble with changing threads mostly arises in interactive development, when the developer expects the session to be called in the same thread. If there is a single session, this is always true.

In light of all of what I wrote above it seems to me that we're aiming for a situation where we know a session is going to be mapped to the same thread for its entire existence. My assumption about the executor logic was that it was providing threads per sessions, not per individual evaluations. (but I might be wrong)

blueberry:
In light of your last messages, I think we have a rather logical, expected, and flexible behavior now:

1. When the session is started and the first connection to it made, the session *does* get one permanent thread for evaluation. It is mapped to this same permanent thread for its entire existence.
2. When another connection gets attached to the same session, its evaluation do happen in the default permanent thread, *unless* that thread is busy, in which case new thread is being created.
3. Executor thread provides threads per individual evaluations, but uses the pool, in which there is always at least default thread already available, which is used prefferably.
4. As it is now, the only case where the default session thread is not used for evaluation is when there is already evaluation running in the same session. I think this is a) rare b) expected, since the alternative is that all connections to the same session block and wait until the first finish with its evaluation.  

I agree that this can be designed in a different way, without thread pools, or with blocking, or with a pool per connection, and in many other (better?) ways. However, all these things require much more time and understanding. It might take months until they land in the code. That's why I propose this solution, with no downsides and the following benefits:

1. It solves the wandering threads for most users all of the time, and for all users most of the time.
2. It does not surprise users, most won't even notice that something has changed.
3. Existing tools continue to work since there are no architectural changes.
4. (most importantly) It is available today. When the time comes for architectural changes that solve this in a more encompassing way, even better! :)

bbatsov:
I'm not trying to argue against your solution, I'm just trying to understand the code we have today well enough to make sure we're not making anything stupid. I'm all for a simple solution, unfortunately that part of the codebase is a bit convoluted. 

Looking closer at the code it seems you me we are right.

```clojure
(defn interruptible-eval
  "Evaluation middleware that supports interrupts.  Returns a handler that supports
   \"eval\" and \"interrupt\" :op-erations that delegates to the given handler
   otherwise."
  [h & configuration]
  (let [executor (:executor configuration @default-executor)]
    (fn [{:keys [op session interrupt-id id transport] :as msg}]
      (case op
```

I don't ever see an usage of this with some configuration passed, so it seems we always have the same executor for all the eval ops, which would mean that all sessions share this thread unless extra evaluations are happening. The tread should be relatively stable across sessions then. All of the code in this file can benefit from some documentation, though.

bbatsov:
And I just found this in the changelog:

```
`0.2.9`:

* `clojure.tools.nrepl.middleware.interruptible-eval` now defines a default
  thread executor used for all evaluations (unless a different executor is
  provided to the configuration of
  `clojure.tools.nrepl.middleware.interruptible-eval/interruptible-eval`). This
  should aid in the development of `interrupt`-capable alternative evaluation
  middlewares/handlers.
```

blueberry:
@vspinu That might be true (or not, I don't know enough about repl servers to judge) but all these problems exist even without this patch. The only change is that now there is one pre-created thread that will be reused by default, while before it was the same, provided that the REPL was active every 30 seconds or less.

bbatsov:
Yeah, I agree that this change is step forward. Can't imagine what can go worse than it used to be - threads dying after each 30 seconds of inactivity are certainly not fun either.

Let's keep thinking about the best long term solution, though.

vspinu:
Sure, but the original issue is not solved and shouldn't be closed.

PR #43, opened 2018-08-18
Remove extra newline in output

### LINKED ISSUE #38: The interactive repl needs to truncate the last newline if it is using println
The interactive REPL output needs to truncate the last newline of any output string if its supplying its own.

samihda:
`:out` is now printed as it is, purely based on the user print statement (i.e., as opposed to `:value`, where a newline is added for convenience). AFAICT, the `:error` string also comes with a newline character in the end, so I changed that too.

This should fix #38.


bbatsov:
Thanks!

PR #46, opened 2018-09-01
server: add support for message authentication/authorization

rlbdv:


rlbdv:
I could easily imagine the current approach might not be acceptable, but I'd be happy to make adjustments.  Please see the commit message for additional explanation and rationale.

I haven't added any tests yet, but "lein test" should still work, if a bit noisily, and the noise can be suppressed:

    NREPL_SUPPRESS_DEFAULT_AUTH_WARNING=true lein test

In any case, assuming we can settle on an initial approach, I'll be happy to add tests and other desired enhancements.

rlbdv:
Oh, and I'd imagine before we're finished, this should include some updates to the manual.

bbatsov:
I'll add a bit of feedback inline and I'll invite @nrepl/nrepl members to share their opinions as well.

src/clojure/nrepl/server.clj
```diff
@@ -64,6 +64,31 @@
                           s))
                       % %))))
 
+(defn no-auth-wrapper
```
bbatsov: I'm not sure I understand the purpose of this wrapper. I guess that if we don't want to use authorization we can simply pass `nil` or something along those lines.

src/clojure/nrepl/server.clj
```diff
@@ -64,6 +64,31 @@
...
+  NREPL_SUPPRESS_DEFAULT_AUTH_WARNING is set to true."
+  [handle-message]
+  (when-not (some-> (System/getenv "NREPL_SUPPRESS_DEFAULT_AUTH_WARNING")
+                    (= "true"))
+    (binding [*out* *err*]
+      (println "nREPL set to require no authorization; allowing all access")
+      (println "Suppress warning with NREPL_SUPPRESS_DEFAULT_AUTH_WARNING=true")))
+  handle-message)
+
+(defn file-auth-wrapper
```
bbatsov: That's named `file-auth-wrapper`, but I don't see it doing anything with files.

src/clojure/nrepl/server.clj
rlbdv: Assuming I understand, that'd be fine too.  Were you saying that you just wanted to special case the "no auth" case at the point of use, i.e. put this code there guarded by a check whether the default-auth-wrapper is set, or are you saying that you don't want to warn at all when there's no protection?

src/clojure/nrepl/server.clj
rlbdv: Oh, right, that's vestigial -- I'll fix it in the next revision.

src/clojure/nrepl/server.clj
```diff
@@ -107,8 +132,9 @@
    * :port — defaults to 0, which autoselects an open port
    * :bind — bind address, by default \"::\" (falling back to \"localhost\"
        if \"::\" isn't resolved by the underlying network stack)
+   * :auth-wrapper - defaults to the result of `(default-auth-wrapper)`
```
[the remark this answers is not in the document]
rlbdv: Sounds reasonable -- I can investigate.  If it's not straightforward, is that something you'd want in the first iteration, or is it possible that could be added later?  (I agree that if it's feasible, it'd be better to have it now.)

src/clojure/nrepl/server.clj
```diff
@@ -120,8 +146,9 @@
    either via `stop-server`, (.close server), or automatically via `with-open`.
    The port that the server is open on is available in the :port slot of the
    server map (useful if the :port option is 0 or was left unspecified."
-  [& {:keys [port bind transport-fn handler ack-port greeting-fn]}]
+  [& {:keys [port bind transport-fn auth-wrapper handler ack-port greeting-fn]}]
   (let [port (or port 0)
+        auth-wrapper (or auth-wrapper default-auth-wrapper)
```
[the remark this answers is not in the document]
rlbdv: OK, perhaps that answers my other question above.  And while it may not have to be addressed in tools.nrepl.  I definitely think that the ecosystem should be secure by default, and so it should eventually require some effort to end up with a backdoor that bypasses all local authentication.  At a minimum that might suggest ensuring that eventually all "front ends", i.e. leiningen, cider, the nrepl command line, etc. default to some kind of authentication.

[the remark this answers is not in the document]
rlbdv: I wonder if it might be plausible to start with no changes to the defaults and/or external behavior (as you suggest), but add something to the README recommending authentication, and mention the desire to eventually change the default.  Then at some point later, we could start warning to *err* about the impending change of the default (under the assumption that most consumers won't be parsing *err*, and won't be disrupted, but people may see it), and then finally, after that's been the norm for a while, change the default, along with an appropriate version change -- i.e. perhaps as a "breaking" change.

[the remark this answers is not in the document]
rlbdv: With respect to middleware -- assuming I understand your comments correctly, I treated the authentication independently under the supposition that authentication is somewhat different.  i.e. it's something that should "go first", so that nothing else can be added before it that accidentally or intentionally acts on incoming, unauthenticated data.  Treating it independently was also a simplification, trying to make it easier for someone to convince themselves that they (or someone else) hadn't accidentally or intentionally, broken or bypassed intended restrictions.

Along those same lines, I was (rightly or wrongly) thinking about the decision to accept a message as potentially independent of the decision about what to do with it once it's been accepted (a la ssh/pam, etc.).

I suppose the alternative might be to just make authentication a normal wrapper, and add loud documentation about always "tacking it back on to the front" if you use your own middleware, or if you add your own wrappers in front of the default wrapper.  (Of course in the latter case, you'd end up authenticating the same message multiple times.)

And one reason I'd provided the default-auth-wrapper (even if the default were identity for now) is so that people could easily wrap the default handler with their own additional policies.

bbatsov:
> With respect to middleware -- assuming I understand your comments correctly, I treated the authentication independently under the supposition that authentication is somewhat different. i.e. it's something that should "go first", so that nothing else can be added before it that accidentally or intentionally acts on incoming, unauthenticated data. Treating it independently was also a simplification, trying to make it easier for someone to convince themselves that they (or someone else) hadn't accidentally or intentionally, broken or bypassed intended restrictions.

Yeah, I see this point. Generally I think that in practice that's unlikely to be a problem as you can obviously intercept any message in an authentication middleware, so unless someone explicitly disables it we're in business. But I'm not insisting on a middleware approach. I just seems to me in line with everything we've done some far (e.g. even sessions are middleware and they could have easily been something special). 

> Along those same lines, I was (rightly or wrongly) thinking about the decision to accept a message as potentially independent of the decision about what to do with it once it's been accepted (a la ssh/pam, etc.).

You'll have to elaborate on this a bit. We still have to authenticate every message regardless of the approach taken, so I'm a bit puzzled by this comment. 

> I suppose the alternative might be to just make authentication a normal wrapper, and add loud documentation about always "tacking it back on to the front" if you use your own middleware, or if you add your own wrappers in front of the default wrapper. (Of course in the latter case, you'd end up authenticating the same message multiple times.)

For the middleware approach this seems perfectly fine to me. If we go with the wrapper approach probably we can just make this some var you can pass on server-init.  I'm thinking we can also have a configurable middleware that could support different auth schemes, but that would probably be an overkill initially. 

Long story short - I try to be really careful with the approach we take, so that we're certain we select something that's going to be flexible, get the job done and relatively consistent with how we've operated so far.

src/clojure/nrepl/server.clj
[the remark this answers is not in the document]
rlbdv: If i pursue it, what were you contemplating here?  I could see doing something like what's done for `--handler`/`--middleware` (depending on how we resolve the middleware-related questions), but that still leaves the token.  I assume we'll need to read it from an environment variable, file, stdin, or similar so that it's not visible to others on the system (via `ps -au`, etc.), which might suggest adding a custom auth handler to cmdline.

I suppose we could also consider defining a "standard", e.g. to look for `~/.nrepl-auth-token` (or whatever) for a candidate token, as a default.

rlbdv:
> You'll have to elaborate on this a bit. We still have to authenticate every message regardless of the approach taken, so I'm a bit puzzled by this comment.

I just meant that I did this:

    (auth-wrapper (or handler (default-handler)))

so that it was really clear that authentication happens before anything else, and you can't side-step it.  

Whether or not that's what we want, is of course, another question, but that's what I was thinking at the time.  (The vague (possibly inappropriate) ssh analog being that you can't run any commands until you've been allowed access, perhaps passed the whole pam stack, etc.).

And with respect to your other suggestions, I have a suspicion I don't completely understand the arrangement you're contemplating.  I was wondering if you might be able to sketch it out a bit more.  i.e. What would the approach(es) you've been thinking of look like?

Thanks

bbatsov:
> so that it was really clear that authentication happens before anything else, and you can't side-step it.

I see. 

> And with respect to your other suggestions, I have a suspicion I don't completely understand the arrangement you're contemplating. I was wondering if you might be able to sketch it out a bit more. i.e. What would the approach(es) you've been thinking of look like?

What I was thinking about was:

* you've got auth middleware
* the behaviour of the middleware is configured on server startup by passing to the server some var (implementation of an auth schema)
* When using `clj` you can also pass this var (or something along those lines) via the command line interface
* This middleware positions itself before everything in the middleware stack and checks each message for the necessary auth data.

I think that's pretty similar to what you're proposing with the slight difference it's a middleware instead of something hardcoded in the server. Once again - I'm not 100% sure which approach would be better. At least superficially your idea seems to be a bit simpler, but on the other hand - not very consistent with with how most functionality has been implemented so far.

src/clojure/nrepl/server.clj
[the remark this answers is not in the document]
bbatsov: For this type of authentication yes. On the server side everything's easy with respect to auth token and auth schemes. The real question is how to teach clients about different schemes. Guess the idea is to just hardcode the support for whatever auth schemes are known in the various clients.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> For this type of authentication yes. On the server side everything's
> easy with respect to auth token and auth schemes. The real question is
> how to teach clients about different schemes. Guess the idea is to
> just hardcode the support for whatever auth schemes are known in the
> various clients.

Hmm, I'd vaguely imagined that past a certain point, tools.nrepl might
not necessarily be involved.  i.e. the critical thing is that
tools.nrepl provide a hook of some kind, and once that's available (in
whatever form), then leiningen, cider, et. al. might decide on a
specification like:

  - messages must have an :auth token
  - check NREPL_AUTH_TOKEN, in the environment; if set, use it
  - check NREPL_AUTH_TOKEN_FILE in the environment; if set, use it
  - otherwise look for PROJECT_ROOT/.nrepl-auth-token
  - otherwise look for ~/.nrepl-auth-token
  - give up.

And so while tools.nrepl *could* be involved and might provide a
built-in function implementing the "lein/cider/whoever strategy", it
needn't be, and/or can wait until they've stabilized their
specification before adopting anything officially.

So I'd planned to badger technomancy (and the cider devs) next, once we
had the hook, but I'm happy to proceed in whatever way seems best, and
it'd also be perfectly fine if tools.nrepl does want to provide
something like the above "out of the box".

If the latter, then I'd be happy to double-check with leiningen and
cider about any proposed semantics.

Oh, and I could also imagine a path where tools.nrepl eventually *only*
provides the hook, and makes the :auth-wrapper/whatever argument
required, so that you have to specify a strategy, even if you just say
"identity".  That way the decision is explicit, and obvious.

Thanks again
-- 
Rob Browning
rlb @defaultvalue.org and @debian.org
GPG as of 2011-07-10 E6A9 DA3C C9FD 1FF8 C676 D2C4 C0F0 39E9 ED1B 597A
GPG as of 2002-11-03 14DD 432F AE39 534D B592 F9A0 25C8 D377 8C7E 73A4

bbatsov:
> And so while tools.nrepl *could* be involved

Btw, it's just nREPL (`nrepl.`) now. The `tools.` part was dropped when we decided to leave Clojure Contrib. Normally we refer to `tools.nrepl` as the legacy nREPL at this point. 

> So I'd planned to badger technomancy (and the cider devs)

I guess you don't know that I'm the author of CIDER. 😉 That's why from the very beginning I've been thinking about this end-to-end. And I happen to be quite familiar with most of the other nREPL clients out there. :-)

rlbdv:
> I guess you don't know that I'm the author of CIDER. wink That's why from the very beginning I've been thinking about this end-to-end. And I happen to be quite familiar with most of the other nREPL clients out there. :-)

Hah, no I didn't.  And well, that's certainly convenient.

So backing up a bit.  I'll be ecstatic if we just make it *possible* for people to enable some form of solid authentication for lein/cider/etc.  Anything more I'd just consider a bonus -- and with respect to a lot of the details, "if you're happy, I'm happy".

Given that, what would you like this first version to look like?  Would it resolve most of the questions if I were to change the auth implementation to be a middleware, assuming I understand what we mean by that, and then provide a default auth-middleware stack that's empty, along with a `nrepl-auth-token-wrapper` (or whatever) that people can include if they want to.  And perhaps that wrapper would implement something like scheme described here?  https://github.com/nrepl/nREPL/pull/46#issuecomment-421841308

Then, once we're happy with that, we can perhaps ping technomancy, if he's not already paying attention...

bbatsov:
> Given that, what would you like this first version to look like? Would it resolve most of the questions if I were to change the auth implementation to be a middleware, assuming I understand what we mean by that, and then provide a default auth-middleware stack that's empty, along with a nrepl-auth-token-wrapper (or whatever) that people can include if they want to. And perhaps that wrapper would implement something like scheme described here? #46 (comment)

Yep, that'd be a good start. I'm guessing that with the middleware approach we can have some param as part of the server init that controls wether the middleware does something or is an no-op. 

> Then, once we're happy with that, we can perhaps ping technomancy, if he's not already paying attention...

Sounds like a plan to me. There's also my small https://github.com/nrepl/lein-nrepl where we can add support for this before it makes it to upstream leiningen. Changes there certainly take a while to happen, so this might be a good stop-gap measure.

rlbdv:
Regarding the middleware approach, were you just thinking that we might start with

    (defn default-auth-middlewares []
      [])

then maybe later change the default to something like:

    (defn default-auth-middlewares []
      [#'nrepl.middleware.auth/require-nrepl-auth-token])

or were you imagining the more complicated arrangement we have on the default-middlewares side, i.e. linearization, etc.?  (If the latter, I'm a bit wary of any more complexity than we really have to have when auth is involved.)

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> Yeah, your suggestion makes sense to me. I'm assuming you plan to make
> the auth-middlewares run before every other middleware, right?

Yep, they'd always go first.

> Btw, it might be good to use `authentication` instead of `auth` in the
> code, as some people also brought up the issue of authorization and
> `auth` can be interpreted as referring to both.

Hah, I'd also been pondering that -- I'll change it.  Would you also be
inclined toward `NREPL_AUTHORIZATION_TOKEN` rather than
`NREPL_AUTH_TOKEN`, etc., and/or an :authorization key in the message
rather than :auth?

bbatsov:
Well, seems to me that'd be better - especially given the fact that i called it `authentication` and you called it `authorization`. :D I just assumed that we'd consider the token just an authentication mechanism and if you are authenticated (have the right token) you're authorized to do anything, but I guess your perspective on this is different - the presence of a token authorizes you to do anything.

bbatsov:
I think blanket authorization will be fine, so let's not overthink this. After all any form of authorization is going to be infinitely better than none. :-) In an ideal world it would have been nice to be able to optionally tie auth tokens to the ability to execute only certain middleware ops, but that might be a bit excessive, so let's go with something simple.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> I think blanket authorization will be fine, so let's not overthink
> this. After all any form of authorization is going to be infinitely
> better than none. :-) In an ideal world it would have been nice to be
> able to optionally tie auth tokens to the ability to execute only
> certain middleware ops, but that might be a bit excessive, so let's go
> with something simple.

Sounds good.  I wondered about keeping the shorter :auth for the key
since it'll be in every message, but probably not significant either
way.  I'm fine with whatever you prefer.

Also, right now I have an additional :explanation key for errors.  Is
that reasonable, or should it be something else?

      (t/send transport
              (response-for msg :status
                            #{:error :unauthorized :done}
                            :explanation "invalid :auth token")))))

(And with respect to the future -- wouldn't be hard to adapt later,
 regardless -- even if that required the addition of some kind of
 message versioning for some change in semantics we didn't want to work
 around some other way, i.e. ":v 1" messages, etc.)

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

>> Sounds good.  I wondered about keeping the shorter :auth for the key
> since it'll be in every message, but probably not significant either
> way.  I'm fine with whatever you prefer.
>
> I'm fine with the shorter name as well, provided the middleware is well documented. 

OK, so sounds like you prefer :auth in the message, and authorization
everywhere else, i.e.:

    (defn find-token
      []
      (or (System/getenv "NREPL_AUTHORIZATION_TOKEN")
          (some-> (System/getenv "NREPL_AUTHORIZATION_TOKEN_FILE") slurp)
          (some-> (try
                    (slurp (File. (System/getProperty "user.home") ".nrepl-authorization-token"))
                    (catch FileNotFoundException ex nil)))))

> We always put errors in the status set, so I guess you can have there
> `:unauthorized :no-auth-token`, `:unauthorized :invalid-auth-token`.

Ahh, ok.  I'll rearrange that.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> Yeah, your suggestion makes sense to me. I'm assuming you plan to make
> the auth-middlewares run before every other middleware, right?

I just realized that maybe I didn't quite understand what you preferred
here, but I was somewhat thinking about ring's wrappers, and here's what
I currently have (which may or may not be quite what we want):

    (defn default-authorization-wrapper
      "Returns identity (i.e. no access restriction), matching previous
      nREPL behavior, but is expected to eventually return
      nrepl.middleware.auth/require-nrepl-authorization-token."
      []
      identity)
...

    (defn start-server
      "Starts a socket-based nREPL server.  Configuration options include:

...

       * :authorization-wrapper - defaults to `(default-authorization-wrapper)`
           which currently does nothing;
           nrepl.middleware.auth/require-nrepl-auth-token is recommended
           when the client supports it.
       * :handler — the nREPL message handler to use for each incoming connection;
           only receives messages permitted by the :authorization-wrapper;
           defaults to `(default-handler)`.

...

        authorization-wrapper (or authorization-wrapper (default-authorization-wrapper))
        handler (or handler (default-handler))

...

        server (Server. ss
                        (.getLocalPort ss)
                        (atom #{})
                        (or transport-fn t/bencode)
                        greeting-fn
                        (authorization-wrapper handler))]

So the caller can specify their own wrapper, use the default one, or
wrap the default one before passing it in, and the authorization-wrapper
always "goes first".

bbatsov:
That seems reasonable to me.

rlbdv:
How is the existing `--handler` argument typically used from the command line, i.e. what's an sample argument value?  (Does it rely on reader evaluation?)

rlbdv:
Would you prefer for the default token file to be ~/.nrepl-authorization-token, an XDG style ~/.config/nrepl/authorization-token, or something else?

bbatsov:
> How is the existing --handler argument typically used from the command line, i.e. what's an sample argument value? (Does it rely on reader evaluation?)

It's a string referring to a symbol/var, that's read (e.g. `cider.nrepl/cider-nrepl-handler`). You either use it or the `--middleware` flag depending on how granular you want to be. 

> Would you prefer for the default token file to be ~/.nrepl-authorization-token, an XDG style ~/.config/nrepl/authorization-token, or something else?

That's a good question. I wonder if it won't best to have some `.nrepl` (or something like this) file which is some edn and allows you to specify different configuration keys (not just the the auth token). Then I guess it'd be fine for this file to be in the home folder.

bbatsov:
Or `.nrepl/config.edn`. Don't have a strong preference for this. It should definitely be possible to instruct the server to look for whatever config file we opt to use by default elsewhere, so people can tweak this if they want to.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

>> How is the existing --handler argument typically used from the command line, i.e. what's an sample argument value? (Does it rely on reader evaluation?)
>

> It's a string referring to a symbol/var, that's read
> (e.g. `cider.nrepl/cider-nrepl-handler`). You either use it or the
> `--middleware` flag depending on how granular you want to be.

Hmm, after mimicking the existing cmdline --handler code, the only way
I've figured out as yet to get the --authorization-wrapper value to work
is via `#=`, i.e. reader evaluation:

    java ... \
      --authorization-wrapper "#=(var nrepl.middleware.auth/require-nrepl-authorization-token)"
      ...

Perhaps I overlooked something more obvious?

> That's a good question. I wonder if it won't best to have some
> `.nrepl` (or something like this) file which is some edn and allows
> you to specify different configuration keys (not just the the auth
> token). Then I guess it'd be fine for this file to be in the home
> folder.

Of course, one potential advantage to a config subdir like
.config/nrepl/, .nrepl/, or whatever is that you can easily put single
values in separate files, which means that it's trival to read shared
values like the token from any language -- no config parser necessary.

bbatsov:
> Of course, one potential advantage to a config subdir like
.config/nrepl/, .nrepl/, or whatever is that you can easily put single
values in separate files, which means that it's trival to read shared
values like the token from any language -- no config parser necessary.

Yeah, that's a pretty good point. OK, let's go with `.nrepl/` and separate files for now. We can call this "experimental" and tweak it a bit down the road. 

> Don't know if that's what we'll want in the end, but it does make it
possible to specify an alternate location for the token file.

I guess that's a reasonable start. No need to get everything perfect from the start. 

>       --authorization-wrapper "#=(var nrepl.middleware.auth/require-nrepl-authorization-token)"

Why the reader macro reader here? Can't you just pass a symbol and resolve it at runtime?

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> Why the reader macro reader here? Can't you just pass a symbol and
> resolve it at runtime?

Well it looked like the code does this (effectively):

    ((read-string VALUE))

And without some explicit eval like this

    ((-> VALUE read-string eval))

I don't yet see how the existing code could work without the #= (since
read-string just returns 'foo for "foo" and '(var bar) for #'bar, and
we can't call those directly).  For example, this crashes in the repl:

    ((-> "#'clojure.string/join" read-string) "-" ["x" "y"])

The #= reader macro sneaks the eval into the read-string, which is of
course why read-string and the related functions are "unsafe" in many
contexts, and avoided in favor of their edn counterparts:
edn/read-string, etc.  But if the existing code does work without the
#=, then I must be misunderstanding something.

Otherwise, assuming we want these options to allow execution of
arbitrary code (which might be plausible, and what's intended, under the
assumption that you could always just add your own jar to the
classpath), I suspect we may need to add an explicit eval there.

Alternately we could be more restrictive, and only allow variable
references:

    (let [split-sym (juxt #(some-> % namespace 'some-default-ns) name)]
      (->> VALUE
           edn/read-string
           symbol
           split-sym
           (map symbol)
           (apply ns-resolve)))
                
Which should allow 

  --handler foo --handler foo/bar

resolving to #'some-defaultns/foo and #'foo/bar, but not anything else.

rlbdv:
OK, I've pushed  a new version that should incorporate much of what we've discussed.  I imagine there are some controversial bits, but I wanted to see *which* bits before I got too much further down a wrong path.  (For example, the tests read the port from a new message to *err* -- won't be surprised if that's not what you'd prefer.)

bbatsov:
> Alternately we could be more restrictive, and only allow variable
references:

I guess that'd be fine for the hander and the auth-wrapper. I was less restrictive with the middleware because it can be a var that resolves to a list of middleware vars or a list of symbols we resolve as vars. 
Tweaking this in the cmdline code should be trivial. 

> OK, I've pushed a new version that should incorporate much of what we've discussed. I imagine there are some controversial bits, but I wanted to see which bits before I got too much further down a wrong path. (For example, the tests read the port from a new message to err -- won't be surprised if that's not what you'd prefer.)

Overall the implementation looks good. The tests definitely seem more complex than I'd like them to be. I'll add some inline comments here and there when I can.

Btw, a general note on security - I see the current implementation just compares the token directly which seems unwise. Shouldn't we just send some hashed version of it over the network in the messages, so it would not be compromised easily?

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> I guess that'd be fine for the hander and the auth-wrapper. I was less
> restrictive with the middleware because it can be a var that resolves
> to a list of middleware vars or a list of symbols we resolve as vars.
> Tweaking this in the cmdline code should be trivial.

OK, *if* we don't want the handling to be fully general, then I'm sure
we can arrange to limit each arg to whatever cases we like, i.e. to just
what you describe above.

On the other hand, if eval ends up being what's wanted, perhaps we just
do that for all the relevant args, and make sure it's quite clear in the
docs, so people can plan accordingly.

> Overall the implementation looks good. The tests definitely seem more
> complex than I'd like them to be. I'll add some inline comments here
> and there when I can.

Sounds good.  I'd thought about simplifying them further, but wanted to
wait until I knew your preferences better.

Of course they're somewhat complicated by trying to ensure they don't
expose the token or an unrestricted "eval port" during the tests by
default (keeping the token file and "home" private without fs races,
etc.), and because they launch a new server each time to directly test
the handling of the environment variables, file loading, etc. by the
actual process.

> Btw, a general note on security - I see the current implementation
> just compares the token directly which seems unwise. Shouldn't we just
> send some hashed version of it over the network in the messages, so it
> would not be compromised easily?

Hmm, well I wasn't intending this for use across an insecure connection,
and even locally, I was thinking of it as just a secure way to "talk to
yourself" given the jvm's long-standing ignorance of unix domain
sockets.  (If we could easily use those, that's definitely what I would
have proposed instead.)

And I'm probably missing something, but I'm not sure I completely
understand the threat you're contemplating here yet.  If we hash the
token, then both sides (server and client) would have to store the hash
somewhere, and then that's just the new token?

But we can certainly try to do something more sophisticated.  However,
past a very near point, I'd really want the assistance/review of a
security expert (i.e. not me), and I'd also want to have a solid
understanding of our threat model -- i.e. what are we trying to protect
against? local .nrepl/authorization-token compromise, replay attacks,
...

bbatsov:
> OK, *if* we don't want the handling to be fully general, then I'm sure
we can arrange to limit each arg to whatever cases we like, i.e. to just
what you describe above.

I'm pretty certain we don't need this to be general. I simply didn't think that much when I wrote the current code. It was quick and it got the job done.

> Hmm, well I wasn't intending this for use across an insecure connection,
and even locally, I was thinking of it as just a secure way to "talk to
yourself" given the jvm's long-standing ignorance of unix domain
sockets.  (If we could easily use those, that's definitely what I would
have proposed instead.)

Ah, sorry. I always thought you wanted to secure remote connections. Then you can forget what I said.
So, ultimately what you want to solve is not sending some messages locally to the wrong server, right? 

Not that I said anything meaningful. Once I woke up I realized for the millionth time I should be responding to any message late in the evening when I can't think. :D 

> But we can certainly try to do something more sophisticated.  However,
past a very near point, I'd really want the assistance/review of a
security expert (i.e. not me), and I'd also want to have a solid
understanding of our threat model -- i.e. what are we trying to protect
against? local .nrepl/authorization-token compromise, replay attacks,

Yeah, yeah. Just forget about all of this. I was mostly concerned that now it'd be easy to just obtain the token over an unencrypted connection, but of course we'd need to encrypt the entire nREPL message to be able to protect ourselves from something like this and I don't want us to go overboard with this.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> I'm pretty certain we don't need this to be general. I simply didn't
> think that much when I wrote the current code. It was quick and it got
> the job done.

Ahh, ok, well, I'm happy to propose something in a separate commit
there.  So for middleware, the options are basically:

    some.thing/foo
    [some.thing/foo some.thing/bar]

And do we want to allow both `#'foo` and `foo`, on the command line or
just `foo` (that actually resolves to the var)?

> Ah, sorry. I always thought you wanted to secure remote
> connections. Then you can forget what I said.  So, ultimately what you
> want to solve is not sending some messages locally to the wrong
> server, right?

Ahh, right -- well, while I'm all in favor of solid, built-in (or as a
trivial add-on) fancier remote authentication/authorization, the
baseline for me, for now, was to try to just make it possible to prevent
random local users from submitting commands.  That coupled with a
default that only binds to localhost should make the most common use
cases safe(er).

...and then eventually, I'd love to make that the default.

> Not that I said anything meaningful. Once I woke up I realized for the
> millionth time I should be responding to any message late in the
> evening when I can't think. :D

Hah, I'm pretty sure there are quite a lot of us in that club.

> Yeah, yeah. Just forget about all of this. I was mostly concerned that
> now it'd be easy to just obtain the token over an unencrypted
> connection, but of course we'd need to encrypt the entire nREPL
> message to be able to protect ourselves from something like this and I
> don't want us to go overboard with this.

Sounds good -- I think we can always pursue more sophisticated options
(that better support some remote arrangements) later.

Thanks again

rlbdv:
OK, I've added `resolve-auth-wrapper-or-die`, a few supporting functions, and adjusted the code to use it.  Does what I've done seem plausible?  Among other things I thought we might want the more descriptive error messages.  I also arranged to flush `*out*` and `*err*` and shutdown-agents on exit (so we won't risk hanging -- assuming that's still a clojure concern).

I could add a commit to extend the use of the exit wrapper to the other System/exit calls if the approach doesn't seem unreasonable.

Note that this implementation doesn't support the '#'some/thing' syntax, only `some/thing` for the `--authorization-wrapper` value because it uses `edn/read-string`, but `some/thing` will still resolve to a var.

If that's preferable, we could consider adjusting `--middleware` to work the same way, but it *would* break backward compatibliity.  If that's not preferable, then I think we may be able to just bind `*read-eval*` to false when calling read-string.

rlbdv:
Rob Browning <notifications@github.com> writes:

> I could add a commit to extend the use of the exit wrapper to the
> other System/exit calls if the approach doesn't seem unreasonable.
>
Oh, and I could also see just having a top-level {:kind ::exit ::status
...} catch that handled the clean up and exit (instead of the scattered
exit calls), but didn't know which you'd prefer, or if you'd want that
much change right now.

bbatsov:
> OK, I've added resolve-auth-wrapper-or-die, a few supporting functions, and adjusted the code to use it. Does what I've done seem plausible? Among other things I thought we might want the more descriptive error messages. I also arranged to flush *out* and *err* and shutdown-agents on exit (so we won't risk hanging -- assuming that's still a clojure concern).

Looks good to me. 

> I could add a commit to extend the use of the exit wrapper to the other System/exit calls if the approach doesn't seem unreasonable.

👍 

> Note that this implementation doesn't support the '#'some/thing' syntax, only some/thing for the --authorization-wrapper value because it uses edn/read-string, but some/thing will still resolve to a var.

That's fine. 

> If that's preferable, we could consider adjusting --middleware to work the same way, but it would break backward compatibliity. If that's not preferable, then I think we may be able to just bind *read-eval* to false when calling read-string.

Yeah, absolutely. I really doubt anyone's evaluating code there, so that should not be a big deal. 

> Oh, and I could also see just having a top-level {:kind ::exit ::status
...} catch that handled the clean up and exit (instead of the scattered
exit calls), but didn't know which you'd prefer, or if you'd want that
much change right now.

Please, expand on this, as I'm not sure I understand your idea.

src/clojure/nrepl/cmdline.clj
```diff
@@ -3,11 +3,25 @@
...
   https://github.com/trptcolin/reply/"
   {:author "Chas Emerick"}
-  (:require [clojure.java.io :as io]
+  (:require [clojure.edn :as edn]
+            [clojure.java.io :as io]
             [nrepl.core :as repl]
             [nrepl.ack :refer [send-ack]]
             [nrepl.server :refer [start-server]]))
 
+(defn exit
```
bbatsov: Please, add some docstring to each method. I'm very committed to documenting everything. :-)

src/clojure/nrepl/cmdline.clj
```diff
@@ -153,12 +184,15 @@
...
       (let [bind (options "--bind")
+            auth-wrapper (some-> (options "--authorization-wrapper")
+                                 resolve-auth-wrapper-or-die)
             ;; if some handler was explicitly passed we'll use it, otherwise we'll build one
             ;; from whatever was passed via --middleware
             handler (and (options "--handler") (read-string (options "--handler")))
             middleware (and (options "--middleware") (read-string (options "--middleware")))
             handler (if handler (handler) (build-handler middleware))
-            server (start-server :port port :bind bind :handler handler)
+            server (apply start-server :port port :bind bind :handler handler
```
bbatsov: I don't think `apply` is needed here, as you can safely pass a nil auth-wrapper and this will result in the default being used.

src/clojure/nrepl/middleware/auth.clj
```diff
@@ -0,0 +1,40 @@
+(ns nrepl.middleware.auth
```
bbatsov: The namespace should be documented extensively before the PR is merged.

src/clojure/nrepl/middleware/auth.clj
```diff
@@ -0,0 +1,40 @@
...
+  (:require
+   [nrepl.misc :refer [response-for]]
+   [nrepl.transport :as t])
+  (:import
+   (java.io File FileNotFoundException)))
+
+(defn find-token
```
bbatsov: Btw, I think we should cache the result of this function unless we won't the token to be something you can change while the server is running, which doesn't seem very wise to me.

src/clojure/nrepl/server.clj
```diff
@@ -142,8 +157,9 @@
                         (atom #{})
                         (or transport-fn t/bencode)
                         greeting-fn
-                        (or handler (default-handler)))]
+                        (authorization-wrapper handler))]
     (future (accept-connection server))
+    (binding [*out* *err*] (println "nREPL listening on" (:port server)))
```
bbatsov: This should ideally be eliminated before we wrap up the PR.

test/clojure/nrepl/auth_test.clj
```diff
@@ -0,0 +1,208 @@
+(ns nrepl.auth-test
+  "Authentication system tests.  Should not run tests that expose local
```
bbatsov: Authentication or Authorization. ;-)

I find it amusing that even in the scope of the PR we keep mixing those up. ;-)

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

>> Oh, and I could also see just having a top-level {:kind ::exit ::status
>> ...} catch that handled the clean up and exit (instead of the scattered
>> exit calls), but didn't know which you'd prefer, or if you'd want that
>> much change right now.
>
> Please, expand on this, as I'm not sure I understand your idea. 

Ahh, I just meant that instead of scattered direct calls to exit, we
might have a top-level catch like this:

    (try
      ...
      (catch ExceptionInfo ex
        (let [data (ex-data ex)]
          (when (= :nrepl/exit (:kind data))
            (shutdown-agents)
            (flush)
            (binding [*out* *err*] (flush))
            (System/exit (or (:nrepl/status data) 2)))
          (throw ex))))

And then all the code that currently calls (System/exit ...) directly
would instead:

    (throw (ex-info msg {:kind :nrepl/exit :nrepl/status 2}))

which might well be wrapped up as:

    (nrepl/exit 2 msg)

or whatever.

Both centralizing the handling, and ensuring that (assuming we want it)
any pending catches or finally blocks still have a chance to run.

-- 
Rob Browning
rlb @defaultvalue.org and @debian.org
GPG as of 2011-07-10 E6A9 DA3C C9FD 1FF8 C676 D2C4 C0F0 39E9 ED1B 597A
GPG as of 2002-11-03 14DD 432F AE39 534D B592 F9A0 25C8 D377 8C7E 73A4

bbatsov:
Yeah, that makes sense. 

Btw, in the mean time I've added https://github.com/nrepl/nrepl/commit/9129bf62c02536d5b7259b9bc6bd4b52bea157fa, so you should leverage `nrepl.config` to some extent in this patch.

rlbdv:
[Above](https://github.com/nrepl/nREPL/pull/46#issuecomment-429754530), in response to my comment:

> If that's preferable, we could consider adjusting --middleware to work the same way, but it would break backward compatibliity. If that's not preferable, then I think we may be able to just bind read-eval to false when calling read-string.

Did you mean that I could/should go ahead and change `--middleware`'s behavior to match `--authorization-wrapper`'s, i.e. to support only `foo/bar` via `edn/read-string` and not `#'foo/bar`, or that I should try to preserve `#'` support for either or both of them somehow, if we can do it safely?

bbatsov:
> Did you mean that I could/should go ahead and change --middleware's behavior to match --authorization-wrapper's, i.e. to support only foo/bar via edn/read-string and not #'foo/bar, or that I should try to preserve #' support for either or both of them somehow, if we can do it safely?

It should only support:

* a single string/symbol (which gets converted to a one-element vector)
* a list of strings/symbols

No need to support var references.

> Would you prefer that, or your current require-and-resolve?

They seem pretty similar now, so I guess you can just add `die` to the existing function and that'd be enough (unless I'm missing something).

src/clojure/nrepl/middleware/auth.clj
bbatsov: Btw, after working on `nrepl.config` I realized that it's bad for middleware/internal functions to lookup config themselves. Finding the token should probably be the responsibility of `lein`/`boot`/`clj` and they should pass this in to nREPL. That adds more flexibility to clients to handle configuration they way they see fit. E.g. I opted not to do env vars in `nrepl.config` as I couldn't figure whether they should have higher or lower precedence than stuff in the global/local config file.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> Btw, after working on `nrepl.config` I realized that it's bad for
> middleware/internal functions to lookup config themselves. Finding the
> token should probably be the responsibility of `lein`/`boot`/`clj` and
> they should pass this in to nREPL. That adds more flexibility to
> clients to handle configuration they way they see fit.

I'm not sure I follow yet -- I'd thought that we were planning to
specify one possible strategy by providing an implementation that tools
could use if they wanted to, and which might or might not become a
"standard", but which they could also completely ignore.

e.g. leingingen could specify its own LEIN_AUTHORIZATION_TOKEN,
./.lein-authorization-token, etc...), or it could just say, "handles
authorization via [nREPL's handler](https://...)", and be done with it,
at least until it needs something that nREPL can't accommodate.

Then perhaps if the default is reasonable enough to be adopted, we might
end up with something that works across the main tools, rather than
bespoke, per-tool solutions.

However, your comments make me realize that at a minimum, if we do
decide to provide a default, it should probably look a bit different
from what i have now.  For example, if nothing else, the current
proposal has no way for a tool to provide the "command line" level token
that might override the other envt var, etc. options.  Assuming we
decide to continue, I'll come up with some changes you can consider.

I also think providing a default might be helpful to the extent that it
allows people who haven't or don't want to think about it much to just
use an approach that's hopefully "sensible", and has solid documentation
surrounding any matters that require care.

And if we don't provide anything, then would that mean that use of nrepl
alone, via nrepl.cmdline would have no security?

> E.g. I opted not to do env vars in `nrepl.config` as I couldn't figure
> whether they should have higher or lower precedence than stuff in the
> global/local config file.

FWIW, in my experience command line arguments take precedence over
environment variables which take precedence over a local config file,
which takes precedence over a user config file, which takes precedence
over a system config file (git being a common example, but there are
plenty of others).  And of course any of those might or might not be
supported by a given tool.

e.g. first specified wins:

  - --foo on the command line
  - FOO in the environment
  - foo in ./.config
  - foo in ~/.config
  - foo in /etc/config

And I'd imagined that nrepl might just eventually provide environment
variables for any appropriate config options -- perhaps not all of them,
and perhaps even by mechanical translation (say by specifying that the
config-file name will be upper-cased and dashes will be converted to
underscores).

In any case, I'm happy to adjust the patches, just wanted to clarify
where I should be headed first.

Thanks much

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> No need to support var references.

Got it.

> They seem pretty similar now, so I guess you can just add `die` to the
> existing function and that'd be enough (unless I'm missing something).

Indeed, now that I notice that resolve should do the same thing as
ns-resolve when given a fully qualified symbol, then they're very
similar, excepting the (possibly) friendlier error handling.  I can
integrate them -- though we could also leave it alone if we don't want
the additional complexity in exchange for that.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> Yeah, that makes sense. 
>
> Btw, in the mean time I've added https://github.com/nrepl/nrepl/commit/9129bf62c02536d5b7259b9bc6bd4b52bea157fa, so you should leverage `nrepl.config` to some extent in this patch.

Looking at this, I see the (handler) invocation was dropped, i.e. this:

    handler (if handler (handler) (build-handler middleware))

changed to this:

    handler (or handler (build-handler middleware))

Just wanted to double-check that it was intentional.

https://github.com/nrepl/nrepl/commit/9129bf62c02536d5b7259b9bc6bd4b52bea157fa#diff-157defd03be20ee96b1f635d628a73a3L171

Thanks

rlbdv:
OK, I've pushed an update, but I haven't changed the authorization code much yet (with respect to our current discussion).  Instead, I've broken the unrelated changes out into separate commits, before the the auth changes, and I've rebased onto current master.

Since the unrelated changes had accumulated, I thought it'd be better to consider them on their own merits.  Please let me know if you'd like me to change or omit any of them, or cherry-pick them if you like, and I'll rebase again.

src/clojure/nrepl/middleware/auth.clj
rlbdv: Specifically regarding "I think we should cache the result of this function", the current arrangement was intended to capture the token at the time the handler's created, and so you might capture a different token if/when you create a new handler, but only then.

Perhaps you're saying you'd prefer the token to be captured only once for the entire JVM process?  (Oh, and given your other comments, and our unfinished discussion there, I'm going to move this default auth middleware to its own commit, so you can decide independently whether you want to keep it at all.)

rlbdv:
I've rebased onto current master and split the auth work into two commits.  The first only adds the `:auth-wrapper`, which I think we may have mostly resolved?  And the second contains the `require-nrepl-authorization-token` wrapper that I think we're less certain about.

Oh, and I switched to `:auth-wrapper` from `:authorization-wrapper` because that option can be either, i.e. it's a mechanism without policy.

src/clojure/nrepl/cmdline.clj
```diff
@@ -157,7 +157,7 @@
   "Require and resolve `thing`
   `thing` can be a string or a symbol."
   [thing]
-  (let [thing (symbol thing)]
+  (let [thing (edn/read-string thing)]
```
bbatsov: Won't this blow up if `thing` is a symbol?

rlbdv: Yep, thanks, should be fixed.

bbatsov:
> I've rebased onto current master and split the auth work into two commits. The first only adds the :auth-wrapper, which I think we may have mostly resolved? And the second contains the require-nrepl-authorization-token wrapper that I think we're less certain about.

My only real concern about this is the use of the env variable there. If you only move the configuration to rely only on `nrepl.config` I'd be (mostly) fine with its current state. On a broader note - there's the philosophical question of whether middleware should really read configuration directly or should always get the configuration passed to it via some other means. 

> Oh, and I switched to :auth-wrapper from :authorization-wrapper because that option itself could specify a handler that does authentication or authorization, i.e. it's a mechanism without policy.

That's fine. I was thinking we might call this `message-filter` or something along those lines, as it's actually even more generic that `auth-wrapper` gives it credit. :-) Ultimately we can advertise this as a way to go to the very bottom of the middleware stack.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> That's fine. I was thinking we might call this `message-filter` or
> something along those lines, as it's actually even more generic that
> `auth-wrapper` gives it credit. :-) Ultimately we can advertise this
> as a way to go to the very bottom of the middleware stack.

Agreed -- :message-filter sounds like it might be more appropriate...

In any case, I've been thinking about this a bit too -- about the fact
that support for :message-filter (or whatever), while possibly
convenient, and perhaps a touch potentially safer (i.e. harder to
avoid), may not really be the key thing.

That is we could (of course) do what's needed with just the existing
:handler or :middleware support, and some policy that all the relevant
parties can agree to (perhaps you were suggesting this earlier).  Given
a policy, could also provide a middleware function that implements it,
so that clojure-based tools can just use it.

But at least right, now, I suspect that the policy's going to end up
including environment variables (since it's one of the safe ways to
provide a value directly -- as opposed to the command line), and for
clojure-based tools, I imagine it'd be a lot more convenient to just
rely on some function that we provide, say:

    (or (some-> command-line-token-file maybe-slurp)
        (System/getenv "FOO_AUTHORIZATION_TOKEN")
        (some-> (System/getenv "FOO_AUTHORIZATION_TOKEN_FILE") slurp)
        (maybe-slurp current-directory-token-file)
        (maybe-slurp home-directory-token-file)
        (maybe-slurp system-level-etc-ish-token-file)))

...or whatever policy at least nrepl, cider, and lein can agree on.
Then they don't have to write the code themselves, nor track any changes
to the policy.

That said, it occurs to me that, ignoring the "too many small artifacts"
concern, it would be better if this function were in its own library so
that a client that wants to use it doesn't have to haul in the whole
server.

Thanks again

rlbdv:
(...and honestly, a system-level /etc based token file seems unlikely to be all that useful in this case.)

bbatsov:
```In any case, I've been thinking about this a bit too -- about the fact
that support for :message-filter (or whatever), while possibly
convenient, and perhaps a touch potentially safer (i.e. harder to
avoid), may not really be the key thing.

That is we could (of course) do what's needed with just the existing
:handler or :middleware support, and some policy that all the relevant
parties can agree to (perhaps you were suggesting this earlier).  Given
a policy, could also provide a middleware function that implements it,
so that clojure-based tools can just use it.

But at least right, now, I suspect that the policy's going to end up
including environment variables (since it's one of the safe ways to
provide a value directly -- as opposed to the command line), and for
clojure-based tools, I imagine it'd be a lot more convenient to just
rely on some function that we provide, say:```

Great points! The more I think about this, the more I realize how careful we have to be with its design. I'd still prefer to use nREPL's config files over env variables, though, but that's a relatively small implementation detail. 

You know what - can you fire up a separate PR that just includes the cmdline improvements you made, so I can merge those (I'm too lazy to cherry pick the commits from this PR 😄 ) and we can continue the conversation just focused on the filter/policy stuff.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> Great points! The more I think about this, the more I realize how
> careful we have to be with its design. I'd still prefer to use nREPL's
> config files over env variables, though, but that's a relatively small
> implementation detail.

I'm not absolutely certain either way myself, and of course it's always
something we can figure out (and add) later if we're not sure, but at a
minimum (and please consider this an incomplete idea), I've imagined
that I might want to provide some standard setting like
NREPL_DISALLOW_UNRESTRICTED_ACCESS that everything respects, and that
would cause nrepl to just refuse to operate when the auth system isn't
active (whatever that ends up meaning).

i.e. I know we don't want to make auth the default yet, but it'd be nice
if there were a way to opt-in to requiring it.

Though now that I think about it, I suppose the setting could just go in
~/.nrepl -- perhaps that'd be sufficient.

jwhitlark:
I hate to pollute the thread, but since it's related, can I ask where SSL/TLS support would be discussed?

rlbdv:
Jason Whitlark <notifications@github.com> writes:

> I hate to pollute the thread, but since it's related, can I ask where
> SSL/TLS support would be discussed?

Not sure -- I'm not planning to try to address that now.  Perhaps a new
issue or something similar?

And apologies for the inactivity -- I've been tied up by various things
recently, particularly Emacs for the buster release, but I hope to get
back to this in the next few weeks.

bbatsov:
@rlbdv Don't worry - in the mean time Lein 2.8.3 came out and it bundles nREPL 0.5.3, which is a great victory as now Lein natively supports the new nREPL configuration files. 

We've also had @cgrand join our team and he has been working on some really great improvements for the next release. 

> Not sure -- I'm not planning to try to address that now.  Perhaps a new
issue or something similar?

Yeah, this seems unrelated. I guess we can have some encrypted transports, but that's definitely not something high priority for me right now. In the mean time using the HTTPS transport (Drawbridge) seems like a decent workaround, or just tunnelling a normal unencrypted connection over SSH.

MalloZup:
autogenerated with https://github.com/MalloZup/doghub:  pr inactive since 200 days. Please update the PR or close it

rlbdv:
Just as an update, I set this aside to pursue #204 first, and may (or may not) return to this once that one's resolved.

bbatsov:
@rlbdv Should we close this one? It has been dormant for quite a while.

rlbdv:
I think the filesystem socket support we added is sufficient for me for now, so perhaps -- I'm not likely to pursue it further right now, though I suppose it might still be interesting for situations where filesystem sockets aren't sufficient.

Issue #12, opened 2017-11-06
Support custom value rendering middleware


cemerick:
Currently, nREPL's interruptible-eval middleware hardcodes a dependency on pr-values. It would be nicer if the user could either specify alternate rendering middleware or if nREPL's built-in middleware accepted an arbitrary function to render values with. Specifically, this would enable pretty-printing REPL output.

I've set up a middleware to demonstrate this, in combination with the Puget printing library:
https://github.com/greglook/whidbey

Currently it is possible to replace the default middleware, but it involves some ugly runtime metadata manipulation which reaches into the nREPL internals. Addressing this would be another step towards simplifying pretty-printing/color integration in the REPL.

(from https://dev.clojure.org/jira/browse/NREPL-55)

bbatsov:
@greglook Would you be interested in contributing this upstream? 

Now we're back in the business of accepting PRs! 🎉

arrdem:
Hummmm this could also drastically simplify the implementation of content-type handlers...

greglook:
Might be able to look at this in a few days; it'd certainly be nice to ditch the metadata wrangling.

bbatsov:
@greglook Certainly! 👍

PR #53, opened 2018-10-14
[WIP] #12: Value rendering function eval parameter

benjaminran:
This is not ready to be merged, but the core of the change for #12 is here (I think--feedback welcome). I'll record my thoughts/questions here as I go.

Some questions:

Should a client be able to add multiple separate rendering functions?
=> No, b/c composition of functions can just be done inside a single :renderf string

(How) should symbols from the nrepl session be isolated from the rendering function?
=> I think the rendering function should be isolated from the nrepl session; that way the rendering function can't cause side effects on the nrepl session state and the nrepl session state can't influence the behavior of the rendering function. TODO: figure out if eval already does this or how to add it.

Sample nrepl-messages log (rendered values in value2):
```
   (-->
     id         "78"
     op         "eval"
     session    "dfcbdd42-eace-497e-8e6f-cb1c79f648c4"
     time-stamp "2018-10-13 23:02:07.091198000"
     code       "(zipmap [:a :b :c :d :e] (repeat (zipmap [:a :b :c :d :e] (take 5 (range)))))"
     renderf    "(fn [v] (let [repr (java.io.StringWriter.)] (clojure.pprint/pprint v repr) (str repr)))"
   )
   (<--
     id         "78"
     session    "dfcbdd42-eace-497e-8e6f-cb1c79f648c4"
     time-stamp "2018-10-13 23:02:07.222323000"
     ns         "user"
     value      "{:a {:a 0, :b 1, :c 2, :d 3, :e 4}, :b {:a 0, :b 1, :c 2, :d 3, :e 4}, :c {:a 0, :b 1, :c 2, :d 3, :e 4}, :d {:a 0, :b 1, :c 2, :d 3, :e 4}, :e {:a 0, :b 1, :c 2, :d 3, :e 4}}"
     value2     "{:a {:a 0, :b 1, :c 2, :d 3, :e 4},
    :b {:a 0, :b 1, :c 2, :d 3, :e 4},
    :c {:a 0, :b 1, :c 2, :d 3, :e 4},
    :d {:a 0, :b 1, :c 2, :d 3, :e 4},
    :e {:a 0, :b 1, :c 2, :d 3, :e 4}}
   "
   )
   (<--
     id         "78"
     session    "dfcbdd42-eace-497e-8e6f-cb1c79f648c4"
     time-stamp "2018-10-13 23:02:07.224126000"
     status     ("done")
   )
   (<--
     id                 "78"
     session            "dfcbdd42-eace-497e-8e6f-cb1c79f648c4"
     time-stamp         "2018-10-13 23:02:07.225471000"
     changed-namespaces (dict)
     repl-type          "clj"
     status             ("state")
   )
```

### DIFF
--- a/src/clojure/nrepl/middleware/pr_values.clj
+++ b/src/clojure/nrepl/middleware/pr_values.clj
@@ -16,7 +16,7 @@
    This allows evaluation contexts to produce printed results in :value
    if they so choose, and opt out of the printing here."
   [h]
-  (fn [{:keys [op ^Transport transport] :as msg}]
+  (fn [{:keys [op ^Transport transport renderf] :as msg}]
     (h (assoc msg
               :transport (reify Transport
                            (recv [this] (.recv transport))
@@ -27,6 +27,13 @@
                                       (dissoc resp :printed-value)
                                       (if-let [[_ v] (find resp :value)]
                                         (assoc resp
+                                               :value2 (let [repr (java.io.StringWriter.)]
+                                                         (apply
+                                                          (if renderf
+                                                            (eval (read-string renderf))
+                                                            #((if *print-dup* print-dup print-method)
+                                                              % repr))
+                                                          [v]))
                                                :value (let [repr (java.io.StringWriter.)]
                                                         (if *print-dup*
                                                           (print-dup v repr)


src/clojure/nrepl/middleware/pr_values.clj
```diff
@@ -27,6 +27,13 @@
                                       (dissoc resp :printed-value)
                                       (if-let [[_ v] (find resp :value)]
                                         (assoc resp
+                                               :value2 (let [repr (java.io.StringWriter.)]
+                                                         (apply
+                                                          (if renderf
+                                                            (eval (read-string renderf))
```
bbatsov: Seems to me that `renderf` (or whatever we end up calling it) should be a symbol that we'd resolve and apply. Passing a lambda here doesn't seem very wise to me.

bbatsov:
> (How) should symbols from the nrepl session be isolated from the rendering function?

Shouldn't we just assume a standard format that the rendering function has and that it returns a string, and just pass a symbol that we'd need to resolve and use? Generally I assume that the most common use case here would be to use something like zprint, etc, instead of people rolling out their own functions. 

Alternative we can add some configuration to the eval middleware that would need to be set by a special op, but that might be an overkill. 

P.S. Thanks for tackling this! It's a very important avenue for improvement that we've been ignoring for quite a while now. 

@arrdem @SevereOverfl0w Any thoughts on this?

benjaminran:
> Shouldn't we just assume a standard format that the rendering function has and that it returns a string, and just pass a symbol that we'd need to resolve and use? Generally I assume that the most common use case here would be to use something like zprint, etc, instead of people rolling out their own functions.

Yeah, that makes a lot more sense. That's just as good feature-wise, maybe makes the api more intuitive, and leaves less that we need to worry about authorization for.

bbatsov:
OK, so I guess we can work in this direction then. Btw, there's one more thing to take into account - config params that might have to be passed to the pprint function (e.g. print-width). Guess this can be some optional map after the function or something like that. 

As a reference - here's how it's done current in cider-nrepl https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/pprint.clj#L56

The code might serve as a nice mainline for the task at hand.

bbatsov:
Btw, another approach for this would be https://github.com/nrepl/nREPL/blob/e77945a70f68989f69e955eef9e6cbe75cedacf4/src/clojure/nrepl/middleware/interruptible_eval.clj#L105

bbatsov:
I'll close this in favour of #59 as the approach is the same and #59 is almost ready to be merged.

@benjaminran At any rate - I'm very grateful from your contribution and I hope you'll pick some other ticket to tackle next!

benjaminran:
Alright, thanks :).

PR #58, opened 2018-10-18
Fix link to manual in README.md

mallt:
This PR fixes the link to the manual.

The link to the history should be changed as well, but I'm not sure if there is a link available with "latest" to avoid having to specify the version number?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ code in remote environments.
 ### How is this different from the "contrib" [tools.nrepl](https://github.com/clojure/tools.nrepl/) project?
 
 Check the brief history of nREPL, available
-[here](https://docs.nrepl.xyz/en/latest/about/history/).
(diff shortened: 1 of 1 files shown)


bbatsov:
> The link to the history should be changed as well, but I'm not sure if there is a link available with "latest" to avoid having to specify the version number?

With the new manual it's `latest` by default.

mallt:
I tried it but [the link with latest](https://nrepl.xyz/nrepl/latest/about/history.html) returns a page not found.

bbatsov:
Ops, my bad. I'm not that skilled with Antora yet, so I'm not sure how to generate the proper links. Actually this build is from `master`, but the name is 0.4.5. I guess you can just link to whatever link we have right now before I educate myself better or @SevereOverfl0w educates me. :D

PR #59, opened 2018-10-19
Custom rendering logic for pr-values middleware

greglook:
Address #12 by porting some code I wrote to do this in https://github.com/greglook/whidbey but merged into the main `pr-values` middleware instead of a separately-installed hack.

Somewhat a WIP - it appears that there are 2-3 tests that are not passing on master, so I don't think I've broken anything. The code does get exercised by the existing test suite (since it's an extension to `pr-values`, but I haven't explicitly tested the custom rendering functionality. Any pointers there would be helpful.

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### New features
 
-* [#12](https://github.com/nrepl/nREPL/issues/10): Support custom rendering
+* [#12](https://github.com/nrepl/nREPL/issues/12): Support custom rendering
   function in `pr-values`, enabling pretty-printed REPL results.
 
 #### Bugs fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ### master (unreleased)
 
+#### New features
+
+* [#12](https://github.com/nrepl/nREPL/issues/10): Support custom rendering
+  function in `pr-values`, enabling pretty-printed REPL results.
+
 #### Bugs fixed
 
 * [#10](https://github.com/nrepl/nREPL/issues/10): Bind *1, *2, *3 and *e in cloned session.
--- a/doc/modules/ROOT/pages/design/middleware.adoc
+++ b/doc/modules/ROOT/pages/design/middleware.adoc
@@ -166,3 +166,11 @@ value instead.
 
 This enables using libraries like link:https://github.com/greglook/puget[puget]
 to pretty-print the evaluation results automatically.
+
+[source,clojure]
+----
+{:op :eval
+ :code "(+ 1 1)"
+ :renderer 'my.custom/print-value
+ :render-options {:print-width 120}}
+----
--- a/doc/modules/ROOT/pages/design/middleware.adoc
+++ b/doc/modules/ROOT/pages/design/middleware.adoc
@@ -154,3 +154,15 @@ fundamental to allowing simultaneous activities in the same nREPL.
 For instance - if you want to evaluate two expressions simultaneously
 you'll have to do this in separate session, as all requests within the
 same session are serialized.
+
+== Pretty Printing
+
+nREPL includes a `pr-values` middleware to render the results of evaluated
+forms as strings for returning to the client. By default, this will use either
+`print-dup` or `print-method` to match the standard Clojure `print` behavior.
+To customize this you can pass a custom `:renderer` as a symbol along with the
+message. If present, this will be resolved to a function and used to render the
+value instead.
+
+This enables using libraries like link:https://github.com/greglook/puget[puget]
+to pretty-print the evaluation results automatically.
--- a/src/clojure/nrepl/middleware/pr_values.clj
+++ b/src/clojure/nrepl/middleware/pr_values.clj
@@ -2,7 +2,7 @@
   {:author "Chas Emerick"}
   (:require
    [clojure.string :as str]
-   [nrepl.middleware :as middleware])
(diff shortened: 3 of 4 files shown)


src/clojure/nrepl/middleware/pr_values.clj
```diff
@@ -1,43 +1,67 @@
 (ns nrepl.middleware.pr-values
   {:author "Chas Emerick"}
   (:require
-   [nrepl.middleware :refer [set-descriptor!]])
+    [clojure.string :as str]
```
bbatsov: I guess the indentation change was unintentional. I'll assume you're not an Emacs user (or you're using non-default indentation settings). :-)

src/clojure/nrepl/middleware/pr_values.clj
```diff
@@ -1,43 +1,67 @@
...
+    [nrepl.middleware :as middleware])
   (:import
-   nrepl.transport.Transport))
+    nrepl.transport.Transport))
+
+
+(defn- print-renderer
```
bbatsov: I guess a better name for this might be `default-print-renderer` or something along those lines.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ### master (unreleased)
 
+#### New features
+
+* [#12](https://github.com/nrepl/nREPL/issues/10): Support custom rendering
```
bbatsov: I think you made a small copy/paste mistake here.

greglook:
Thanks for the :eyes:, I'll fix up the issues.

src/clojure/nrepl/middleware/pr_values.clj
```diff
@@ -1,43 +1,67 @@
...
+  `:value` slot.
+
+  If `:value` is already a string, and a sent message's `:printed-value` slot
+  contains any truthy value, then `:value` will not be re-printed.  This allows
+  evaluation contexts to produce printed results in `:value` if they so choose,
+  and opt out of the printing here."
+  [handler]
+  (fn [{:keys [op ^Transport transport renderer] :as msg}]
+    (let [render-fn (if renderer
+                      (find-var (symbol renderer))
```
bbatsov: One more small problem - `find-var` is going to return nil unless the namespace of the target var is required, so it might be better to use something like a combination of require + resolve (or try find-var first, then try to require/resolve and fallback to the default).

doc/modules/ROOT/pages/design/middleware.adoc
```diff
@@ -154,3 +154,15 @@ fundamental to allowing simultaneous activities in the same nREPL.
...
+
+nREPL includes a `pr-values` middleware to render the results of evaluated
+forms as strings for returning to the client. By default, this will use either
+`print-dup` or `print-method` to match the standard Clojure `print` behavior.
+To customize this you can pass a custom `:renderer` as a symbol along with the
+message. If present, this will be resolved to a function and used to render the
+value instead.
+
+This enables using libraries like link:https://github.com/greglook/puget[puget]
+to pretty-print the evaluation results automatically.
```
bbatsov: We might add here some example eval request using either Clojure's built-in pprint or puget so tools writers will have a better idea what they are supposed to do. I guess it's not immediately obvious that we'll pull out the `:renderer` from an `eval` op request.

bbatsov:
@greglook Apart from my small documentation comment and some missing test everything looks good now. Seems there are no failing tests after all. :-)

Just something small I forgot about - are we going to assume that the render function is always going to be configured via dynamic var or should we also make it possible to pass `renderer-params` (or whatever) that's some vector that's going to be passed to the function when it's invoked. I'm just not very familiar with projects like fipp, zprint and so on and I wonder what's the way people normally configure their behaviour.

greglook:
> Apart from my small documentation comment and some missing test everything looks good now.

Good! I'll see if I can find the right place to add a test or two.

> are we going to assume that the render function is always going to be configured via dynamic var or should we also make it possible to pass renderer-params (or whatever) that's some vector that's going to be passed to the function when it's invoked.

That's a good point - one open issue on whidbey is about exactly this: https://github.com/greglook/whidbey/issues/15
Since we are passing a symbol to resolve instead of a function closure, we can't use partial application or anonymous functions to wrap the options up. I think supporting an additional `eval` message parameter to indicate `:render-options` would work well.

src/clojure/nrepl/middleware/pr_values.clj
```diff
@@ -19,10 +19,12 @@
...
   (when-let [var-sym (and var-sym (symbol var-sym))]
-    ; TODO: log a warning if we were given a symbol but couldn't find it?
-    (or (find-var var-sym)
-        (do (require (symbol (namespace var-sym)))
-            (resolve var-sym)))))
+    (try
+      (require (symbol (namespace var-sym)))
+      (resolve var-sym)
+      (catch Exception ex
+        ; TODO: emit a warning here?
```
bbatsov: Probably you can simply use `log` here.

greglook:
@bbatsov one final thought I had - what do you think about memoizing `resolve-renderer`? It seems likely that there will be many, many calls with the same renderer symbol, so this could save some repeated work. On the other hand, if the resolution process is cheap or we need the dynamism to pick up new var definitions, then it should stay as-is.

bbatsov:
I think `find-var` should be reasonably fast, so let's not do any premature optimizations here. 

The PR looks great now! Thanks for the outstanding contribution! 🙇 

I hope we'll be seeing you around helping with some of the other open tickets! 😉

PR #65, opened 2018-10-24
[Fix #33] Add ability to change value of *print-namespace-maps*

### LINKED ISSUE #33: *print-namespace-maps* is always true
## Expected behavior, actual behavior and steps to reproduce the problem

Evaluating a map with namespaced keys like
`{:person/name "Fred" :person/age 50}`
in the cider repl prints the following return value
`#:person{:name "Fred", :age 50}`
This is due to `*print-namespace-maps*` defaulting to `true`.
Normally, it should be possible to change the value of this var permanently with `set!` thusly:
```
user> (set! *print-namespace-maps* false)
false
user> *print-namespace-maps*
false
```
However, the actual behavior is thus:
```
user> (set! *print-namespace-maps* false)
false
user> *print-namespace-maps*
true
```
Similarly, it should be possible to temporarily change the var with `binding`:
```
user> (binding [*print-namespace-maps* false]
        [*print-namespace-maps* {:person/name "Fred" :person/age 50}])
[false {:person/name "Fred" :person/age 50}]
```
But, although the var seems to change, the desired behavior remains to be seen;
```
user> (binding [*print-namespace-maps* false]
        [*print-namespace-maps* {:person/name "Fred" :person/age 50}])
[false #:person{:name "Fred", :age 50}]
```

I first posted this problem as an issue at the CIDER repo: https://github.com/clojure-emacs/cider/issues/2280 
@bbatsov pointed out that it must be a problem in nrepl, so I reposted the issue here and closed the original issue.
Some of the discussion in the thread underneath the original issue might prove useful. 

Following the issue instructions there, I included information about my CIDER version, emacs version and operating system. I'll include them here, as well;
 
### CIDER version information

```
;; CIDER 0.15.0 (London), nREPL 0.2.13
;; Clojure 1.9.0, Java 1.8.0_92
```

### Emacs version

GNU Emacs 24.5.1 (x86_64-apple-darwin15.3.0, NS apple-appkit-1404.34) of 2016-06-01

### Operating system

OSX El Capitan, Version 10.11.6

mallt:
This PR adds the ability for nREPL clients to change the value of `*print-namespace-maps*`.

- [X] All tests are passing
- [X] You've updated the [changelog](../CHANGELOG.md)(that only applies to user-visible changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 #### Bugs fixed
 
 * [#10](https://github.com/nrepl/nREPL/issues/10): Bind *1, *2, *3 and *e in cloned session.
+* [#33](https://github.com/nrepl/nREPL/issues/33): Add ability to change value of `*print-namespace-maps*`.
 
 #### Changes
 
--- a/src/clojure/nrepl/middleware/interruptible_eval.clj
+++ b/src/clojure/nrepl/middleware/interruptible_eval.clj
@@ -21,6 +21,13 @@
   "Function returning the evaluation of its argument."
   nil)
 
+(defmacro ^{:private true} def-dynamic-when-not-resolved
+  [sym value]
+  (when-not (resolve sym)
+    `(def ~(vary-meta sym assoc :dynamic true) ~value)))
+
+(def-dynamic-when-not-resolved *print-namespace-maps* nil)
+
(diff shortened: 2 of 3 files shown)


bbatsov:
Looks good to me!

PR #66, opened 2018-10-28
Introduce external server configuration

bbatsov:
That's something I've been thinking about for a long time and it's finally coming in 0.5 (after a polish it and add some docs). 

Basically the idea is to make it easier to have host-wide and project-specific configs, so you can have stable bind-addresses, ports, transports, middleware, etc. 

Obviously that's also going to require a bit of changes to tools like `boot` and `lein`, but those are going to be trivial. Feedback is welcome!

I also plan to follow this up with `clj` writing in `.nrepl/` some information about each running server, so it's easy for something like CIDER to just go check this dir and suggest all the local running instances as something you can connect to. That's going to be a massive improvement over the `ps` grepping we do right now and will finally work on Windows as well. :-)

### DIFF
--- a/src/clojure/nrepl/config.clj
+++ b/src/clojure/nrepl/config.clj
@@ -1,0 +1,56 @@
+(ns nrepl.config
+  "Server configuration utilities.
+  Some server defaults can be configured via
+  a configuration file or env variables.
+  This namespace provides convenient API to work
+  with them."
+  {:author "Bozhidar Batsov"}
+  (:require
+   [clojure.java.io :as io]
+   [clojure.edn :as edn]))
+
+(def ^:private home-dir
+  "The user's home directory."
+  (System/getProperty "user.home"))
+
+(def config-dir
+  "nREPL's configuration directory.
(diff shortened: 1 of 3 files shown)


src/clojure/nrepl/config.clj
```diff
@@ -0,0 +1,72 @@
...
+  "Load edn from an io/reader source (filename or io/resource)."
+  [source]
+  (try
+    (with-open [r (io/reader source)]
+      (edn/read (java.io.PushbackReader. r)))
+
+    (catch java.io.IOException e
+      (printf "Couldn't open '%s': %s\n" source (.getMessage e)))
+    (catch RuntimeException e
+      (printf "Error parsing edn file '%s': %s\n" source (.getMessage e)))))
```
isker: Why do we catch these exceptions at all?

bbatsov: Good point! I was just playing with some things here. I guess I should just clean this up and silently return `nil` or `{}`.

isker:
Looks convincing to me, though this is my first time reading nrepl code 👍.

PR #69, opened 2018-10-31
[Fix #68] Avoid illegal access warning in set-line! of interruptible-eval

### LINKED ISSUE #68: Eliminate illegal access warnings
When we start the server today on JDK 9+ we get:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by nrepl.middleware.interruptible_eval$set_line_BANG_ to field java.io.FilterReader.in
WARNING: Please consider reporting this to the maintainers of nrepl.middleware.interruptible_eval$set_line_BANG_
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
There we're setting some protected fields, which was never really a great approach, but now it seems this might be blocked down the road. We need to see how we can set the metadata without breaking data encapsulation. Maybe we can get some inspiration from `clojure.main`?

@mallt Would like to tackle this? I assume it'd be pretty straight-forward.

mallt:
This PR fixes the illegal access warning caused by the set-line! function of interruptible-eval. Since [this commit](https://github.com/clojure/clojure/commit/3d216d45b63781586bb79144268bb2fb88dd6f28) a `setLineNumber` method is available on the `LineNumberingPushbackReader` (added in clojure v1.7.0). 

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * [#10](https://github.com/nrepl/nREPL/issues/10): Bind *1, *2, *3 and *e in cloned session.
 * [#33](https://github.com/nrepl/nREPL/issues/33): Add ability to change value of `*print-namespace-maps*`.
+* [#68](https://github.com/nrepl/nREPL/issues/68): Avoid illegal access warning caused by set-line!.
 
 #### Changes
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Nice! 

Btw, what about `set-column!`? Is there now a better way to handle it?

mallt:
No I'm afraid not. But for some reason the set-column! doesn't seem to trigger the illegal access warning.

bbatsov:
Yeah, I noticed this, but I'm curious why. It seems to me that we're making a private field accessible there. Anyways, it is what it is! Thanks for fixing this so quickly!

PR #72, opened 2018-11-02
make logo in README @2x

tonsky:


tonsky:
something is not right with that coverage tool

PR #79, opened 2018-11-10
Can truncate value and outputs returned from eval op

### LINKED ISSUE #75: Make it possible to limit the size of returned values and output
Some clients (e.g. Emacs) can't handle very well huge strings in their REPL buffers. It's also questionable how good UX this constitutes as well. 

Ideally we should be able to specify some value/output limit in eval ops, so that we can trim the result accordingly. We can either return the full and trimmed results in the same message or just have clients request to eval the same code again without any size limits if they happen to need the big result. 

See https://github.com/clojure-emacs/cider/issues/1115 and https://github.com/clojure-emacs/cider/issues/2507 for more details. @gonewest818 started working on something very similar a while back (https://github.com/clojure-emacs/cider/issues/1115#issuecomment-374487527), so someone might pick up his work if they want to, even though a simpler approach with no caching of the results/output will be fine as well.

pfeodrippe:
Add :trim as a optional argument to eval op that receives a integer
and use `subs` to get a substring until the `trim` size.

Try to fix #75

- [x] All tests are passing

Thanks!


bbatsov:
Thanks for tackling this! You're certainly on the right track, but there are some extra consideration:

* the values are not the only thing that can be problematic size-wise - we should also handle long strout/stderr output (e.g. `println` producing something huge). Output should be trimmed in the eval middleware I guess.
* we need some clear indication for clients that a value/output was truncated and we also need to give them access to the full result/output. There are several way to approach this, but generally we'll probably need to return two separate keys (truncated and original) or store the full results in some hashmap and provide an extra op to query it. (that's going to be a more effective variant from a networking perspective).

One example of this would be to send `:trim something` and get `:value foo :truncated true :raw_value some_key` or something along those lines. 

Obviously the first approach (returning two keys like `:value` and `:truncated_value`) would be simpler. The pros and cons of both approaches have to examined a bit more carefully. 

* when truncating values they'll likely be made non pretty-printable (just a guess, I haven't really thought much about this). 

@SevereOverfl0w Any thoughts on this?

pfeodrippe:
I will add a :trimmed key to the returned message

pfeodrippe:
Oh, saw your comment now here.

I think returning two keys is a better approach, gonna code it tonight, thanks for the comments!

src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -72,7 +72,7 @@
 
    It is assumed that `bindings` already contains useful/appropriate entries
    for all vars indicated by `clojure.main/with-bindings`."
-  [bindings {:keys [code ns transport session eval file line column] :as msg}]
+  [bindings {:keys [code ns transport session eval file line column trim] :as msg}]
```
bbatsov: I'd probably call this `truncate`, as `trim` usually carries a different meaning.

src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -117,8 +117,12 @@
                     (.flush ^Writer out)
                     (reset! session (maybe-restore-original-ns @bindings))
                     (t/send transport (response-for msg
-                                                    {:value v
-                                                     :ns (-> *ns* ns-name str)})))
+                                                    (merge
+                                                     {:value v
+                                                      :ns (-> *ns* ns-name str)}
+                                                     (if trim
```
bbatsov: I think the `merge` and `if` are redundant. You can simply pass `trim` as nil if it's not in the request.

pfeodrippe:
It was created a `nrepl.middleware.truncate` (maybe I could think harder for a better name) which is required by `nrepl.middleware.pr-values` and `nrepl.middleware.session`.

Is it a good approach in general? Maybe non eval operations could be affected (if they have a :value, :out or :err keys at their response messages).

bbatsov:
Yeah, I think this is a good approach indeed. This also handles the print over issue as we can apply the pr-value middleware first and truncate value afterwards. 

At this point we just need to think a bit about the truncation strategy. Ideally for collections we should do the truncation at some value boundaries that are less than the target size, but that's not a big deal. We might also make it possible for users to pass a `truncate-fn` or something like this to alter the default truncation strategy. We did something very similar recently with `pr-value` and the `renderer` function there.

bbatsov:
We should probably rename the middleware, but I'm not coming up with brilliant ideas either.

bbatsov:
Btw, just realized there's one small problem with the current design - you can't truncate only values/output and you can't have different truncation limits for them. Maybe those can be two middleware? Not sure how big of an issue this is and what's the best course of action.

pfeodrippe:
> Btw, just realized there's one small problem with the current design - you can't truncate only values/output and you can't have different truncation limits for them. Maybe those can be two middleware? Not sure how big of an issue this is and what's the best course of action.

Good point! This appears to be the main issue here, have to think about it before going on with the rest of the PR, thanks!

bbatsov:
> About the collections, you mean using something like (take truncate coll) instead of truncating the string representation of the collection?

Ah, no. I think that this would complicate the code too much. I was mostly meant it'd be nice to truncate the printed representation in a nice manner - e.g. `[100 200 30000 5000 600000]` with size 10 would become just `[100 200`. That's definitely not a big deal, though, so I wouldn't not fret on it too much. I guess splitting by whitespaces would be too simplistic as you can have some collection of strings with spaces in them.

pfeodrippe:
@bbatsov Added `truncate-value-strategy` and `truncate-output-strategy`, besides the two middlewares, maybe separation between truncation of `err` and `out` would be too much?

src/clojure/nrepl/middleware/truncate_values.clj
```diff
@@ -0,0 +1,62 @@
...
+      (.recv transport))
+    (recv [this timeout]
+      (.recv transport timeout))
+    (send [this resp]
+      (.send transport (truncate resp truncate-fn opts))
+      this)))
+
+(defn truncate-values
+  "Middleware that returns a handler which truncates any `:value` slot
+  in a message to a specified size using the `:truncate` slot."
```
bbatsov: That docstring seems outdated.

pfeodrippe:
Thanks for the review! Just fixed the docstrings of the new middlewares.

bbatsov:
Hmm, seems the message I wrote in the morning here got lost somehow... :-(

So, I thought more about this and have some doubts about the current design and I'd like to share them:

* We can achieve pretty easy the same results client side - every client can truncate easily results and output directly. 
* With the current design we still send potentially huge output to clients, which might have massive implications on their performance.
* I wonder if we can somehow have pretty-printers take care of the truncation of values, as I'm not super happy with having two middlewares for printing and truncation.

I won't offer solutions, as I don't really know the best course of action. Instead I'd like to ask other users of nREPL like @SevereOverfl0w, @PEZ and @liquidz to share what would be a really useful functionality for them when it comes to dealing with huge results. 

I want us to find a solution collaboratively and to all feel comfortable about it.

liquidz:
I have two opinions of nREPL client's side and developer side.

* nREPL client's side
  * I cannot agree with the design returning original key and truncated key because the message becomes very huge.
  * Since the truncating cases are various, it seems there are more inconvenient  controling it on the server side.
* Developer side
  * I want a full output while debugging, but want truncated output when huge outputs occur with unexpected error.
    * On the other hand, if I don't know the cause of the error from the truncated output, I want to check the full output.
  * The point is that the result developer want depends on the operation they are currently doing, so I cannot decide the best.

> store the full results in some hashmap and provide an extra op to query it. (that's going to be a more effective variant from a networking perspective).

I think this is a better desgn.

PEZ:
What if the client says (using `:trim`) that it wants to be shielded from results above a certain size and the server does that,  returning a stringified result `subs`:ed to that size, along with a `:trimmed true` and maybe also `:full-size` iinforming about that the message has been trimmed (and what the full size of it is). And then that extra op to query the full result.

That would allow me to build nice UI:s for evaluation, I think.

bbatsov:
Looking at the feedback from me, @liquidz and @PEZ it seems we're all on the same page about what would be the ideal implementation. :-)

PEZ:
Pondering the UI I want to provide, I think that most of the times the full structure will not be requested.

(I imagine the implementation will need some strategy for what to do with interested results.)

pfeodrippe:
Thanks for the comments, everyone!

So the main objective here is to add one op to query the cached full result if the message is truncated, correct?

Continue with one or two middlewares for truncation?

bbatsov:
I guess it can be just one with handles both values and output unless someone thinks this is a bad idea.

pfeodrippe:
Now the PR adds one middleware instead of two and can fetch the result using the `fetch-raw` op which receives a `raw-id` as argument.

bbatsov:
@pfeodrippe Looks great to me!

The only thing missing is returning some info about the size of the original value. That's be useful in UIs to show something like (displaying only 100 characters out of 34343434). 

You can also document the middleware descriptor and this to the manual. We should add some note that the API is experimental at this point and subject to changes.

src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -4,6 +4,7 @@
    clojure.main
    [nrepl.middleware :refer [set-descriptor!]]
    nrepl.middleware.pr-values
+   nrepl.middleware.truncate-values-and-outputs
```
bbatsov: I guess we can shorten the name to simply `truncate`, but that's not a big deal.

src/clojure/nrepl/middleware/truncate_values_and_outputs.clj
```diff
@@ -0,0 +1,99 @@
+(ns nrepl.middleware.truncate-values-and-outputs
```
bbatsov: Add a bit of explanations here about the rationale and high-level logic, plus a note we've added this in 0.5.

src/clojure/nrepl/middleware/truncate_values_and_outputs.clj
```diff
@@ -0,0 +1,99 @@
...
+  [var-sym]
+  (when-let [var-sym (and var-sym (symbol var-sym))]
+    (try
+      (require (symbol (namespace var-sym)))
+      (resolve var-sym)
+      (catch Exception ex
+        (log/warn "Couldn't resolve truncate strategy function" var-sym)
+        nil))))
+
+(defonce raw-results (atom {}))
```
bbatsov: I think this should be private.

bbatsov:
Btw, probably it'd be nice if users could truncate out and err with different settings. Generally err messages are more useful untruncated or they could have a bigger truncation limit.

pfeodrippe:
> Btw, probably it'd be nice if users could truncate out and err with different settings. Generally err messages are more useful untruncated or they could have a bigger truncation limit.

Gonna work on that, thanks!

vspinu:
I am late to the party, but I am worried that tackling this on the server side doesn't offer a flexible enough solution and one would still need to fix it on the client side. Thus, resulting in confusing set of inter-playing configuration settings.

To my mind there is no huge overhead in piping long output to the client as long as the client can interrupt the stream. The real bottleneck on the client side is redisplay which should be tackled on the client side by accumulating output into an invisible field and displaying  a memo like `[truncated ... click to show 5000 more characters, [C-c C-b] to interrupt]`. There might be many of such truncated messages during the session and the client is free to uncover any of those as long as the REPL buffer is not erased.

Tackling it on the client side is pretty straightforward and would avoid dealing with caches which cannot (shouldn't) be implemented on the server side.

pfeodrippe:
Thanks for taking the time for the great comments, @vspinu!

Maybe this should be coded on its own repo as a isolated and generic middleware, if really needed. This is showing to be really server-side controversial.

pfeodrippe:
Should I close this PR? It appears to be really a lot of server effort for a simple client problem, but I'm open for other points of view.

bbatsov:
Yeah, I guess we should close this. Thanks for all of the efforts you've put into this, @pfeodrippe! I think that regardless of the result your work produced some valuable insights as to how to tackle this properly down the road.

pfeodrippe:
> Yeah, I guess we should close this. Thanks for all of the efforts you've put into this, @pfeodrippe! I think that regardless of the result your work produced some valuable insights as to how to tackle this properly down the road.

Thanks for your time and the time of all the other contributors who commented here =) see you in other PRs o/

PR #80, opened 2018-11-12
Fix cljdoc.edn

martinklepsch:
For now these empty maps are needed, will be fixed here: https://github.com/cljdoc/cljdoc/issues/205


bbatsov:
Thanks!

PR #81, opened 2018-11-14
Handle interrupt in interactive session

mallt:
- [x] All tests are passing

This PR adds handling of an interrupt signal in an interactive session.


bbatsov:
Very cool! 

I think we should also mention in the message the REPL prints on startup that now you can use `Control-d` and `Control-c` (same message REPL-y prints).

bbatsov:
Great work!

PR #83, opened 2018-11-16
fix some issues with hiccup style syntax in cljdoc.edn

martinklepsch:
Unfortunately the `verify-cljdoc-edn` step that's part of your CI doesn't yet catch syntax issues.

I thought just checking that the `:file` exists is sufficient but looks like we should expand this and ensure the value conforms to the spec.


bbatsov:
Thanks, Martin! Best customer service ever!

PR #86, opened 2018-11-21
[Fix #62] Capture output of background threads

### LINKED ISSUE #62: Investigate handling of background thread
Posting this issue here as a reference to [a CIDER issue](https://github.com/clojure-emacs/cider/issues/2474) where a background thread is assigned a different nrepl message id.

mallt:
- [x] All tests are passing

This PR alters the root binding of *out* and *msg* to be able to capture the output of background threads in responses and keep the :id of the original message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [#68](https://github.com/nrepl/nREPL/issues/68): Avoid illegal access warning on JDK 9+ caused by `nrepl.middleware.interruptible-eval/set-line!`.
 * [#77](https://github.com/nrepl/nREPL/issues/77): Exit cleanly after pressing `ctrl-d` in an interactive REPL.
 * [#13](https://github.com/nrepl/nREPL/issues/13): Catch ThreadDeath exception thrown by interrupt.
+* [#62](https://github.com/nrepl/nREPL/issues/62): Capture output of background threads.
 
 #### Changes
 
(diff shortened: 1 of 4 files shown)


mallt:
Thanks for your feedback!

> I am not familiar with this part of the code but airing a few concerns _just in case_.
> 
> Are you sure it should be `*out*` and not `out` from the `@bindings` above? Most likely they are the same, but who knows.
> 

I will change this indeed, it's definitely better to take the out from the `@bindings`.

> Output from java side is not going through `*out*`, right? So to completely solve the issues one would need to move cider-nrepl/out middleware logic here.
> 

I'm not sure about the java side. @bbatsov Do you have any thoughts about this?

> There is also `*err*` which should be taken care.
> 

Good point, I will check this.

> Can it be that two subsequent `evaluate` bind different `*out*`? If so, then you will still be loosing output from previous evals. In any case, it seems to me that we should refrain from resetting `*out*` root-var or `setOut` on every evaluate. Just doesn't feel right.
> 

I'm not sure about this either. Looking at the code the `evaluate` function seemed a logical place to put this.

> Finally, setting thread local value of _out_ instead of the root value would be achieving the same goal, no?

The reason for this change is an issue in CIDER where the printed output of a background thread got lost, a thread-local value of out wouldn't help in that case.

bbatsov:
The approach seems reasonable to me. This would basically render the `out` middleware redundant (which I always wanted us to achieve). 

Btw, this change won't mess up the regular output for the server, right? (e.g. some logging functionality)

I'd also love to ask @SevereOverfl0w and @arrdem to take a look here (if they can find some time, that is).

mallt:
I will close this PR and move the discussion to #62.

PR #87, opened 2018-11-24
Add verbose option to create-server to log in and out streams

### LINKED ISSUE #85: Implement message logging functionality

Often when people encounter an issue with nREPL the best way to debug the issues is to expect the message exchange between their client and nREPL, so we/they can check whether the request and response messages are what we're expecting them to be (both in their encoded and decoded form). 

Ideally we'll need some way to log all messages as Clojure maps and in their encoded format (e.g. bencode) to a file. Perhaps we can have a logging middleware that checks for some `:verbose` key in messages and would log those. Clients can easily implement a toggle turning on/off message logging.  We can probably have different options for showing the messages encoded or decoded. Likely we should have one log file per each server that goes to `.nrepl/` or the current directory, so logs for different servers don't get interleaved (or we can have one log file and just prefix messages accordingly there). 

**Describe alternatives you've considered**

CIDER has some client-side message logging functionality, but there's no easy way to inspect what happens with the message on the server's end. 

This doesn't really have to be a middleware - it can also be some server config (e.g. we can add a `verbose` flag to `start-server` and just check for it on the transport level), but I think a middleware will be a more flexible solution.  

Looking at #79 it seems that @pfeodrippe should be able to easy handle this. Given his work on #84 he should also immediately appreciate the importance of having this. :D

pfeodrippe:
There is a default-logger that can be changed by `logger op`,
it changes the state of `logger` when `verbose` is set.

Resolves #85.

- [x] All tests are passing

Thanks!


src/clojure/nrepl/misc.clj
```diff
@@ -46,3 +46,15 @@
                                                (-> session meta :id)
                                                session)}))]
     (merge basis response)))
+
+(defn resolve-ns-symbol
```
bbatsov: I'd rather keep this private in the ns that uses it, otherwise we're committing to it be part of the API. Maybe we can have a separate ns for useful stuff internally to nREPL that  we can say are not public API.

pfeodrippe: Didn't think about that, gonna add at server ns as private, thanks!

pfeodrippe:
> Can you also document this in the "Hacking" or "Troubleshooting" sections of the manual?

Added it, is it what you want? I can add more examples but I have to test it locally.
Maybe embedding nREPL in some application is the best way to test it as a real user? 
I would like to use `cider-connect`, but would have to hack Cider as well, adding the verbose key and stuff, right?

pfeodrippe:
Do I add code to cmdline, or would it preferably be made in a future PR?

pfeodrippe:
> > Can you also document this in the "Hacking" or "Troubleshooting" sections of the manual?
> 
> Added it, is it what you want? I can add more examples but I have to test it locally.
> 
> Maybe embedding nREPL in some application is the best way to test it as a real user?
> I would like to use `cider-connect`, but would have to hack Cider as well, adding the verbose key and stuff, right?

Could test it with Cider \o/, the verbose and logger works as expected

bbatsov:
> Added it, is it what you want? I can add more examples but I have to test it locally.

Yeah, that's what I meant. People need to know it's there and how to use it.

> Maybe embedding nREPL in some application is the best way to test it as a real user?

I guess most people would start the server via the command line, but some embedded examples will be fine as well. 

> I would like to use cider-connect, but would have to hack Cider as well, adding the verbose key and stuff, right?

Yeah, we should add some command to toggle the logging on and off there (maybe), provided the server is already running in verbose mode.

bbatsov:
@pfeodrippe Looking at the failing build it seems you used a string function available only in Clojure 1.8 (`starts-with?`). After replacing it we can merge this.

bbatsov:
> Also added logger tests for when fastlane dependency exists. Regarding it, we should put [nrepl/fastlane "0.1.0-SNAPSHOT"] in clojars, right?

Done.

src/clojure/nrepl/server.clj
```diff
@@ -34,12 +49,64 @@
...
+  (let [in (io/input-stream s)
+        out (io/output-stream s)
+        pin-out (PipedOutputStream.)
+        pin-in (PipedInputStream. pin-out)
+        pout-out (PipedOutputStream.)
+        pout-in (PipedInputStream. pout-out)]
+    (future
+      (try
+        (loop []
+          (let [buf (byte-array 8192)
```
bbatsov: Btw, is there are any particular reason for this size?

doc/modules/ROOT/pages/hacking_on_nrepl.adoc
```diff
@@ -40,3 +40,14 @@ specific version of Clojure like this:
...
 ----
+
+== Logging nREPL messages
+
+You can start the server with the verbose flag set to true, there's
+a default logger that logs the input and output messages with `log/info`,
+but a op `logger` is available so you can set the field `logger` to some
+namespaced symbol representing a function.
+
+See more at test `logger` at `nrepl.core-test`.
```
bbatsov: You might also include some sample output from the log, so users would know what to expect.

src/clojure/nrepl/server.clj
```diff
@@ -10,8 +12,21 @@
    [nrepl.misc :refer [log response-for returning]]
    [nrepl.transport :as t])
   (:import
+   [java.io PipedInputStream PipedOutputStream]
    [java.net InetAddress InetSocketAddress ServerSocket Socket SocketException]))
 
+(defn- resolve-ns-symbol
+  "Resolve a namespaced symbol to a var. Returns the var or nil if
+  the argument is nil or not resolvable."
+  [tag var-sym]
```
bbatsov: Btw, what do we need this tag for?

bbatsov:
Btw, this logs only the "encoded" data, right? I wonder how we can best combine some logging of both encoded and decoded messages (the request response maps).

src/clojure/nrepl/cmdline.clj
```diff
@@ -271,8 +272,10 @@
             handler (or handler (build-handler middleware))
             transport (some->> (:transport options) (require-and-resolve :transport))
             greeting-fn (if (= transport #'transport/tty) #'transport/tty-greeting)
+            verbose (:verbose options)
```
bbatsov: I guess we should also check if this was set in the config file here.

pfeodrippe: Gonna code it o/

src/clojure/nrepl/server.clj
pfeodrippe: Not really, has some number in mind?

src/clojure/nrepl/server.clj
pfeodrippe: Sorry, it was supposed to be generic to tell us what was trying to resolve if at fail, gonna remove it

pfeodrippe:
> Btw, this logs only the "encoded" data, right? I wonder how we can best combine some logging of both encoded and decoded messages (the request response maps).

Messages from the server to the client are not so difficult to reconcile adding more to the design a little bit, the problem is much more at input (communication from the client to the server) because it doesn't know where a message begins and ends (it depends on transport), but the client of nREPL could easily pass a function (`logger`) to read the data back as a stream and decode it using the appropriate decoder. 

Maybe we should let the future users of this feature to say what they want from all this verbosity and what's a pain.

And I would say that it resolves #85 partially, you can see what the server sees, but reconciling is hard.

bbatsov:
> Messages from the server to the client are not so difficult to reconcile adding more to the design a little bit, the problem is much more at input (communication from the client to the server) because it doesn't know where a message begins and ends (it depends on transport), but the client of nREPL could easily pass a function (logger) to read the data back as a stream and decode it using the appropriate decoder.

Isn't it much easier to log the incoming messages at the middleware level? They are basically the request map there. That's why originally I had suggested a logging middleware, before I realized this wasn't enough for the wire data.

pfeodrippe:
> > Messages from the server to the client are not so difficult to reconcile adding more to the design a little bit, the problem is much more at input (communication from the client to the server) because it doesn't know where a message begins and ends (it depends on transport), but the client of nREPL could easily pass a function (logger) to read the data back as a stream and decode it using the appropriate decoder.
> 
> Isn't it much easier to log the incoming messages at the middleware level? They are basically the request map there. That's why originally I had suggested a logging middleware, before I realized this wasn't enough for the wire data.

Yes, a middleware would work greatly with the request maps, both input and output. What I'm saying is that it's hard to know which bytes exactly generated a specific request map for the incoming message.

I'll add the same concepts of a `logger` function (with some other name) that the client can pass to this new middleware, it can't be the same one used for the wire data logging. Then the client would have freedom to do whatever it wants with the request map (append it to a file, send it to twitter etc) as well as with the wire data.

Sorry for the delay, these days are harder, but at the weekend I'll have more time to deal with the PRs here o/

bbatsov:
Btw, one more update - we had to remove the tools.logging dep due to #89, so now we also have to add some simple logging to nREPL itself - probably some namespace that provides basic logging functions that dump stuff to stdout/err and uses `tools.logging` if available.

src/clojure/nrepl/cmdline.clj
pfeodrippe: It already works because `options` is already merged, but I've added it to the docs o/

pfeodrippe:
Created `nrepl.logging` and added a middleware to handle the decoded messages

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
   function in `pr-values`, enabling pretty-printed REPL results.
 * [#66](https://github.com/nrepl/nrepl/pull/66): Add support for a global and local configuration file.
 * [#63](https://github.com/nrepl/nrepl/issues/63): Make it possible to specify the transport via the command-line client (`--transport/-t`).
+* [#85](https://github.com/nrepl/nrepl/issues/85): Implement message logging functionality.
```
bbatsov: This has to be moved to the `master` section.

src/clojure/nrepl/cmdline.clj
bbatsov: Ah, excellent!

src/clojure/nrepl/logging.clj
```diff
@@ -0,0 +1,16 @@
...
+(def default-output *out*)
+
+(defn info
+  [& args]
+  (binding [*out* default-output]
+    (apply println "INFO: " args)))
+
+(defn warn
+  [& args]
+  (binding [*out* default-output]
```
bbatsov: Warn should probably output to `err`, same as `error`.

src/clojure/nrepl/logging.clj
```diff
@@ -0,0 +1,16 @@
+(ns nrepl.logging
+  "Logging to be used at nREPL"
+  {:added  "0.6.0"})
+
+;; TODO: Use clojure.tools.logging if available
```
bbatsov: Seems to me the easiest way to do this would be to check if tools.logging is available, and defer in our to functions to its functions if that's the case. If not ours just dump stuff to stdout and stderr. I guess you're working in this direction already.

src/clojure/nrepl/server.clj
bbatsov: Ah, no. Just seemed somewhat random and I decided to ask.

bbatsov:
I think we're getting pretty close to shipping this. Right now my only concern is the usefulness of raw byte arrays in the log, as I definitely envisioned this as something higher-level (e.g. the exact bencode/json message) sent or received. 

As we know that the messages are serialized within the scope of a session it seems reasonable to just read them until we get a complete map and then display it in its "native" format. How are you using the byte-array logs, btw? Do you just copy/paste them in the REPL and decode them there?

pfeodrippe:
> I think we're getting pretty close to shipping this. Right now my only concern is the usefulness of raw byte arrays in the log, as I definitely envisioned this as something higher-level (e.g. the exact bencode/json message) sent or received.
> 
> As we know that the messages are serialized within the scope of a session it seems reasonable to just read them until we get a complete map and then display it in its "native" format. How are you using the byte-array logs, btw? Do you just copy/paste them in the REPL and decode them there?

I log them in byte-array format because there are transports that does not map one byte to one char (as `transit+msgpack`) and, as another problem, transports like `transit+json` or `transit+json-verbose` come in multiple chunks of bytes to build a message, the input stream given to their internal readers just read them until they get a complete message, it's opaque, it appears to not have a good way to match encoded with decoded in these cases.

The loggers are, by now, independents, the encoded reads from the input and output streams and the decoded one is in a simple middleware.

bbatsov:
> I log them in byte-array format because there are transports that does not map one byte to one char (as transit+msgpack) and, as another problem, transports like transit+json or transit+json-verbose come in multiple chunks of bytes to build a message, the input stream given to their internal readers just read them until they get a complete message, it's opaque, it appears to not have a good way to match encoded with decoded in these cases.

Yeah, but we still know the encoding scheme for them, which was my point. Unless I'm missing something.

> The loggers are, by now, independents, the encoded reads from the input and output streams and the decoded one is in a simple middleware.

Yeah, I saw this. The middleware is great. I just wanted somehow to be able to monitor what bencode/json/whatever/messages are passed back and forth. That's my only concern. The middleware is pretty useful as the decoded maps help a lot, but for the byte-arrays I simply don't see much value for day-to-day development. That's why I asked how exactly were you using those to get an idea how they were valuable for you.

pfeodrippe:
Let me know if I'm understanding you right, you're saying that it's better to log the encoded data stringified (if possible) byte array (I agree).

Going in another way, maybe we shouldn't do anything if the `logger` function is not resolved, so the client can decide what it wants to do with the data and, different of what happens at the current moment, you wouldn't get the (maybe) unwanted logs you have from the time you start the server until you set `logger`.

bbatsov:
> Let me know if I'm understanding you right, you're saying that it's better to log the encoded data stringified (if possible) byte array (I agree).

Yep.

> Going in another way, maybe we shouldn't do anything if the logger function is not resolved, so the client can decide what it wants to do with the data and, different of what happens at the current moment, you wouldn't get the (maybe) unwanted logs you have from the time you start the server until you set logger.

Yeah, probably that'd be best, although I guess we can have logging functions supplied with the respective transports and just use some multimethod to select them, in case the users don't supply a custom function themselves.

pfeodrippe:
Instead of multimethods, now the client can pass at initialization a `default-logger-behaviour` option to set how it wants the default logger to log (silenced, encoded as string or encoded as bytes), it gives to the user complete control and it's independent of the transport

bbatsov:
I'll try to review the current version later today. I'd also love for @cgrand to take a look and share his perspective of the approach.

src/clojure/nrepl/server.clj
```diff
@@ -34,16 +55,76 @@
     (catch java.io.IOException e
       (log e "Failed to close " x))))
 
+(defn- default-bytes-logger
+  [channel ba]
+  (log/info channel (map byte ba)))
+
+(defn- default-string-logger
+  [channel ba]
+  (log/info channel (str/join (map char ba))))
```
bbatsov: Don't you need the encoding scheme of the transport to try to get some meaningful string out of this?

bbatsov:
Btw, because I'm not convinced of the usefulness of the wire logging in its current form maybe we can simplify the current PR to feature just the message logging on the middleware level and revisit the rest independently? 

But before doing anything in this direction let's see what @cgrand will say. Btw, @SevereOverfl0w any thoughts on this from you? What logging would be most useful for your use-cases?

src/clojure/nrepl/server.clj
pfeodrippe: You don't need to use the transport as the current transports have the bytes representing chars, it just sniffs the socket independent of the transport, we certainly need comments of other people here to see the usefulness of it or if they need other ways to do it.

pfeodrippe:
> Btw, because I'm not convinced of the usefulness of the wire logging in its current form maybe we can simplify the current PR to feature just the message logging on the middleware level and revisit the rest independently?

Yes, I'm foud of the smallest PRs possible and starting with the easiest ones (the request map logging).

> But before doing anything in this direction let's see what @cgrand will say. Btw, @SevereOverfl0w any thoughts on this from you? What logging would be most useful for your use-cases?

I totally agree with them helping out! As it was one of my first PRs here and I've learned a ton of stuff, maybe it's not optimized enough and I feel the code could be much less hacky than it's now. 
I'm gonna wait for their comments and I'd would be happy with a totally new design or even closing this PR in favour of a better approach =)

bbatsov:
I know @cgrand commented here, but now I don't see where his comment went. Anyways, it seems that there's consensus we should reduce the scope of this PR to logging the request/response maps and we'll revisit logging the raw data down the road.

pfeodrippe:
@bbatsov His comments at https://github.com/nrepl/nrepl/issues/85?

Gonna check this again after the holidays =)

bbatsov:
> @bbatsov His comments at #85?

Ah, yeah!

> Gonna check this again after the holidays =)

Roger that! Happy holidays! 🎄

pfeodrippe:
> Roger that! Happy holidays!

Thanks! That it be happy for you too! \o/ 
Um brinde de Recife

pfeodrippe:
I'll tackle on this on this weekend, so we just have to keep the request map interceptors, right, excluding the raw data logging at input and output streams?

bbatsov:
Yep. At this stage let's keep just the request/response maps logging.

bbatsov:
Ping :-)

pfeodrippe:
I was wondering if only logging the response maps would be useful for the client, maybe we should close this and open other PR in the future with this if really needed?

pfeodrippe:
Gonna close this so we can focus in other stuff o/

bbatsov:
👍 Thanks for all the work you've put into this! I'm really sorry we didn't get to anything merged in the end, but I think this was a valuable exercise never-the-less.

Issue #37, opened 2018-07-26
Convert the ops middleware docs generation logic into a lein task


bbatsov:
There's some docs generation logic here https://github.com/nrepl/nREPL/blob/master/src/clojure/nrepl/middleware.clj#L204

It's used to generate https://github.com/nrepl/nREPL/blob/master/doc/ops.md

Now it's pretty cumbersome to regenerate the ops document and some lein task would be a good step forward.

bbatsov:
On second thought - we should probably just an `--describe-ops` command line flag, as that'd be simpler and more generic. 

When extra middleware is loaded (e.g. cider-nrepl), we'd be able to generate documentation for it as well.

SevereOverfl0w:
Even simpler than a lein task, write a main and use `lein run`

PR #94, opened 2018-12-09
Add ops.adoc generation script behind `lein docs`

benjaminran:
Putting this up here in the hopes that someone else will point out my mistake before I find it myself or offer guidance (on approach, etc.). This adds an nrepl.-private/-main routine to regenerate ops.adoc. The docs helper functions were moved verbatim from middleware.clj into the new namespace, and the new -main sends a verbose "describe" request through the default handler via a simple transport.

The issue I'm currently debugging: when wrap-describe is executed on the verbose "describe" message the -main is sending, :descriptors isn't set on the message. It looks like the handler function is supposed to be composed with wrap-conj-descriptor (which sets :descriptors for describe messages) in set-descriptor!, but for some reason that doesn't appear to be happening. Maybe I set something up wrong in the handler, but I'm still trying to figure out what. Let me know if any pointers.

The result is an ops.adoc without ops 😢 :
```
bnran at dca90496b22a in ~/dev/clj/nREPL on master*
$ lein run -m nrepl.-private.docs && cat doc/modules/ROOT/pages/ops.adoc
done
////
This file is _generated_ by #'nrepl.-private.docs/-main
   *Do not edit!*
////
= Supported nREPL operations

[small]#generated from a verbose 'describe' response (nREPL v0.6.0-SNAPSHOT)#

== Operationsbnran at dca90496b22a in ~/dev/clj/nREPL on master*
$ 
```

As for the general approach, it would be nice to avoid including this in the released jar used to run other projects' repls. I guess that's what a lein task is (still new to clojure) but I'm having trouble finding docs on how to create one. Thoughts on how important this actually is would be welcome.

P.S. also wondering who sets system prop 'nrepl.basedir'.

Issue: https://github.com/nrepl/nrepl/issues/37

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -13,7 +13,8 @@
   :dependencies [[nrepl/bencode "1.0.0"]]
 
   :aliases {"bump-version" ["change" "version" "leiningen.release/bump-version"]
-            "test-all" ["with-profile" "+1.7:+1.8:+1.9:+fastlane" "test"]}
+            "test-all" ["with-profile" "+1.7:+1.8:+1.9:+fastlane" "test"]
+            "docs" ["with-profile" "+maint" "run" "-m" "nrepl.impl.docs"]}
 
   :release-tasks [["vcs" "assert-committed"]
                   ["bump-version" "release"]
@@ -38,6 +39,7 @@
                       :dependencies [[org.clojure/clojure "1.10.0-master-SNAPSHOT"]]}
 
              :sysutils {:plugins [[lein-sysutils "0.2.0"]]}
+             :maint {:source-paths ["src/maint"]}
 
              ;; CI tools
              :cloverage {:plugins [[lein-cloverage "1.0.13"]]}
--- a/project.clj
+++ b/project.clj
@@ -14,7 +14,7 @@
 
   :aliases {"bump-version" ["change" "version" "leiningen.release/bump-version"]
             "test-all" ["with-profile" "+1.7:+1.8:+1.9:+fastlane" "test"]
-            "docs" ["with-profile" "+maint" "run" "-m" "nrepl.impl.docs"]}
+            "docs" ["with-profile" "+maint" "run" "-m" "nrepl.impl.docs" "--lein"]}
 
   :release-tasks [["vcs" "assert-committed"]
                   ["bump-version" "release"]
@@ -39,7 +39,8 @@
                       :dependencies [[org.clojure/clojure "1.10.0-master-SNAPSHOT"]]}
 
              :sysutils {:plugins [[lein-sysutils "0.2.0"]]}
-             :maint {:source-paths ["src/maint"]}
+             :maint {:source-paths ["src/maint"]
+                     :dependencies [[org.clojure/tools.cli "0.4.1"]]}
 
              ;; CI tools
              :cloverage {:plugins [[lein-cloverage "1.0.13"]]}
--- a/project.clj
+++ b/project.clj
@@ -14,7 +14,9 @@
 
   :aliases {"bump-version" ["change" "version" "leiningen.release/bump-version"]
             "test-all" ["with-profile" "+1.7:+1.8:+1.9:+fastlane" "test"]
-            "docs" ["with-profile" "+maint" "run" "-m" "nrepl.impl.docs" "--lein"]}
+            "docs" ["with-profile" "+maint" "run" "-m" "nrepl.impl.docs" "--file"
+                    ~(clojure.java.io/as-relative-path
+                      (clojure.java.io/file "doc" "modules" "ROOT" "pages" "ops.adoc"))]}
 
   :release-tasks [["vcs" "assert-committed"]
                   ["bump-version" "release"]
--- a/src/maint/nrepl/impl/docs.clj
+++ b/src/maint/nrepl/impl/docs.clj
@@ -1,4 +1,5 @@
 (ns nrepl.impl.docs
+  "Doc generation utilities"
   (:require
    [clojure.tools.cli :as cli]
    [clojure.java.io :as io]
@@ -10,21 +11,21 @@
   (:import
    (java.io File)))
 
+(declare -main)
+
 (def cli-options
   [["-f" "--file FILE" "File to write output into (defaults to standard output)"
     :parse-fn #(File. %) :default *out*]
    ["-h" "--help" "Show this help message"]
-   ["-l" "--lein"]
    ["-o" "--output OUT" "One of: raw, adoc, md" :default "adoc"
     :validate [#(contains? #{"raw" "adoc" "md"} %)]]])
-;; io/file (File. (System/getProperty "nrepl.basedir" ".")) "doc" "modules" "ROOT" "pages" "ops.adoc")
 
 (defn- exit [status msg]
   (println msg)
   (System/exit status))
 
 (defn- usage [options-summary]
-  (->> ["Regenerate and output the ops documentation to the specified destination in the specified format."
+  (->> [(:doc (meta #'-main))
         ""
         "Usage: lein -m +maint run nrepl.impl.docs [options]"
         ""
@@ -101,10 +102,8 @@ use in e.g. wiki pages, github, etc."
                 (message-slot-adoc (sort returns))
                 "\n"))))
 
-(declare -main)
-
 (defn- format-response [format resp]
-  (cond (= format "raw") (pr-str resp)
+  (cond (= format "raw") (pr-str (select-keys resp [:ops :versions]))
         (= format "md") (str "<!-- This file is *generated* by " #'-main
                              "\n   **Do not edit!** -->\n"
                              (describe-markdown resp))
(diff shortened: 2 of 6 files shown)


SevereOverfl0w:
fwiw, I've never seen a namespace start with `-`. This might be something that coincidentally works, but isn't supported.

I generally see `.impl` to indicate a namespace isn't part of the api, so `nrepl.impl.docs` or something.

SevereOverfl0w:
> As for the general approach, it would be nice to avoid including this in the released jar used to run other projects' repls. I guess that's what a lein task is (still new to clojure) but I'm having trouble finding docs on how to create one. Thoughts on how important this actually is would be welcome.

Create an additional directory to house this namespace, something like `mains`. So you'd have `mains/nrepl/impl/docs.clj`. Then the project.clj can be updated to include this when running the main / exclude this when running the build.

SevereOverfl0w:
I just loaded this up in a REPL, the issue is your FnTransport. You've put `print t` and you meant `(print t)`.

You will want to have side effects in order to get a return value. I came up with this:

```
(defn- collect-responses
  [handler op]
  (let [res (atom [])]
    (handler (assoc op
                    :transport
                    (transport/fn-transport
                      (fn [t] t)
                      (fn [m] (swap! res conj m))
                      (fn []))))
    @res))

(comment
  (collect-responses
    (server/default-handler)
    {:op "describe"
     :verbose? true
     :id "1"}))
```

The invocation in the `comment` gives me back a vector of results :smile:

benjaminran:
Thanks @SevereOverfl0w for your help. The latest revision adds a maint profile/source-dir to hold codebase-maintenance-related scripts, like nrepl.impl.docs. The ops.adoc was generated with `$ lein docs`. Removing 'WIP' from the title.

src/maint/nrepl/impl/docs.clj
```diff
@@ -0,0 +1,99 @@
+(ns nrepl.impl.docs
+  (:refer-clojure :exclude [comparator])
```
bbatsov: I think you've copied this here accidentally.

benjaminran: yeah, thanks

bbatsov:
Looks good to me overall. Just a couple of minor remarks:

* now there's no easy way to generate this in markdown format. Might be good to add some basic command-line flag about the output format

* now there's also no easy way to just dump the generated output to the console, so you can redirect it wherever. That might be handy for people looking to generate some documentation for themselves - e.g. just filter out middleware they are working on or whatever.

bbatsov:
I think there were also a few references in the manual that you'll have to update that mentioned the old var names.

benjaminran:
Looking through the manual...

bbatsov:
See https://nrepl.xyz/nrepl/design/middleware.html

benjaminran:
Updated middleware.adoc and factored out a `generate-ops-info` function to return the describe response map out of -main, which now does command line parsing and output. Output can be written in md, adoc, or pr-str'd and can be written file or stdout.

Let me know if you want a changelog entry (don't expect it) or tests for this.

Some sample cli interactions:
```
bnran at dca90496b22a in ~/dev/clj/nREPL on master▲
$ lein with-profile +maint run -m nrepl.impl.docs --help
Regenerate and output the ops documentation to the specified destination in the specified format.

Usage: lein -m +maint run nrepl.impl.docs [options]

Options:
  -f, --file FILE   java.io.OutputStreamWriter@3b05a99b  File to write output into (defaults to standard output)
  -h, --help                                             Show this help message
  -l, --lein
  -o, --output OUT  adoc                                 One of: raw, adoc, md
bnran at dca90496b22a in ~/dev/clj/nREPL on master▲
$ lein with-profile +maint run -m nrepl.impl.docs --file /tmp/raw.txt --output raw
Regenerated /tmp/raw.txt
bnran at dca90496b22a in ~/dev/clj/nREPL on master▲
$ lein with-profile +maint run -m nrepl.impl.docs --file /tmp/ops.md --output raw
Regenerated /tmp/ops.md
bnran at dca90496b22a in ~/dev/clj/nREPL on master▲
$ lein docs
Regenerated ./doc/modules/ROOT/pages/ops.adoc
```

src/maint/nrepl/impl/docs.clj
```diff
@@ -0,0 +1,146 @@
+(ns nrepl.impl.docs
```
bbatsov: I guess you can add some docstring overview of the purpose of the ns. That's always useful for people exploring the code.

src/maint/nrepl/impl/docs.clj
```diff
@@ -0,0 +1,146 @@
...
+   (java.io File)))
+
+(def cli-options
+  [["-f" "--file FILE" "File to write output into (defaults to standard output)"
+    :parse-fn #(File. %) :default *out*]
+   ["-h" "--help" "Show this help message"]
+   ["-l" "--lein"]
+   ["-o" "--output OUT" "One of: raw, adoc, md" :default "adoc"
+    :validate [#(contains? #{"raw" "adoc" "md"} %)]]])
+;; io/file (File. (System/getProperty "nrepl.basedir" ".")) "doc" "modules" "ROOT" "pages" "ops.adoc")
```
bbatsov: Seems you forgot to remove this.

bbatsov:
Overall - it looks good to me. 

What's `raw`, btw? Is that EDN?

src/maint/nrepl/impl/docs.clj
```diff
@@ -0,0 +1,146 @@
...
+   [nrepl.transport :as transport]
+   [nrepl.version :as version])
+  (:import
+   (java.io File)))
+
+(def cli-options
+  [["-f" "--file FILE" "File to write output into (defaults to standard output)"
+    :parse-fn #(File. %) :default *out*]
+   ["-h" "--help" "Show this help message"]
+   ["-l" "--lein"]
```
bbatsov: This command flag seems weird. Why is it needed?

project.clj
```diff
@@ -13,7 +13,8 @@
   :dependencies [[nrepl/bencode "1.0.0"]]
 
   :aliases {"bump-version" ["change" "version" "leiningen.release/bump-version"]
-            "test-all" ["with-profile" "+1.7:+1.8:+1.9:+fastlane" "test"]}
+            "test-all" ["with-profile" "+1.7:+1.8:+1.9:+fastlane" "test"]
+            "docs" ["with-profile" "+maint" "run" "-m" "nrepl.impl.docs" "--lein"]}
```
bbatsov: I saw this `--lein` here and in the code, but it doesn't seem to be used anywhere, so I assume it's some oversight.

project.clj
```diff
@@ -38,6 +39,8 @@
                       :dependencies [[org.clojure/clojure "1.10.0-master-SNAPSHOT"]]}
 
              :sysutils {:plugins [[lein-sysutils "0.2.0"]]}
+             :maint {:source-paths ["src/maint"]
```
bbatsov: maint stands for maintenance, right?

project.clj
benjaminran: It's intentional, but there may be a better way to achieve what I want. I want `lein docs` to write to nREPL/doc/modules/ROOT/pages/op.adoc but I want lein with-profile +maint run nrepl.impl.docs to print to stdout by default.

The `--lein` flag is in the 'lein docs' alias definition then -main uses that as a short-circuit flag to write the output to the existing ops.adoc path instead of considering the `--file` param. This isn't very straightforward :/

project.clj
benjaminran: yep

benjaminran:
> What's raw, btw? Is that EDN?
Raw is just the output of `pr` on the verbose describe result, so output is printed 'in a way that objects can be read by the reader'. I should actually filter that to exclude keys other that 'ops' and 'versions' (or maybe just 'ops?). Will update.

project.clj
benjaminran: tbh didn't actually actually check if -main is invoke with cwd = the project root or if it's actually the cwd where lein is invoked from--assumed the latter. Will check....

bbatsov: I see. I think think that `--lein` is a pretty confusing flag name, so it should probably be something else, but I don't have an amazing idea for this either.

benjaminran: It turns out cwd is always the project root when calling a function via `lein run`, so in my latest revision I did away with `--lein` and just use `--file` with the relative path from the project root in the lein alias definition.

src/maint/nrepl/impl/docs.clj
benjaminran: Added in latest revision

benjaminran:
The latest revision addresses your comments so far by removing the --lein flag and adding a simple namespace docstring (let me know if you think I should say more in the ns docstring). I also filter the keys written in the raw output format to just include :ops to be consistent with the info included in the adoc and md output formats.

bbatsov:
I think the code is in a good enough shape for now and we can obviously keep tuning! 
Thanks for working on this! I really appreciate it! 🙇

benjaminran:
Thanks! Opened https://github.com/nrepl/nrepl/pull/104 to fix an accidentally included function.

PR #98, opened 2018-12-12
Session thread

cgrand:
Fix for #16:
Each registered session now has its own thread for `:eval`. Upon successful `interrupt` this thread is replaced. So don't interrupt when you care about thread identity.

### DIFF
--- a/src/clojure/nrepl/middleware/session.clj
+++ b/src/clojure/nrepl/middleware/session.clj
@@ -11,7 +11,8 @@
   (:import
    clojure.lang.LineNumberingPushbackReader
    [java.io PrintWriter Reader Writer]
-   java.util.concurrent.LinkedBlockingQueue))
+   java.util.concurrent.atomic.AtomicLong
+   [java.util.concurrent LinkedBlockingQueue BlockingQueue Executor SynchronousQueue ThreadFactory ThreadPoolExecutor TimeUnit]))
 
 (def ^{:private true} sessions (atom {}))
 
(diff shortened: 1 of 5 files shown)


src/clojure/nrepl/middleware/session.clj
```diff
@@ -24,6 +25,40 @@
...
+           queue (SynchronousQueue.)}}]
+  (let [^ThreadFactory thread-factory (or thread-factory (configure-thread-factory))]
+    (ThreadPoolExecutor. 1 Integer/MAX_VALUE
+                         (long 30000) TimeUnit/MILLISECONDS
+                         ^BlockingQueue queue
+                         thread-factory)))
+
+(def default-executor (delay (configure-executor)))
+
+(defn default-exec [id ^Runnable thunk ^Runnable ack]
```
bbatsov: I guess we can add some docstring for this as its public API.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -137,14 +172,51 @@
          (atom (merge baseline-bindings (dissoc (get-thread-bindings) #'*agent*))
                :meta {:id id
                       :stdin-reader stdin-reader
-                      :input-queue input-queue}))))))
+                      :input-queue input-queue
+                      :exec default-exec}))))))
+
+(defn session-exec [id]
```
bbatsov: Same here.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -137,14 +172,51 @@
...
+             (.put queue [exec-id r ack]))}))
 
 (defn- register-session
   "Registers a new session containing the baseline bindings contained in the
    given message's :session."
   [{:keys [session transport] :as msg}]
   (let [session (create-session transport @session)
-        id (-> session meta :id)]
+        {:keys [id]} (meta session)]
+    (alter-meta! session into (session-exec id))
```
bbatsov: I wonder if we shouldn't tackle this in `create-session`, but it's not a big deal.

test/clojure/nrepl/core_test.clj
```diff
@@ -358,7 +358,7 @@
                                                        (def halted? false)))})]
     (Thread/sleep 100)
     (is (= #{"done"} (-> session (message {:op :interrupt}) first :status set)))
-    (is (= #{"done" "interrupted"} (-> resp combine-responses :status)))
+    (is (= #{} (reduce disj #{"done" "interrupted"} (-> resp combine-responses :status))))
```
bbatsov: What can be the extra status?

bbatsov:
The code looks great overall! Just a couple of small remarks before we proceed with merging it:

* make sure that https://nrepl.xyz/nrepl/usage/misc.html#_hot_loading_dependencies is not affected by the changes
* Please, mention in the docs on sessions (https://nrepl.xyz/nrepl/design/middleware.html#_sessions) that they are backing the evaluation thread and perhaps that things we discussed on Slack about the serialization of eval ops vs all other ops. I guess that'd be pretty useful to implementors. Also you can add there some note that people who were using `default-executor` should use the new API.

src/clojure/nrepl/middleware/session.clj
cgrand: How so? In `create-session` we don't know yet if the session is going to be long-lived (what's the current practice for client implementors? 1 long-lived session for all interactions?) that's why `create-session` defaults to threadpool execution.

Oh! Maybe you meant to pass a flag to `create-session`, is this it?

test/clojure/nrepl/core_test.clj
cgrand: An `eval-error` managed to sneak in. (Because before stopping I first interrupt the thread)

src/clojure/nrepl/middleware/session.clj
cgrand: Is `clone` ever used to fork a new long-lived session from another long-lived session?

bbatsov: The option certainly exists. Don't know how many people make use of it. I know some people were complaining that cloning a long-lived session didn't copy the dynamic vars from the original session (which has been fixed since then), so I guess someone does this.

Issue #100, opened 2018-12-14
Bindings and sessions


cgrand:
# Why do we mess with bindings?

Sessions initial purpose is to remember the dynamic bindings between call to `:eval`. In a plain repl, there's no need for them because 1 connection == 1 session == 1 thread.

Since nrepl multiplexes sessions over a single connection a session id is required to demux them and route them to the repl threads.

With a 1 session == 1 thread mapping there would still be no need for saving bindings. However interrupting an eval entails killing its thread. That's why we need to save bindings so as to able to create a new thread with the same bindings (the illusion is not perfect because thread identity and other ThreadLocals have changed).

Additionally it allows other ops to inspect the dynamic scope (modifying it is not a good idea: you have to serialize with `: eval` to be sure your changes are not overwritten).

# Existing lifecycle

There are two places where dynamic scope is handled: `:clone` (session creation) and `:eval`. I believe it's one place too many: `:clone` should defer to `:eval`.

`:eval` starts by setting the sessions bindings then starts `clojure.main/repl` which sets defaults bindings and then further customization is done in `:init`. After each print bindings are saved.

# Proposed changes

## clone should defer to eval

So as not to have to maintain two ways to capture scope.

## eval bindings cleanup

I think this should be simplified: we shouldn't set bindings to have them erased by `clojure.ùain/repl` and then retweak them.

Additionally I believe `:prompt` is a slightly better place

## Remove special handlings of `*agent*`

nrepl used to be agent-based but it is not anymore.

PR #101, opened 2018-12-14
Rework bindings management (see #100)

cgrand:


src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -47,88 +47,65 @@
...
-                                                              :root-ex (-> root-ex class str)}))
-                         (clojure.main/repl-caught e)))))
-          (finally
-            (.flush ^Writer out)
-            (.flush ^Writer err)))))
-    (maybe-restore-original-ns @bindings)))
+      (try
+        (clojure.main/repl
+         :eval (if eval (find-var (symbol eval)) clojure.core/eval)
+         ;; clojure.main/repl paves over certain vars even if they're already thread-bound
```
bbatsov: Btw, do you know what exactly is paved over by cloure.main/repl?

src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -47,88 +47,65 @@
...
+                          (cond-> explicit-ns (assoc #'*ns* explicit-ns)
+                                  file (assoc #'*file* file)))]
+                  (pop-thread-bindings)
+                  (push-thread-bindings bindings))
+         :read (if (string? code)
+                 (let [reader (source-logging-pushback-reader code line column)]
+                   #(read {:read-cond :allow :eof %2} reader))
+                 (let [code (.iterator ^Iterable code)]
+                   #(or (and (.hasNext code) (.next code)) %2)))
+         :prompt #(reset! session (maybe-restore-original-ns (capture-thread-bindings)))
```
bbatsov: That's called after init, right?

src/clojure/nrepl/middleware/session.clj
```diff
@@ -156,31 +156,27 @@
...
+   (let [id (uuid)
+         out (session-out :out id transport)
+         {:keys [input-queue stdin-reader]} (session-in id transport)
+         session (atom (into baseline-bindings
+                             {#'*out* out
+                              #'*err* (session-out :err id transport)
+                              #'*in* stdin-reader
+                              #'*ns* (create-ns 'user)
+                              #'*out-limit* (or (baseline-bindings #'*out-limit*) 1024)
+                          ;; clojure.test captures *out* at load-time, so we need to make sure
```
bbatsov: This comment seems misaligned.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -156,31 +156,27 @@
...
+                          ;; runtime output of test status/results is redirected properly
+                          ;; TODO: is this something we need to consider in general, or is this
+                          ;; specific hack reasonable?
+                              #'clojure.test/*test-out* out})
```
bbatsov: Btw, any thoughts on this hack at all? Perhaps there's a more generic way to approach this.

bbatsov:
That was fast! Apart from my small remarks everything looks good. 👍

src/clojure/nrepl/middleware/interruptible_eval.clj
cgrand: I should have removed this comment, since it's now irrelevant. Look here for the list:
See here https://github.com/clojure/clojure/blob/master/src/clj/clojure/main.clj#L81

It's irrelevant because of the get/pop/push sequence which overwrites them with session values.

So if Clojure 1.11 adds a new dynvar we get the default value on session creation (dummy evaluate) and after that we force the session value in init, so we are in control.

src/clojure/nrepl/middleware/interruptible_eval.clj
cgrand: it's called after init and before read, it means it's called before EOF too.
As a bonus we can leave the `*N` vars management to clojure.main/repl!

src/clojure/nrepl/middleware/session.clj
cgrand: I haven't thought about it at all

src/clojure/nrepl/middleware/interruptible_eval.clj
bbatsov: Got it. Thanks for the explanation!

bbatsov:
🙇 🎆

PR #102, opened 2018-12-15
Fix `run-repl` using a different transport

pfeodrippe:
When testing with fastlane transports, the REPL was being
started using bencode and was giving an "The transport's socket
appears to have lost its connection to the nREPL server" error.

We add a `transport` key to `run-repl` args so it can be
connect to the server using the right transport.

- [ ] All tests are passing

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ### master (unreleased)
 
-### Changes
+#### Bugs fixed
+
+* When starting the REPL, make sure the transport option is used correctly.
+
+#### Changes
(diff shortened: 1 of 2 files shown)


bbatsov:
Nice catch!

Given the current complexity of the built-in UI probably we should consider finally writing a few unit tests for it. :-)

pfeodrippe:
Yes, I was starting to tackle with the cmdline tests locally, but no success until now, it hangs at some point just before connecting to some running server at `nrepl.ack/send-ack`.

PR #103, opened 2018-12-15
Fixes nREPL initialization with `:ack` port at cmd line

pfeodrippe:
Start nREPL with a ack port from the command line
raises a "java.util.IllegalFormatConversionException"
exception because the `format` expects a integer.

This fixes it by converting to int if `:ack` field
is given.

- [ ] All tests are passing

Thanks!

### DIFF
--- a/src/clojure/nrepl/cmdline.clj
+++ b/src/clojure/nrepl/cmdline.clj
@@ -269,7 +269,7 @@
             greeting-fn (if (= transport #'transport/tty) #'transport/tty-greeting)
             server (start-server :port port :bind bind :handler handler
                                  :transport-fn transport :greeting-fn greeting-fn)]
-        (when-let [ack-port (:ack options)]
+        (when-let [ack-port (some-> (:ack options) ->int)]
           (binding [*out* *err*]
             (println (format "ack'ing my port %d to other server running on port %d"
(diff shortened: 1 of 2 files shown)


src/clojure/nrepl/cmdline.clj
```diff
@@ -269,7 +269,7 @@
             greeting-fn (if (= transport #'transport/tty) #'transport/tty-greeting)
             server (start-server :port port :bind bind :handler handler
                                  :transport-fn transport :greeting-fn greeting-fn)]
-        (when-let [ack-port (:ack options)]
+        (when-let [ack-port (some-> (:ack options) ->int)]
```
bbatsov: I guess this function is safe to call on a real integer, right?

pfeodrippe: Yes, it checks if it's a number before parsing, it's used for `:port` param

bbatsov: 👍

bbatsov:
Thanks for tackling this!

bbatsov:
One more argument for some cmdline tests I guess. 😉

pfeodrippe:
Thanks for merging with the master o/

PR #104, opened 2018-12-16
Remove unused function (#37)

benjaminran:
Removing an unused function which was accidentally included in https://github.com/nrepl/nrepl/pull/94.


bbatsov:
👍

PR #106, opened 2018-12-17
Rich printing

### LINKED ISSUE #92: Elisions and other rich printing
The unrepl printer provides elision: large locations are truncated and a *cough* hypermedia affordance *cough* is put where truncation happened; this hyperlink of sorts contains a regular form that you just have to evaluate to get back the remaining part... which itself can be elided if still too large. So it's related to #75.

Clients can render elisions as buttons and, on click, the button is replaced with more data. In practice it becomes a generic browsing facility (and Clojure 1.10 datafy/nav is going to make things even easier.) For example, unrepl printing of a random Java object contains a beans map of the object -- except this map is completely elided and will be created only when the user decides to navigate the object graph.

The catch is that an EDN representation of clojure values is returned. For example `#"abc"` is printed as `#unrepl/pattern "abc"` because regex patterns are not EDN. Same goes for rationals. The EDN representation allows to disambiguate between the printing of symbols and classes.

So it definitely adds some rendering complexity to the client, and cause one to reconsider how to do pretty printing when more data may appear in the middle of already pretty printed stuff (how to best reflow a pretty print?)

The best integration course seems tp make this an optional behaviour. You would opt for it by setting a custom `:printer`. 

The unrepl printer emits edn representations of clojure data

**Describe alternatives you've considered**

**Additional context**

Add any other context about the feature request here.

cgrand:
Fix #92.

Enable rich printing by setting `:printer` to `nrepl.elisions/printer` in an `:eval` message.

Long/infinite data, MIME attachments and `datafy`/`nav` are supported through elisions and custom literals.

Pretty printing has to be done on the client side.


src/clojure/nrepl/elisions.clj
```diff
@@ -0,0 +1,44 @@
+(ns nrepl.elisions
```
bbatsov: I wonder if we shouldn't put this namespace and rich-printer under something like `nrepl.print` or whatever. Might be useful to structure the code like this if we ever come up with new printers.

bbatsov: Some short summary of the purpose of the ns and add `:added 0.6` metadata would be a nice addition as well.

src/clojure/nrepl/rich_printer.clj
```diff
@@ -0,0 +1,460 @@
...
+  Features elisions, MIME attachments and general purpose object browsing.
+  Initially developed for unrepl."
+  (:refer-clojure :exclude [pr pr-str])
+  (:require
+   [clojure.core :as core]
+   [clojure.string :as str]
+   [clojure.edn :as edn]
+   [clojure.main :as main]))
+
+; even with limits on breadth and depth of printing, we can easily get big values:
```
bbatsov: I think this big comment can be converted to a docstring.

src/clojure/nrepl/rich_printer.clj
```diff
@@ -0,0 +1,460 @@
...
+
+(defn ensure-defaults [bindings]
+  (let [bindings (merge-with #(or %1 %2) bindings defaults)]
+    (assoc bindings #'*print-budget*
+           (long (min (* 1N (bindings #'*print-level*) (bindings #'*print-length*)) Long/MAX_VALUE)))))
+
+(defprotocol MachinePrintable
+  (-print-on [x write rem-depth]))
+
+;; clojure 1.10 support
```
bbatsov: Or datafy/nav support? :-) Probably this will be a better description once 1.11 is out. :-)

bbatsov:
Apart from the build failure my general feedback is:

* it'd be nice if the new printer had some unit tests, as I don't think one integration test is enough to cover well that it's working properly
* All public functions should ideally have some docstring. 
* We need to add a bit of documentation to the manual about tool writers and how they can leverage the new printer. 

Apart from this - the code looks very solid!

src/clojure/nrepl/elisions.clj
bbatsov: Btw, why is this a separate ns to begin with? I wonder if it wouldn't better to have some `soft-store` ns and everything else to be in the `rich-print` ns.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -178,23 +178,21 @@
...
-            "clone" (register-session msg)
-            "close" (close-session msg)
-            "ls-sessions" (t/send transport (response-for msg :status :done
-                                                          :sessions (or (keys @sessions) [])))
-            (h msg)))))))
+    (if-some [the-session (if session
+                            (@sessions session)
+                            (create-session transport))]
+      (let [msg (-> msg (assoc :session the-session) (assoc-in [:print-options :nrepl/session] the-session))]
+        ;; TODO: yak, this is ugly; need to cleanly thread out-limit through to
```
bbatsov: Btw, do we still need this in light of the recent changes you've made?

src/clojure/nrepl/rich_printer.clj
```diff
@@ -0,0 +1,460 @@
...
+        (when (and *print-meta* (meta x))
+          (write "]"))
+        (when (datafiable? x)
+          (write " ")
+          (set! *print-budget* (bump *print-budget* 1))
+          (print-on write (tagged-literal 'unrepl/... (*elide* (lazy-seq [(list (browsify x))]))) (inc rem-depth))
+          (write "]"))))))
+
+(defn base64-encode [^java.io.InputStream in]
+  (let [table "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
```
bbatsov: Shouldn't this be some constant?

src/clojure/nrepl/rich_printer.clj
```diff
@@ -0,0 +1,460 @@
...
+        sb (StringBuilder.)]
+    (loop [shift 4 buf 0]
```
bbatsov: There a few magic numbers if this function - I wonder if we shouldn't give them some names. For me it's hard to figure out why those values are the ones needed.

src/clojure/nrepl/rich_printer.clj
```diff
@@ -0,0 +1,460 @@
...
+                                                                                      (-> .getGraphics (.drawImage img 0 0 nil)))
+                                                                                    "png" bos)
+                                                                               (java.io.ByteArrayInputStream. (.toByteArray bos)))))))})))
+
+   Object (fn [x]
+            (if (-> x class .isArray)
+              (seq x)
+              (str x)))})
+
+; datafy and friends shoud be but to good use there
```
bbatsov: Top-level comments are supposed to have 2+ `;`. At least according to common Lisp lore. :-)

src/clojure/nrepl/rich_printer.clj
```diff
@@ -0,0 +1,460 @@
+(ns nrepl.rich-printer
+  "A printer that produces a pure EDN representation of a Clojure value.
+  Features elisions, MIME attachments and general purpose object browsing.
+  Initially developed for unrepl."
```
arichiardi: Awesome!

cgrand:
@bbatsov:
> Apart from the build failure 

Do you have any idea why Eastwood does not exit? That's what makes the build to fail.

src/clojure/nrepl/elisions.clj
cgrand: `elisions` contains the most specific parts: the printer function (for pr-values) and the elision store strategy.

`nrepl.print.elisions` and `nrepl.print.elisions.rich-printer`?

bbatsov: Yeah, that sounds good to me.

bbatsov:
> Do you have any idea why Eastwood does not exit? That's what makes the build to fail.

Likely you've hit some bug in it. Does it hang locally as well? Perhaps @slipset might be able to help debugging this?

slipset:
I've reproduced this with the version of eastwood you're running (0.2.6), but cannot reproduce with current eastwood (0.3.4). 0.3.4 has a new linter which catches another error. I'll create a PR with upgraded eastwood and a fix for the error.

slipset:
PR #110 should fix this problem

cgrand:
@slipset The CI still hangs with 0.3.4 https://travis-ci.org/nrepl/nrepl/jobs/477316949

bbatsov:
I'll restart the build. If it hangs again I guess we'll have to disable it until @slipset can help us figuring out what's wrong.

cichli:
Can the eliding phase be separated from the printing phase? So that, for example, other middleware/ops could inspect and use the elided value? Or a custom pretty printer could be used?

bbatsov:
That's a good point!

@cgrand ping :-)

cgrand:
@bbatsov pong!
For elision to be decoupled from printing, printing should be decoupled from traversal.
Which brings us to what @bbatsov nicknamed `printify` recently: a function that given a value returns a representation of this value in a print-safe subset (more or less EDN). (Could `datafy` be used for that?)

There's a second consideration: representing elisions inline or outline. Unrepl represents them inline but it could be conceivable that a more text-based approach would prefer to first have the plain text and after the plain text a list of offsets+length+elisions telling which parts of the original output are elisions ("hypermedia affordances").

bbatsov:
The outline approach seems interesting, but I want to make sure I understand well the idea. @cgrand Can you suggest how a response would look if we opted for the outline approach?

cgrand:
@bbatsov first I most certainly meant offline and not outline :-)
Then here are the options as I see them:

1. use side effects (with dynamic binding around `*out*`) to record where elisions are in the text stream; this precludes building the output in a more functional manner as printing is mandatory. At the end you send the text and a list describing where elisions are. This is the "offline" approach.
2. use a more markupy approach (custom ANSI control sequences?). This is the inline approach.

I personally prefer the second option except that a syntax (well an escape sequence) for the markup is required.

bbatsov:
> @bbatsov first I most certainly meant offline and not outline :-)

🤣 

> I personally prefer the second option except that a syntax (well an escape sequence) for the markup is required.

I'm with you on this one. How complex (implementation-wise) do you estimate it would be go in this direction?

cgrand:
>  How complex (implementation-wise) do you estimate it would be go in this direction?

What are we talking about exactly?
 * transforming a value for print
 * designing a markup/escape mechanism (do we need an escape or are we ok with unlikely accidental collision -- like using a private unicode char or an unused ANSI OSC? it would definitely be simpler)

bbatsov:
> What are we talking about exactly?

I was referring to both points, as they differ from the approach we initially took (this PR). Some solution beats no solution (most of the time), but I'd definitely fond of separation of the ellisions and print steps idea. 

> designing a markup/escape mechanism (do we need an escape or are we ok with unlikely accidental collision -- like using a private unicode char or an unused ANSI OSC? it would definitely be simpler)

I'd go with the simpler solution.

bbatsov:
5 years later, I guess it's time to finally close this PR. :-)

PR #107, opened 2018-12-18
Stop reading and evaluating code on first read error

alexander-yakushev:
This has bothered me for a long long time.

Consider this example (default Clojure REPL):

```clj
user=> (comment {:a} (println "BOOM!"))
Syntax error reading source at (REPL:1:14).
Map literal must contain an even number of forms
BOOM!
nil
Syntax error reading source at (REPL:1:33).
Unmatched delimiter: )
```

Or this one:

```clj
user=> (defn foo [] (println {:a}) (println "BOOM!"))
Syntax error reading source at (REPL:8:27).
Map literal must contain an even number of forms
Syntax error reading source at (REPL:8:28).
Unmatched delimiter: )
BOOM!
nil
Syntax error reading source at (REPL:8:47).
Unmatched delimiter: )
```

Observe how because of a *syntactic error* (odd number of elements in a map) the println form gets executed, even though it shouldn't (since it's in `comment` and `defn` form respectively).

Besides this potentially dangerous but rare behavior, there is also an much more common inconvenience when you have to debug such errors. What should be the result of evaluating a single form triggers multiple exceptions, all but one of which don't make any sense. They obscure the original problem and make life harder, especially to the beginners. This is even more jarring in CIDER since it only displays the last exception in its stacktrace popup buffer, and that stacktrace is completely unrelated to the actual problem. The only thing to do is to remember that `Unmatched delimiter: )` means you probably have an odd-arg map somewhere. Bleh.

While in default Clojure REPL there isn't much to be done because it is stream based, in nREPL (which is message-based) it is sensible to skip over the whole message if a read error occurs. This is exactly what this patch does – on read error, it discards the remaining content of the StringReader which is passed to the `clojure.main/repl`. Only one (primary) error remains, and the rest of the broken code is not executed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 #### Changes
 
 * [#16](https://github.com/nrepl/nrepl/issues/16): Use a single session thread per evaluation.
+* [#107](https://github.com/nrepl/nrepl/issues/107): Stop reading and evaluating code on first read error.
 * Support single-arity printer functions.
 * Restore the `nrepl.bencode` namespace.
 
--- a/doc/modules/ROOT/pages/ops.adoc
+++ b/doc/modules/ROOT/pages/ops.adoc
@@ -59,7 +59,7 @@ Returns::
 
 === `:eval`
 
-Evaluates code.
+Evaluates code. Note that unlike regular stream-based Clojure REPLs, nREPL's ``:eval`` short-circuits on first read error and will not try to read and execute the remaining code in the message.
 
 Required parameters::
 * `:code` The code to be evaluated.
--- a/src/clojure/nrepl/middleware/interruptible_eval.clj
+++ b/src/clojure/nrepl/middleware/interruptible_eval.clj
@@ -71,7 +71,13 @@
                     (push-thread-bindings bindings))
            :read (if (string? code)
                    (let [reader (source-logging-pushback-reader code line column)]
-                     #(read {:read-cond :allow :eof %2} reader))
+                     #(try (read {:read-cond :allow :eof %2} reader)
+                           (catch RuntimeException e
+                             ;; If error happens during reading the string, we
+                             ;; don't want eval to start reading and executing the
+                             ;; rest of it. So we skip over the remaining text.
+                             (.skip ^LineNumberingPushbackReader reader Long/MAX_VALUE)
+                             (throw e))))
                    (let [code (.iterator ^Iterable code)]
                      #(or (and (.hasNext code) (.next code)) %2)))
            :prompt #(reset! session (maybe-restore-original-ns (capture-thread-bindings)))
@@ -138,7 +144,7 @@
                  {:requires #{"clone" "close" #'nrepl.middleware.pr-values/pr-values}
                   :expects #{}
                   :handles {"eval"
-                            {:doc "Evaluates code."
+                            {:doc "Evaluates code. Note that unlike regular stream-based Clojure REPLs, nREPL's `:eval` short-circuits on first read error and will not try to read and execute the remaining code in the message."
                              :requires {"code" "The code to be evaluated."
                                         "session" "The ID of the session within which to evaluate the code."}
                              :optional {"id" "An opaque message ID that will be included in responses related to the evaluation, and which may be used to restrict the scope of a later \"interrupt\" operation."
--- a/test/clojure/nrepl/sanity_test.clj
+++ b/test/clojure/nrepl/sanity_test.clj
@@ -70,6 +70,23 @@
     ["" 4 20 1 0] "(+ 2 2) (* *1 5) (/ *2 4) (- *3 4)"
     ["" nil] '*1))
 
+(deftest read-error-short-circuits-execution
+  (testing "read error prevents the remaining code from being read and executed"
+    (let [[err out & vals] (internal-eval "(comment {:a} (println \"BOOM!\"))")]
+      (if (and (= (:major *clojure-version*) 1)
+               (<= (:minor *clojure-version*) 9))
+        (is (re-matches #"(?s)^RuntimeException Map literal must contain an even number of forms[^\n]+\n$" err))
+        (is (re-matches #"(?s)^Syntax error reading source at[^\n]+\nMap literal must contain an even number of forms\n" err)))
+      (is (= "" out))
+      (is (= vals [nil]))))
+
+  (testing "exactly one read error is produced even if there is remaining code in the message"
+    (let [[err out & vals] (internal-eval ")]} 42")]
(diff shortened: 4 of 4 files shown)


bbatsov:
The suggested change makes sense to me. @cgrand @SevereOverfl0w Any thoughts from you?

test/clojure/nrepl/sanity_test.clj
```diff
@@ -70,6 +70,18 @@
     ["" 4 20 1 0] "(+ 2 2) (* *1 5) (/ *2 4) (- *3 4)"
     ["" nil] '*1))
 
+(deftest read-error-short-curcuits-execution
```
bbatsov: I think you using a couple of `context`s will make this test a bit better.

bbatsov:
> While in default Clojure REPL there isn't much to be done because it is stream based, in nREPL (which is message-based) it is sensible to skip over the whole message if a read error occurs. This is exactly what this patch does – on read error, it discards the remaining content of the StringReader which is passed to the clojure.main/repl. Only one (primary) error remains, and the rest of the broken code is not executed.

Btw, I also think this is worth documenting somewhere in the manual - e.g. in the section of the `eval` middleware.

test/clojure/nrepl/sanity_test.clj
alexander-yakushev: What's a context?

cgrand:
Why not just tag the following exceptions? This would avoid departing from the standard repl experience.

alexander-yakushev:
What do you mean by tagging exceptions?

bbatsov:
> Why not just tag the following exceptions? This would avoid departing from the standard repl experience.

Yeah, that'd be a nice middle ground. It would allow smarter clients to discard irrelevant exceptions, but this also opens the question whether those exception are ever useful. 

> What do you mean by tagging exceptions?

I guess @cgrand means adding some extra key to exception responses to indicate whether those are "root" exceptions (or vice versa). Then clients would be able to process them differently, if they are aware that the tags exist.

test/clojure/nrepl/sanity_test.clj
bbatsov: Sorry, too much Ruby - I meant https://clojuredocs.org/clojure.test/testing. It's `context` in RSpec. :-)

alexander-yakushev:
> Yeah, that'd be a nice middle ground. It would allow smarter clients to discard irrelevant exceptions, but this also opens the question whether those exception are ever useful.

However, this doesn't fix the issue that the remaining code is attempted to be read and executed which could lead to unforeseen consequences.

bbatsov:
Btw, it seems there's also some test failure on 1.10, because of the changed error message format.

alexander-yakushev:
Fixed 1.10 failing test and rebased.

bbatsov:
Thanks!

I'll merge this soon unless @cgrand or @SevereOverfl0w have any objections. I'm trying to think of a scenario when this can cause a problem, but nothing comes to my mind. If we really want to be on the safe side I guess we can add some extra flag to eval to control whether we're short-circuiting or not.

src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -76,7 +76,13 @@
                     (push-thread-bindings bindings))
            :read (if (string? code)
                    (let [reader (source-logging-pushback-reader code line column)]
-                     #(read {:read-cond :allow :eof %2} reader))
+                     #(try (read {:read-cond :allow :eof %2} reader)
+                           (catch RuntimeException e
+                             ;; If error happens during reading the string, we
+                             ;; don't want eval to start reading and executing the
+                             ;; rest of it. So we skip over the remaining text.
+                             (while (pos? (.skip ^LineNumberingPushbackReader reader 10000)))
```
cichli: Would calling `.close` on the reader work rather than skipping an arbitrary number of characters?

test/clojure/nrepl/sanity_test.clj
```diff
@@ -70,6 +70,23 @@
...
```
cichli: Typo: `curcuit` -> `circuit`

src/clojure/nrepl/middleware/interruptible_eval.clj
alexander-yakushev: It doesn't skip arbitrary number of characters - it skips everything left in the reader. If you refer to `10000`, that's just a "skipping step" since `.skip` doesn't have an API to "skip everything".

Calling `read` on a closed reader throws an exception, so it's not the cleanest way to do things. Skipping over the remaining content seems to me like the most natural thing to do, so that EOF logic remains unchanged.

cichli: You're right, `.close` is no good. Maybe `(.skip ^LineNumberingPushbackReader reader Long/MAX_VALUE)` without the loop and check?

alexander-yakushev: Yeah, I considered that too but went with a loop. I guess that might be better, let me change it.

doc/modules/ROOT/pages/ops.adoc
```diff
@@ -59,7 +59,7 @@ Returns::
 
 === `:eval`
 
-Evaluates code.
+Evaluates code. Note that unlike regular stream-based Clojure REPLs, nREPL's `:eval` short-circuits on first read error and will not try to read and execute the remaining code in the message.
```
bbatsov: Btw, that’s an auto-generated page, so you can’t edit it directly. You have to edit the op’s description and regen the page from it.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * [CLI] When starting the REPL, make sure the transport option is used correctly.
 * [CLI] Make sure calling `send-ack` at `cmdline` ns works with the correct transport.
 * [#8](https://github.com/nrepl/nrepl/issues/8): Clean up context classloader after eval.
+* [#107](https://github.com/nrepl/nrepl/issues/107) Stop reading and evaluating code on first read error.
```
bbatsov: Probably we should file this under “Changes”. There’s also a “:” missing here.

doc/modules/ROOT/pages/ops.adoc
alexander-yakushev: How to regen?

alexander-yakushev:
All done.

bbatsov:
Thanks for working on this! I really appreciate it! 🙇

alexander-yakushev:
Thanks for helping me merge this. Hope to see it someday in main build tools (maybe, in 5 years? 😆 )

Issue #97, opened 2018-12-12
Sideloader


cgrand:
Allow clients to act as network classloaders for the target repl. Thus dependencies can be added on the fly without requiring a shared FS.

PR #109, opened 2018-12-20
Add sideloader (see #97)

cgrand:
It was tricky to test (as tracking responses to several messages at the same time is required)
BTW all this sequences (client code) tied to effects and control make me uneasy.

### DIFF
--- a/src/clojure/nrepl/middleware/interruptible_eval.clj
+++ b/src/clojure/nrepl/middleware/interruptible_eval.clj
@@ -3,6 +3,7 @@
   (:require
    clojure.main
    clojure.test
+   [clojure.java.io :as io]
    [nrepl.middleware :refer [set-descriptor!]]
    [nrepl.middleware.caught :as caught]
    [nrepl.middleware.print :as print]
@@ -51,6 +52,23 @@
   (or (instance? ThreadDeath (clojure.main/root-cause e))
       (and (instance? Compiler$CompilerException e)
            (instance? ThreadDeath (.getCause e)))))
+(defn classloader
+  "Creates a classloader that obey standard delegating policy.
+   Takes two arguments: a parent classloader and a function which
+   takes a keyword (:resource or :class) and a string (a resource or a class name) and returns an array of bytes
+   or nil."
+  [parent f]
+  (proxy [clojure.lang.DynamicClassLoader] [parent]
+    (findResource [name]
+      (when-some  [bytes (f :resource name)]
+        (let [file (doto (java.io.File/createTempFile "unrepl-sideload-" (str "-" (re-find #"[^/]*$" name)))
+                     .deleteOnExit)]
+          (io/copy bytes file)
+          (-> file .toURI .toURL))))
+    (findClass [name]
+      (if-some  [bytes (f :class name)]
+        (.defineClass ^clojure.lang.DynamicClassLoader this name bytes nil)
+        (throw (ClassNotFoundException. name))))))
 
 (defn evaluate
   "Evaluates a msg's code within the dynamic context of its session.
@@ -91,9 +109,14 @@
                                    ;; specific hack reasonable?
                                    #'clojure.test/*test-out* out})
                             (cond-> explicit-ns (assoc #'*ns* explicit-ns)
-                                    file (assoc #'*file* file)))]
+                                    file (assoc #'*file* file)))
+                        cl (classloader (.getContextClassLoader (Thread/currentThread))
+                                        (fn [type name]
+                                          (when-some [f (:sideloader/resolve (meta session))]
+                                            (f type name))))]
                     (pop-thread-bindings)
-                    (push-thread-bindings bindings))
+                    (push-thread-bindings bindings)
+                    (.setContextClassLoader (Thread/currentThread) cl))
            :read (if (string? code)
                    (let [reader (source-logging-pushback-reader code line column)]
                      #(try (read {:read-cond :allow :eof %2} reader)
--- a/src/clojure/nrepl/middleware/sideloader.clj
+++ b/src/clojure/nrepl/middleware/sideloader.clj
@@ -7,7 +7,7 @@
   (:import nrepl.transport.Transport))
 
 ;; TODO: dedup with base64 in elisions branch once both are merged 
-(defn base64-encode [^java.io.InputStream in]
+(defn- base64-encode [^java.io.InputStream in]
   (let [table "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
         sb (StringBuilder.)]
     (loop [shift 4 buf 0]
@@ -31,7 +31,7 @@
                   (recur 4 0))
                 (recur shift (bit-shift-left buf 6))))))))))
 
-(defn base64-decode [^String s]
+(defn- base64-decode [^String s]
   (let [table "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
         in (java.io.StringReader. s)
         bos (java.io.ByteArrayOutputStream.)]
--- a/src/clojure/nrepl/middleware/sideloader.clj
+++ b/src/clojure/nrepl/middleware/sideloader.clj
@@ -1,0 +1,96 @@
+(ns nrepl.middleware.sideloader
+  {:author "Christophe Grand"}
+  (:require
+   [nrepl.middleware :as middleware :refer [set-descriptor!]]
+   [nrepl.transport :as t]
+   [nrepl.misc :refer [response-for]])
+  (:import nrepl.transport.Transport))
+
+;; TODO: dedup with base64 in elisions branch once both are merged 
+(defn base64-encode [^java.io.InputStream in]
+  (let [table "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+        sb (StringBuilder.)]
+    (loop [shift 4 buf 0]
+      (let [got (.read in)]
+        (if (neg? got)
+          (do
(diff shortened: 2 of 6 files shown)


bbatsov:
> BTW all this sequences (client code) tied to effects and control make me uneasy.

Can you elaborate on this?

cgrand:
Writing maximally lazy lazy sequences ops has always been an art and since the introduction of chunked sequences the laziness guarantees have been relaxed. So using them for anything queue-like is perilous (eg you don't get a message because something in the middle is not as lazy (it can be as stupid as a `&` in destructuring where a `rest` would have work)  so it waits for the next message before delivering the current one... but the next one won't occur until you reply or act on the current one).

It's so Clojure 1.0 (see `seque`, `pmap` etc.) :-)

bbatsov:
>  so it waits for the next message before delivering the current one... but the next one won't occur until you reply or act on the current one

So, the main concern is whether the provided things are going to be loaded in the right order without being serialized, right? 

 Btw, it's a bit hard for me to figure out the exact message exchange that has to happen between the client and the server here. Can you write down some pseudo-code message exchange? 

I assume the client does `sideloader-start` and then sends a bunch of `sideloader-provide`s, but I couldn't infer this from the tests - I just remember our conversation on the topic. :-)

cgrand:
No, the main concern is that in my experience using sequences as queues ia a fragile equilibrium. It can be made to work but is very brittle. In the present case it could be possible to have a deadlock purely because seq processing introduced an accidental lag.

I definitely have to document the side loader more.
Everything occurs in a single session.
You first send a `sideloader-start` message to, well, start the sideloading mechanism.
This message never gets a "done" response. It only gets a stream of `sideloader-lookup` responses (as required by evaluation).
In reaction to these `sideloader-lookup` the client MUST send a new message (different uuid) `sideloader-provide` possibly with empty `content` (when the resource is not known to the client). The server will reply with `done`.

bbatsov:
Makes sense. Thanks for the explanations!

Btw, how can we make this work with middleware - one problem that I see is that currently you have to start the server with right handler (a handler using all the middleware needed by the client) before connecting to the server. The sideloading would handle nice the case where you just need more libraries to eval code from, but I'm not certain how we can approach the middleware problem (and potentially transports as well).

cgrand:
@bbatsov for transports, do you have an actual scenario in mind?

For middlewares I can see two scenarios :
1. Setting up the stack.
1. Adding to the stack.

The first case should be covered by injection. My idea for injection is to first send a bootstrap payload which turns the current socket repl in something which resembles (from a wire perspective) to EDN (or bencode but  hacky)  transport + side loader. This payload puts the socket repl in a state where it's basically evaluating the equivalent of start-server so it's going to request all classes and clj files as needed.

```clj
(with-bootstrap-sideloader ; pseudo code as it could me more verbose to be deps free
  (require 'nrepl.server)
  (nrepl.server/upgrade (read)))
[additional middlewares] ; read by the read call above
```

arichiardi:
> You first send a sideloader-start message to, well, start the sideloading mechanism.
This message never gets a "done" response. It only gets a stream of sideloader-lookup responses (as required by evaluation).
In reaction to these sideloader-lookup the client MUST send a new message (different uuid) sideloader-provide possibly with empty content (when the resource is not known to the client). The server will reply with done.

Nice and concise explanation! 

I am sure this is probably not true, but I think I have a nice name for the message from server to client requesting a dependency: `sideloader-requirement xyz`! To which the client responds `sideloader-provide xyz` :smile:

Yeah a bit long :smile: Sorry for abruptly barging in but I am your best fan (...creepy...) :smile: :+1:

cgrand:
@arichiardi well I don't feel that `requirement` is a good match for the semantics (it's too strong imo). What about changing `sideloader-provide` to `sideloader-found`?

arichiardi:
@cgrand I found `lookup` not too descriptive but you are the master so I am not going to debate more :smile: I like `found` as well ;)

bbatsov:
@cgrand Sorry for dropping the ball on this, but now I'm back from my vacation and I really want us to get this ready and in the hands of the users. Can you rebase on top of `master` to see how things are with the new CI?

src/clojure/nrepl/middleware/sideloader.clj
```diff
@@ -0,0 +1,96 @@
+(ns nrepl.middleware.sideloader
+  {:author "Christophe Grand"}
+  (:require
+   [nrepl.middleware :as middleware :refer [set-descriptor!]]
+   [nrepl.transport :as t]
+   [nrepl.misc :refer [response-for]])
+  (:import nrepl.transport.Transport))
+
+;; TODO: dedup with base64 in elisions branch once both are merged 
+(defn base64-encode [^java.io.InputStream in]
```
bbatsov: I guess the base64 functions can be private for now, right?

bbatsov:
> @arichiardi well I don't feel that requirement is a good match for the semantics (it's too strong imo). What about changing sideloader-provide to sideloader-found?

Yeah, I guess most of the time we'd have to have some fallback in case the extra resources are not available to be loaded.

On the other hand name like `sideloader-require` and `sideloader-provide` sound good to me, but I don't have a strong preference here. I think if we've documented things well the names won't matter that much. Naming is hard. :-)

Overall the code looks good to me, so the only things that we need to decide are the exactly naming of the three ops. And we need to add a bit of documentation illustrating how people are supposed to leverage the feature. 

@nrepl/nrepl Any ideas/comments?

cgrand:
@bbatsov 
> Yeah, I guess most of the time we'd have to have some fallback in case the extra resources are not available to be loaded.
Would you mind elaborating?

bbatsov:
> Yeah, I guess most of the time we'd have to have some fallback in case the extra resources are not available to be loaded.

Well, I assumed you meant that requirement is too strong, because we should make the REPL somehow work even if the client fails to provide it with some capability. I might have misinterpreted your words. E.g. we might want to inject java.classpath, but if it's not around we might fallback to just checking the JVM property and so on. I guess you can actually clarify for me what you meant. :D

bbatsov:
@shen-tian In this failing job (https://circleci.com/gh/nrepl/nrepl/56), I can't see what exactly caused the job to fail. I also see this in the beginning `lein with-profile +1.7 test :javaunknown` which seems weird.

shen-tian:
Had a quick look at failing tests:

- 3 of them related to #132, and pre-date the PR. @cichli, thought?. This is an problem though, as it means CI will be as often red as green.
- Not sure what's going on with the cloverage one. Might be worth ignoring/skipping the job?
- The Java 11/Clojure 1.7 failure is odd. I'm going to investigate.

bbatsov:
> Not sure what's going on with the cloverage one. Might be worth ignoring/skipping the job?

I'd be fine with dropping this one.

shen-tian:
Small update: `:javaunknown` is because `sysutils` is hard coded to recognise Java versions, and does not support versions from 11 onwards. I'll send them a PR, but not NB I think.

(Do we select based on Java version?)

https://github.com/gonewest818/lein-sysutils/blob/cee145d7e38d8086edcea5b00cfd3b3d998587f9/src/leiningen/sysutils.clj#L17-L24

cgrand:
@bbatsov : 
> Well, I assumed you meant that requirement is too strong, because we should make the REPL somehow work even if the client fails to provide it with some capability. I might have misinterpreted your words. E.g. we might want to inject java.classpath, but if it's not around we might fallback to just checking the JVM property and so on. I guess you can actually clarify for me what you meant. :D

What I meant is that `require` sounds too imperious to me: a `require` have to succeed while a mere `lookup` can commonly return no match.

bbatsov:
> (Do we select based on Java version?)

@shen-tian Yeah, there were some tests that were invoked only for a specific Java version. 

> What I meant is that require sounds too imperious to me: a require have to succeed while a mere lookup can commonly return no match.

@cgrand Roger that! Can you add a few examples as docs about using the API, so tooling writers would have some reference? Afterwards we can ship this with 0.7 as an experimental feature.

src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -51,6 +52,23 @@
...
+(defn classloader
+  "Creates a classloader that obey standard delegating policy.
+   Takes two arguments: a parent classloader and a function which
+   takes a keyword (:resource or :class) and a string (a resource or a class name) and returns an array of bytes
+   or nil."
+  [parent f]
+  (proxy [clojure.lang.DynamicClassLoader] [parent]
+    (findResource [name]
+      (when-some  [bytes (f :resource name)]
+        (let [file (doto (java.io.File/createTempFile "unrepl-sideload-" (str "-" (re-find #"[^/]*$" name)))
```
bbatsov: unrepl -> nrepl :-)

bbatsov:
And a reminder about the extra documentation you mentioned here:

> I definitely have to document the side loader more.
> Everything occurs in a single session.
> You first send a sideloader-start message to, well, start the sideloading mechanism.
> This message never gets a "done" response. It only gets a stream of sideloader-lookup responses > > (as required by evaluation).
> In reaction to these sideloader-lookup the client MUST send a new message (different uuid)  sideloader-provide possibly with empty content (when the resource is not known to the client). The server will reply with done.

shen-tian:
I'd like to have a go at finishing this. I don't have commit rights to this branch. Should I open a new PR? I've done the rebasing on my fork.

Also, any pointers other than what's already discussed?

bbatsov:
> I'd like to have a go at finishing this. I don't have commit rights to this branch. Should I open a new PR? I've done the rebasing on my fork.

I think you can just open a new PR. 

> Also, any pointers other than what's already discussed?

Nope, it's all here and in #97. I guess you can also take a look at the same functionality in unrepl (which is where the idea came to have in this in nREPL as well)

bbatsov:
I'll close this one in favour of #162.

PR #115, opened 2019-01-04
Add more servers and more clients to their respective lists

technomancy:
Here's a bunch more implementations I was able to find.

I haven't tried them all, but they should probably be listed anyway.

Also I noticed on the clients page it tells you how to connect to an nREPL server on a remote host but doesn't mention anything about adding encryption or authentication; I think that section of the docs should probably be removed as it is misleading and potentially dangerous.

### DIFF
--- a/doc/modules/ROOT/pages/beyond_clojure.adoc
+++ b/doc/modules/ROOT/pages/beyond_clojure.adoc
@@ -1,13 +1,23 @@
 = Beyond Clojure
 
 As mentioned earlier the nREPL protocol is language-agnostic and can
-be leveraged for many languages that have similar semantics to Clojure
-(in particular the ability to evaluate code at runtime).
+be leveraged for many languages that have the ability to evaluate code
+at runtime.
 
-Here's a listing of alternative nREPL implementations.
+Note than certain clients may make Clojure-specific assumptions which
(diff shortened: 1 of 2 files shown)


bbatsov:
@arichiardi I told you a while ago you can think of nREPL as LSP alternative. I guess now you know what I mean. ;-) 

@technomancy Thanks!

> Also I noticed on the clients page it tells you how to connect to an nREPL server on a remote host but doesn't mention anything about adding encryption or authentication; I think that section of the docs should probably be removed as it is misleading and potentially dangerous.

I'll probably add something on the subject. We already have the following text on the server page:

> Keep in mind that running a nREPL server on a public address is an epic security hole! As the connections are insecure (no authentication, no authorization) by default anyone can connect to your app and modify its behaviour or run code on the remote host.

We've also been discussing with @rlbdv for a while here https://github.com/nrepl/nrepl/pull/46 how to add some authorization/authentication to the default protocol, so you're welcome to join that conversation. It's a tricky subject and we haven't made much headway. Still using ssh tunneling is the best way to work with remote hosts.

technomancy:
Oh, yeah if the security warnings are present elsewhere then this part is probably fine. Thanks.

PR #127, opened 2019-01-29
Move interrupt handling to session middleware

cichli:
Followup to #98 – keep the exec + interrupt logic together. Other middlewares that use `exec` for interruptible computations need not depend on `interruptible-eval`.

Also a minor tweak to the interrupt logic: we were returning `interrupt-id-mismatch` when attempting to interrupt an ephemeral session, which contradicts the docstring for that status. Instead add a new status for this case.


bbatsov:
Looks great! 💯 

I guess this would make it easy to simplify the implementation of things like cider-nrepl’s test middleware.

Issue #99, opened 2018-12-13
Preserve :actual-value of evaluation in middleware pipeline


RickMoynihan:
I've been implementing a middleware that can send evaluation results to the new cognitect REBL tool.  It's still early stages but it essentially works; with some rough edges to iron out.

Rich Hickey recently asked on slack clojurians on the `#rebl` channel how it works.  It seems when he'd looked into supporting REBL in nrepl he didn't think it was possible in the general case due to nrepl representing the result of the evaluation as a string.  

REBL makes use of the new clojure 1.10RC5 protocol functions in metadata extensions, and therefore having a string `:value` means these are lost if they're read back in; presumably resulting in a degraded experience for objects which can't have a simple print representation.  Instead Rich said he thought it would be beneficial to have an `:actual-value` also flow through the middleware response pipeline which would allow tools like nrebl.middleware and REBL an opportunity to do more than reading a string back would enable.

The conversation ended with me saying I'd approach the nrepl maintainers to chase this up and see if we could introduce an `:actual-value` in the response maps in interruptible-eval, in addition to the current `:value`.  The difference being that `:actual-value` would be the raw value until it gets to the wire; at which point it would be discarded... whilst `:value` would be serialised over the wire.

Before writing this though, I thought I'd check back at my implementation and the nrepl code to see how it worked; and I realised that I'd been mistaken and that nrepl wasn't giving my middleware a string at all.  I'd simply forgotten about the hour I spent trying to inject middleware between interruptible-eval and pr-values!

So I believe my nrebl middleware is working correctly; though I'm not 100% sure if I'm reliably inserting myself between the two middlewares or not (if you can think of a better way for me to do it I'd appreciate it - [see here](https://github.com/RickMoynihan/nrebl.middleware/blob/6f8f410f2e50b50ee1b554317fda1fdd825c7a75/src/nrebl/middleware.clj#L37-L40)).

So the good news is that it seems I can make my middleware work without introducing an `:actual-value` to nrepl.  However it also seems like it might also be a good idea to provide a value intended for inprocess middlewares to intercept the true return value and not just a printed one.  It seems this might be quite trivial by simply adding a new `:actual-value v` kvp [here](https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/middleware/interruptible_eval.clj#L110) in addition to the existing `:value` which I'd propose is left as is for backwards compatibility.

Does this sound sensible?  I've only begun studying nrepl's code, so could well be missing something.

[...]

PR #128, opened 2019-01-30
Add middleware for handling evaluation errors

cichli:
Since #101 we only update the session atom in the `:prompt` REPL phase. This means any middlewares using `(get @session #'*e)` during the `:caught` phase to access the thrown error will no longer work (`cider.nrepl.middleware.inspect` is the case that I have found). Arguably that was relying on an implementation detail of nREPL – I think a better way is to simply convey the thrown error as a value in the response, and handle it elsewhere.

For now we just call `clojure.main/repl-caught` on the error as before, and still return `ex` / `root-ex` from eval to maintain compatibility. In the future, we can look at making it configurable.

### DIFF
--- a/doc/modules/ROOT/pages/design/middleware.adoc
+++ b/doc/modules/ROOT/pages/design/middleware.adoc
@@ -275,3 +275,40 @@ a middleware descriptor's `:requires` set contains
 * Any requests it handles will contain the key `:nrepl.middleware.print/print-fn`,
   whose value is a function that calls the given printer function with the given
   options – i.e. its signature is `[value writer]`.
+
+== Evaluation Errors
+
+nREPL includes a `caught` middleware which provides a configurable hook for any
+`java.lang.Throwable` that should be conveyed interactively (generally by
+printing to `\*err*`). Like the `print` middleware, any options may be provided
+in either requests or responses (the former taking precedence over the latter if
+any options are specified in both). The following options are supported:
+
+* `:nrepl.middleware.caught/caught`: a fully-qualified symbol naming a var whose
+  function to use to convey interactive errors. Must point to a function that
+  takes a `java.lang.Throwable` as its sole argument. Defaults to
+  `clojure.main/repl-caught`.
+
+* `:nrepl.middleware.caught/print?`: if logical true, the printed value of any
+  interactive errors will be returned in the response (otherwise they will be
+  elided). Delegates to `nrepl.middleware.print` to perform the printing.
+  Defaults to false.
+
+[source,clojure]
+----
+{:op :eval
+ :code "(/ 1 0)"
+ :nrepl.middleware.caught/caught 'my.custom/print-stacktrace
+ :nrepl.middleware.caught/print? true}
+----
+
+The functionality of the `caught` middleware is reusable by other middlewares.
+If a middleware descriptor's `:requires` set contains
+`#'nrepl.middleware.caught/wrap-caught`, then it can expect:
+
+* Any returned responses containing the key `:nrepl.middleware.caught/throwable`
+  will have that key's corresponding value passed to the hook.
+
+* Any handled requests will contain the key
+  `:nrepl.middleware.caught/caught-fn`, whose value is a function that can be
+  called on a `java.lang.Throwable` to convey errors interactively.
--- a/src/clojure/nrepl/middleware/caught.clj
+++ b/src/clojure/nrepl/middleware/caught.clj
@@ -1,0 +1,101 @@
+(ns nrepl.middleware.caught
+  "Support for a hook for conveying errors interactively, akin to the `:caught`
+  option of `clojure.main/repl`. See the docstring of `wrap-caught` and the
+  Evaluation Errors section of the Middleware documentation for more
+  information."
+  {:author "Michael Griffiths"
+   :added  "0.6.0"}
+  (:require
+   [clojure.main]
+   [nrepl.middleware :refer [set-descriptor!]]
+   [nrepl.middleware.print :as print]
+   [nrepl.misc :as misc]
+   [nrepl.transport :as transport])
+  (:import
+   (nrepl.transport Transport)))
+
+(def ^:dynamic *caught-fn*
+  "Function to use to convey interactive errors (generally by printing to
+  `*err*`). Takes one argument, a `java.lang.Throwable`."
+  clojure.main/repl-caught)
+
+(def default-bindings
+  {#'*caught-fn* *caught-fn*})
+
+(defn- bound-configuration
+  []
+  {::caught-fn *caught-fn*})
+
+(def configuration-keys
+  [::caught-fn ::print?])
+
+(defn- resolve-caught
+  [{:keys [::caught transport] :as msg}]
+  (when-let [var-sym (some-> caught (symbol))]
+    (let [caught-var (misc/requiring-resolve var-sym)]
+      (when-not caught-var
+        (let [resp {:status ::error
+                    ::error (str "Couldn't resolve var " var-sym)}]
+          (transport/send transport (misc/response-for msg resp))))
+      caught-var)))
+
+(defn- caught-transport
+  [{:keys [transport] :as msg} opts]
+  (reify Transport
+    (recv [this]
+      (transport/recv transport))
+    (recv [this timeout]
+      (transport/recv transport timeout))
+    (send [this {:keys [::throwable] :as resp}]
+      (let [{:keys [::caught-fn ::print?]} (-> (merge msg (bound-configuration) resp opts)
+                                               (select-keys configuration-keys))]
+        (when throwable
+          (caught-fn throwable))
+        (transport/send transport (cond-> (apply dissoc resp configuration-keys)
+                                    (and throwable print?)
+                                    (update ::print/keys (fnil conj []) ::throwable)
+                                    (not print?)
+                                    (dissoc ::throwable))))
+      this)))
+
+(defn wrap-caught
+  "Middleware that provides a hook for any `java.lang.Throwable` that should be
+  conveyed interactively (generally by printing to `*err*`).
+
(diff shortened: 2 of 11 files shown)


bbatsov:
Hmm, I never thought about this. I agree with the general premise that we should probably expose the evaluation errors in the response messages, I wonder if this can’t as easily be a helper used by `load-file` and `eval`. 

If we go down the middleware route we’ll also need a better name for it and some documentation, otherwise people are going to have a very hard time figuring out where the exceptions are being attached to the message. I also find it weird we can have a middleware that’s considered implementation detail. :-)

@alexander-yakushev @cgrand What’s your take on this.

src/clojure/nrepl/middleware/caught.clj
```diff
@@ -0,0 +1,31 @@
+(ns nrepl.middleware.caught
```
bbatsov: I’m not vary fond of the term `caught`, as least not without some qualifier. Something with error/exception in its name would probably convey more sense.

bbatsov:
My bad - I didn’t read the code carefully enough. (It’s so hard to read diff without a computer :D)

So, I’ve got new feedback:

* `load-file` can also throw an exception
* probably we need a generic middleware that can remove any value from the response. This was already discussed in the context of #99. 

Now I understood why there was a todo about making the middleware configurable. :-) That’s definitely important. Anyways, now I really like this - the name misled me. Sorry about the noise!

src/clojure/nrepl/middleware/caught.clj
bbatsov: Or, as I just realize, we can make this a generic middleware for removing values from the responses and then we’d need a better name. :-)

alexander-yakushev:
I don't have an educated opinion on that. I used `(get @session #'*e)` in the inspector because it was the only way I found how to figure out an exception has happened during the eval phase.

cichli:
I have pushed a new version which makes the hook configurable (either in the request or by setting dynamic var, like the print middleware). Now you can do:

```clojure
user=> (set! nrepl.middleware.caught/*caught-fn* #(do (clojure.main/repl-caught %)
                                                      (clojure.repl/pst %)))
#function[user/eval5937/fn--5938]
user=> (first 1)
Execution error (IllegalArgumentException) at user/eval5941 (form-init5607983694156671525.clj:1).
Don't know how to create ISeq from: java.lang.Long
IllegalArgumentException Don't know how to create ISeq from: java.lang.Long
	clojure.lang.RT.seqFrom (RT.java:553)
	clojure.lang.RT.seq (RT.java:533)
	clojure.lang.RT.first (RT.java:689)
	clojure.core/first--5369 (core.clj:55)
	clojure.core/first--5369 (core.clj:55)
	user/eval5941 (form-init5607983694156671525.clj:1)
	user/eval5941 (form-init5607983694156671525.clj:1)
	clojure.lang.Compiler.eval (Compiler.java:7176)
	clojure.lang.Compiler.eval (Compiler.java:7131)
	clojure.core/eval (core.clj:3214)
	clojure.core/eval (core.clj:3210)
	clojure.main/repl/read-eval-print--9068/fn--9071 (main.clj:414)

user=>
```

or

```clojure
user=> (set! nrepl.middleware.caught/*caught-fn* io.aviso.repl/pretty-pst)
#function[io.aviso.repl/pretty-pst]
user=> (first 1)
 clojure.core/eval   core.clj: 3214
               ...
     user/eval5886  REPL Input
clojure.core/first   core.clj:   55
               ...
java.lang.IllegalArgumentException: Don't know how to create ISeq from: java.lang.Long

user=>
```

There is also a request-level option that clients can pass if they also want the throwable returned (printed by the printing middleware) –  this is for if they are capable of reading EDN client-side and want to use that data.

> * `load-file` can also throw an exception

It can, but since it delegates entirely to the `eval` middleware, this will still be handled correctly. The transport that `load-file` adds simply does `(.send transport (dissoc resp :ns))` so the ordering doesn't matter.

> probably we need a generic middleware that can remove any value from the response.

I'm not quite sure what you mean by this but I'll read through the ticket you linked again.

> I’m not vary fond of the term caught, as least not without some qualifier. Something with error/exception in its name would probably convey more sense.

Yeah, it's not great, but we're essentially just wrapping the `:caught` option of `clojure.main/repl`, with the same semantics and default (`clojure.main/repl-caught`). So I think any new name should be compelling (I haven't thought of one). I didn't want to include `handle` in the name as it's just a hook; it doesn't dictate any kind of control flow. Technically it handles both `Error`s and `Exception`s as they're subtypes of `Throwable`, so I'm reticent to use either of those in the name too.

bbatsov:
Well, in light of the current version of the middleware I think it’s fine for it to be standalone and to be merged as is. 

My original comments were sparkled by the fact that originally all the middleware did was to remove the exception from response (or it seemed so to me), that’s I suggested making this generic for any non-bencodable key that’d be useful on the middleware level, but doesn’t make much sense to be returned in the final response (like the linked discussion about having somewhere a non-printed value key, etc). 

Anyways, all of this is not really time-sensitive and is mostly orthogonal to the problem at hand. As far as I’m concerned we can merge this and cut nREPL 0.6 to unblock the changes needed in cider-nrepl and cider itself. Plus, small releases are always better.

doc/modules/ROOT/pages/design/middleware.adoc
```diff
@@ -275,3 +275,40 @@ a middleware descriptor's `:requires` set contains
...
+
+== Evaluation Errors
+
+nREPL includes a `caught` middleware which provides a configurable hook for any
+`java.lang.Throwable` that should be conveyed interactively (generally by
+printing to `\*err*`). Like the `print` middleware, any options may be provided
+in either requests or responses (the former taking precedence over the latter if
+any options are specified in both). The following options are supported:
+
+* `:nrepl.middleware.caught/caught`: a fully-qualified symbol naming a var whose
```
bbatsov: This should probably be `caught-fn` for consistency with `print-fn`.

src/clojure/nrepl/middleware/caught.clj
```diff
@@ -0,0 +1,97 @@
...
+  "Support for a hook for conveying errors interactively, akin to the `:caught`
+  option of `clojure.main/repl`. See the docstring of `wrap-caught` and the
+  Evaluation Errors section of the Middleware documentation for more
+  information."
+  {:author "Michael Griffiths"}
```
bbatsov: You can also add `:added “0.6”` here as a bit of extra meta info.

src/clojure/nrepl/misc.clj
```diff
@@ -48,3 +53,23 @@
                                                (-> session meta :id)
                                                session)}))]
     (merge basis response)))
+
+(defn requiring-resolve
```
bbatsov: I think internally we have at least 5 definitions of this here and there, so you can also replace those once this PR is merged. It’d be nice if we unified them.

src/clojure/nrepl/misc.clj
```diff
@@ -48,3 +53,23 @@
...
+  [sym]
+  (or (resolve sym)
+      (try
+        (require (symbol (namespace sym)))
+        (resolve sym)
+        (catch Exception _))))
+
+;; TODO: this should live in `nrepl.middleware.session`, but that would
+;; introduce a circular dependency with `nrepl.middleware.interruptible-eval`
+(defn interrupted?
```
bbatsov: Probably we can just duplicate it in the two namespaces to avoid having this weird placement. But clearly there’s a bit of more work to be done to cleanly decouple the session and the eval middlewares. My main concern here is that we’re making a public API change here, that we’ll undo pretty quickly most likely. I guess no one would actually use this outside of nREPL, but we have to be careful with introducing/removing public API elements regardless.

bbatsov:
I switched to ninja mode and managed to provide a bit of inline feedback as well! :-) Just a couple of tiny remarks here and there - overall everything seems solid and I stand by my assessment that we’re ready for the release.

cichli:
>My original comments were sparkled by the fact that originally all the middleware did was to remove the exception from response (or it seemed so to me), that’s I suggested making this generic for any non-bencodable key that’d be useful on the middleware level, but doesn’t make much sense to be returned in the final response (like the linked discussion about having somewhere a non-printed value key, etc).

Maybe we could add a "non-strict" mode to the bencode encoding step that drops any keys it can't encode properly and logs an informative warning. At the moment if you accidentally return something it can't encode, it throws an exception (meaning no response is returned at all) with a somewhat-unhelpful stacktrace.

>Also - forgive my tenseness and the lack of inline comments, but it’s hard to write proper feedback on a phone.

No problem – I appreciate you're doing what you can!

> I switched to ninja mode and managed to provide a bit of inline feedback as well!

I will respond inline.

src/clojure/nrepl/middleware/caught.clj
cichli: Done.

src/clojure/nrepl/misc.clj
cichli: Will do.

src/clojure/nrepl/misc.clj
cichli: Yeah, duplicating it is better – updated to add the public version to `nrepl.middleware.session` and a private version to `nrepl.middleware.interruptible-eval`.

doc/modules/ROOT/pages/design/middleware.adoc
cichli: I didn't want to use the same name for both options – we use `caught` for the name of the var (i.e. only at request level) and `caught-fn` for the actual function object. Using `caught` / `print` for the "var name" semantic in the request is consistent with the [`eval` option](https://github.com/nrepl/nrepl/blob/0.5.3/src/clojure/nrepl/middleware/interruptible_eval.clj#L256) in the eval middleware.

bbatsov:
Everything looks great now. No further remarks from me.

PR #133, opened 2019-03-22
Cleaning up CI a bit

shen-tian:
- Adding JUnit output, so CircleCI knows more about history of what tests tend to fail etc. Have a look [here](https://circleci.com/build-insights/gh/shen-tian/nrepl/junit-output) to see what it looks like.
- Default to Clojure 1.10 for `Makefile`.
- Skipping the Cloverage job. It's causing more noise than it's worth at the moment.


bbatsov:
Thanks!

PR #67, opened 2018-10-30
Added EDN transport

dotemacs:
With the introduction of EDN transport, it theoretically simplifies
the lives of all the clients since they don't have to use Bencode, but
can use the native EDN format.

When reading the EDN payload, it has to be cast to PushbackReader
because of:

https://dev.clojure.org/jira/browse/CLJ-1611
https://groups.google.com/forum/#!topic/clojure/3HSoA12v5nc

Adhered to:

- [x] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [x] You've added tests to cover your change(s)
- [x] All tests are passing
- [x] The new code is not generating reflection warnings
- [x] You've updated the [changelog](../CHANGELOG.md)(that only applies to user-visible changes)

Issue #132, opened 2019-03-11
Intermittent test failure relating to streamed-printing and bencode.


shen-tian:
## Expected behavior

CI tests should consistently pass or fail

## Actual behavior

CI tests are failing intermittently on different Java and Clojure versions, both on Travis and CircleCI

### Examples:

CircleCI [link](https://circleci.com/gh/nrepl/nrepl/9) Java 11, Clojure 1.10

```
lein test nrepl.core-test

lein test :only nrepl.core-test/streamed-printing

ERROR in (streamed-printing) (bencode.clj:177)
bencode transport
 respects buffer-size option
expected: (= [{:value "(0 1 2 3"} {:value " 4 5 6 7"} {:value " 8 9 10 "} {:value "11 12 13"} {:value " 14 15)"} {:ns "user"} {:status ["done"]}] (->> (message client {:op :eval, :code (code (range 16)), :nrepl.middleware.print/stream? 1, :nrepl.middleware.print/buffer-size 8}) (mapv (fn* [p1__3565#] (dissoc p1__3565# :id :session)))))
  actual: java.lang.ClassCastException: class clojure.lang.PersistentArrayMap cannot be cast to class [B (clojure.lang.PersistentArrayMap is in unnamed module of loader 'app'; [B is in module java.base of loader 'bootstrap')
 at nrepl.bencode$string_LT_payload.invokeStatic (bencode.clj:177)
    nrepl.bencode$string_LT_payload.invoke (bencode.clj:177)
    nrepl.bencode$read_map$fn__375.invoke (bencode.clj:291)
    clojure.core$map$fn__5847$fn__5848.invoke (core.clj:2742)
    clojure.core$partition_all$fn__8520$fn__8521.invoke (core.clj:7264)
    clojure.core.protocols$fn__8144.invokeStatic (protocols.clj:168)
    clojure.core.protocols/fn (protocols.clj:124)
```
Travis [link](https://travis-ci.org/nrepl/nrepl/jobs/504008959) Java 8, Clojure 1.10

```
lein test nrepl.core-test
lein test :only nrepl.core-test/streamed-printing
ERROR in (streamed-printing) (bencode.clj:177)
bencode transport
 respects buffer-size option
[...]

PR #135, opened 2019-03-24
Adding EDN transport, take 2

### LINKED ISSUE #60: Implement an EDN transport
Having an EDN transport would simplify the lives of people using Clojure/ClojureScript/Java clients for nREPL. It should be pretty easy to create a transport based on the default bencode transport that simply uses EDN instead of bencode. Here's the implementation of the `bencode` transport as a reference:

```clj
(defn bencode
  "Returns a Transport implementation that serializes messages
   over the given Socket or InputStream/OutputStream using bencode."
  ([^Socket s] (bencode s s s))
  ([in out & [^Socket s]]
   (let [in (PushbackInputStream. (io/input-stream in))
         out (io/output-stream out)]
     (fn-transport
      #(let [payload (rethrow-on-disconnection s (bencode/read-bencode in))
             unencoded (<bytes (payload "-unencoded"))
             to-decode (apply dissoc payload "-unencoded" unencoded)]
         (merge (dissoc payload "-unencoded")
                (when unencoded {"-unencoded" unencoded})
                (<bytes to-decode)))
      #(rethrow-on-disconnection s
                                 (locking out
                                   (doto out
                                     (bencode/write-bencode %)
                                     .flush)))
      (fn []
        (if s
          (.close s)
          (do
            (.close in)
            (.close out))))))))
```

The best thing about this transport is that it requires no extra deps, so it can be easily bundled with nREPL itself.

shen-tian:
My attempt to finish what @dotemacs started in #67. Closes #60 

I followed the discussion under the [previous PR](#67), which provided some great sign-posts. What I did, in rough order:

- Rebased @dotemacs's commits onto master,
- Pending the discussion that started in the last PR, and I'm sure will continue, around how EDN transport will work, I've taken a permissive approach of aggressively converting all keywords (keys or values) to strings within the transport for both send and receive. This may not be what we ultimately want to do, but allowed rest of the work to continue.
- Update: we can do without this now! ~~[Hacked around](https://github.com/nrepl/nrepl/commit/34ce2bc18f358be8b971dd78484bdabd7c4b0b6e) a "can't read UUID as a number" bug, which only happened for UUIDs started with a digit. The workaround is all our rand UUIDs start with a letter. This can't go into production, of course. Ideas? @pfeodrippe , you talked a bit about the bug [here](https://github.com/nrepl/nrepl/pull/67#issuecomment-445779570)~~
- Converted all the test to use a "normalized" form of `:status`, a set of keywords. The converting to sets in NB because the EDN transport preserves the set structure, whereas bencode did not. The keywordizing might not be as important, pending the previous question. This is mostly testing code though, so not a design decision.
- The biggest breakthrough was this [commit](https://github.com/nrepl/nrepl/commit/7057d0ceadf9a571d250acf4b814d8558ecf3135). I don't know why this works, but it fixed most of the printing related tests. `*print-readably*` is being left in an odd state by the print middleware, as far as I can gather. @cichli might have some insight?

There's still quite some work left to do, but I think the problem's back is broken. 

To finish this, from me:

- ~~Deciding how permissive/strict the EDN transport should be, and update accordingly.~~
- ~~Figuring out the underlying reason for the UUID reading bug, and remove the workaround.~~
- Figuring out the underlying reason for needing the `*print-readably*` binding.
- ~~Fixing the remaining test or two,~~ Update: fixed ~~and have a look at the assertion that was removed [here](https://github.com/nrepl/nrepl/commit/c17b8c43c66498562cf5d4f050769490b1c24538)~~
- ~~Some manual testing with a REPL implementation?~~ 

In the meanwhile, comments/ideas welcome.

### Checklist

- [x] All tests are passing

### DIFF
--- a/src/clojure/nrepl/core.clj
+++ b/src/clojure/nrepl/core.clj
@@ -219,6 +219,8 @@
                            (merge connect-defaults
                                   (socket-info uri))))))
 
+(add-socket-connect-method! "edn" {:transport-fn transport/nrepl+edn
+                                   :port 7888})
 (add-socket-connect-method! "nrepl" {:transport-fn transport/bencode
                                      :port 7888})
 (add-socket-connect-method! "telnet" {:transport-fn transport/tty})
--- a/src/clojure/nrepl/core.clj
+++ b/src/clojure/nrepl/core.clj
@@ -3,7 +3,7 @@
   {:author "Chas Emerick"}
   (:require
    clojure.set
-   [nrepl.misc :refer [uuid]]
+   [nrepl.misc :refer [uuid keyworded-set]]
    [nrepl.transport :as transport]
    [nrepl.version :as version])
   (:import
@@ -66,7 +66,9 @@
 (defn- delimited-transport-seq
   [client termination-statuses delimited-slots]
   (with-meta
-    (comp (partial take-until (comp #(seq (clojure.set/intersection % termination-statuses))
+    (comp (partial take-until (comp #(seq (clojure.set/intersection
+                                           (keyworded-set %)
+                                           (keyworded-set termination-statuses)))
                                     set
                                     :status))
           (let [keys (keys delimited-slots)]
--- a/src/clojure/nrepl/core.clj
+++ b/src/clojure/nrepl/core.clj
@@ -3,7 +3,7 @@
   {:author "Chas Emerick"}
   (:require
    clojure.set
-   [nrepl.misc :refer [uuid keyworded-set]]
+   [nrepl.misc :refer [uuid]]
    [nrepl.transport :as transport]
    [nrepl.version :as version])
   (:import
@@ -66,9 +66,7 @@
 (defn- delimited-transport-seq
   [client termination-statuses delimited-slots]
   (with-meta
-    (comp (partial take-until (comp #(seq (clojure.set/intersection
-                                           (keyworded-set %)
-                                           (keyworded-set termination-statuses)))
+    (comp (partial take-until (comp #(seq (clojure.set/intersection % termination-statuses))
                                     set
                                     :status))
           (let [keys (keys delimited-slots)]
--- a/src/clojure/nrepl/core.clj
+++ b/src/clojure/nrepl/core.clj
@@ -221,8 +221,8 @@
                            (merge connect-defaults
                                   (socket-info uri))))))
 
-(add-socket-connect-method! "edn" {:transport-fn transport/nrepl+edn
-                                   :port 7888})
+(add-socket-connect-method! "nrepl+edn" {:transport-fn transport/edn
+                                         :port 7888})
 (add-socket-connect-method! "nrepl" {:transport-fn transport/bencode
                                      :port 7888})
 (add-socket-connect-method! "telnet" {:transport-fn transport/tty})
--- a/src/clojure/nrepl/core.clj
+++ b/src/clojure/nrepl/core.clj
@@ -83,7 +83,7 @@
    messages related to the message :id that will terminate upon receipt of a
    \"done\" :status."
   [client {:keys [id] :as msg :or {id (uuid)}}]
-  (let [f (delimited-transport-seq client #{"done"} {:id id})]
+  (let [f (delimited-transport-seq client #{"done" :done} {:id id})]
     (f (assoc msg :id id))))
 
 (defn new-session
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -60,8 +60,3 @@
         (require (symbol (namespace sym)))
         (resolve sym)
         (catch Exception _))))
-
-(defn keyworded-set
-  "Turn the `coll` into a set of keywords."
-  [coll]
-  (set (map keyword coll)))
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -60,3 +60,8 @@
         (require (symbol (namespace sym)))
         (resolve sym)
         (catch Exception _))))
+
+(defn keyworded-set
+  "Turn the `coll` into a set of keywords."
+  [coll]
+  (set (map keyword coll)))
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -20,12 +20,7 @@
 (defn uuid
   "Returns a new UUID string."
   []
-  ;; HACK!!! don't merge with this!!
-  ;; Somehow, uuids starting with [0-9] are causing issues. Go figure
-  (let [newone (str (java.util.UUID/randomUUID))]
-    (if-not (re-matches #"^\d.*" newone)
-      newone
-      (uuid))))
+  (str (java.util.UUID/randomUUID)))
 
 (defn response-for
   "Returns a map containing the :session and :id from the \"request\" `msg`
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -20,7 +20,12 @@
 (defn uuid
   "Returns a new UUID string."
   []
-  (str (java.util.UUID/randomUUID)))
+  ;; HACK!!! don't merge with this!!
+  ;; Somehow, uuids starting with [0-9] are causing issues. Go figure
+  (let [newone (str (java.util.UUID/randomUUID))]
+    (if-not (re-matches #"^\d.*" newone)
+      newone
+      (uuid))))
 
 (defn response-for
   "Returns a map containing the :session and :id from the \"request\" `msg`
--- a/src/clojure/nrepl/transport.clj
+++ b/src/clojure/nrepl/transport.clj
@@ -92,9 +92,13 @@
   `(try
      ~@body
      (catch RuntimeException e#
-       (throw (SocketException. "The transport's socket appears to have lost its connection to the nREPL server")))
+       (if (= "EOF while reading" (.getMessage e#))
+         (throw (SocketException. "The transport's socket appears to have lost its connection to the nREPL server"))
+         (throw e#)))
      (catch EOFException e#
-       (throw (SocketException. "The transport's socket appears to have lost its connection to the nREPL server")))
+       (if (= "Invalid netstring. Unexpected end of input." (.getMessage e#))
+         (throw (SocketException. "The transport's socket appears to have lost its connection to the nREPL server"))
+         (throw e#)))
      (catch Throwable e#
        (if (and ~s (not (.isConnected ~s)))
          (throw (SocketException. "The transport's socket appears to have lost its connection to the nREPL server"))
--- a/src/clojure/nrepl/transport.clj
+++ b/src/clojure/nrepl/transport.clj
@@ -91,10 +91,6 @@
   [^Socket s & body]
   `(try
      ~@body
-     (catch RuntimeException e#
-       (throw (SocketException. "The transport's socket appears to have lost its connection to the nREPL server")))
-     (catch java.lang.NullPointerException e#
-       (throw (SocketException. "The transport's socket doesn't appears to have have received any data")))
      (catch EOFException e#
        (throw (SocketException. "The transport's socket appears to have lost its connection to the nREPL server")))
      (catch Throwable e#
--- a/src/clojure/nrepl/transport.clj
+++ b/src/clojure/nrepl/transport.clj
@@ -38,10 +38,9 @@
 
 (deftype FnTransport [recv-fn send-fn close]
   Transport
-  ;; TODO: this keywordization/stringification has no business being in FnTransport
-  (send [this msg] (-> msg stringify-keys send-fn) this)
+  (send [this msg] (send-fn msg) this)
   (recv [this] (.recv this Long/MAX_VALUE))
-  (recv [this timeout] (walk/keywordize-keys (recv-fn timeout)))
+  (recv [this timeout] (recv-fn timeout))
   java.io.Closeable
   (close [this] (close)))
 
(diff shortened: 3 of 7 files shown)


arichiardi:
Wow awesome job, thank you! Following from far🍿

shen-tian:
Managed to fix the last consistently breaking test! Gonna make a list of things that sporadically fail to squish:

- [ ] `edn-transport-communication - nrepl.edn-test` Something to do with not closing a socket/releasing the port properly? [example](https://circleci.com/gh/shen-tian/nrepl/475)
- [ ] `interrupt-load-file - nrepl.core-test` [example](https://circleci.com/gh/shen-tian/nrepl/473)

Fails but not in scope

- `streamed-printing - nrepl.core-test` - As reported in #132

bbatsov:
Great progress! I'll review the code a bit later.

Are things behaving well when you test them manually in a REPL?

src/clojure/nrepl/core.clj
```diff
@@ -219,6 +221,8 @@
                            (merge connect-defaults
                                   (socket-info uri))))))
 
+(add-socket-connect-method! "edn" {:transport-fn transport/nrepl+edn
```
bbatsov: I actually wanted us to name the protocol `nrepl+edn` and the function itself `edn`.

src/clojure/nrepl/core.clj
```diff
@@ -66,7 +66,9 @@
 (defn- delimited-transport-seq
   [client termination-statuses delimited-slots]
   (with-meta
-    (comp (partial take-until (comp #(seq (clojure.set/intersection % termination-statuses))
+    (comp (partial take-until (comp #(seq (clojure.set/intersection
+                                           (keyworded-set %)
```
bbatsov: Why is this change needed?

src/clojure/nrepl/misc.clj
```diff
@@ -60,3 +60,8 @@
         (require (symbol (namespace sym)))
         (resolve sym)
         (catch Exception _))))
+
+(defn keyworded-set
```
bbatsov: Such helpers should be kept private and in the namespaces that need them. The misc ns exposes functions useful to client writes (the name is a bit confusing).

src/clojure/nrepl/transport.clj
```diff
@@ -123,6 +130,40 @@
             (.close in)
             (.close out))))))))
 
+(defn stringify-everything
```
bbatsov: Helpers should be private.

bbatsov: Btw, this affects only keywords, right?

src/clojure/nrepl/transport.clj
```diff
@@ -123,6 +130,40 @@
...
+
+(defn nrepl+edn
+  "Returns a Transport implementation that serializes messages
+   over the given Socket or InputStream/OutputStream using EDN."
+  ([^Socket s] (nrepl+edn s s s))
+  ([in out & [^Socket s]]
+   (let [in (java.io.PushbackReader. (io/reader in))
+         out (io/writer out)]
+     (fn-transport
+      #(rethrow-on-disconnection s (stringify-everything (edn/read in)))
```
bbatsov: That was done mostly to appease the tests, right? In general it seems redundant to convert the keywords to strings.

shen-tian:
@bbatsov I haven't tried yet. I am a bit worried about all the dynamic print related vars. Not sure how much this is just a feature of nREPL v.s. something odd going on elsewhere in the system. Will try that later. In the meanwhile:

## On the EDN transport protocol:

What do we accept, and what do we guarantee in sending back? The default solution is that we just stick to the implicit (because it's never been exposed?) internal protocol for both, e.g.

```clj
{:op :eval :code "(+ 2 2)"}
```
### Pros:

A major plus for EDN transport is to serve Clojure(Script) clients, which will all be able to speak EDN natively. Also, we can always be strict now, and relax things later, rather than become stricter and causing breaking changes.

### Cons:

Cons: This will take quite some documenting/checking, as it's more prescriptive than the simpler bencode transport, which is stringify-everything. The would extend to all the middleware too, as they'd all be able to speak EDN now. This is kind of like having everyone seeing colour for the first time... and immediately requiring colour coordinated uniforms. I'm not even sure this is possible without a coordinated effort with all middlewares, as in essence, we are requiring more detail in the specification of all of them. Overall, this might not be possible in practice? @cgrand mentioned Postel's law in #67. Maybe this is more applicable to third party middlewares than clients?

### Alternatives:

#### Stringy-fy everything:

Requires not much more specification than what already exists. We'd need to specify `:status` is either a string or a set of strings etc., but I'm not aware of any other EDN structures that we are now exposing from the internals, that was previously hidden behind the simplicity of bencode. 

The PR as it stands follows this, though more to facilitate discussion, (and having passing tests as a starting point) than it being my preference.

#### Permissive on input, strict on output

Think we'll still have the same problem as before, as some stuff that's pretty much implementation detail now might become the defacto spec, possibly without notice to middlware authors.

#### Strict on everything we know about, be relaxed on the rest?

Between `cider-nrepl` `refactor-nrepl` `piggieback` and the build-in middlewares make up the bulk of real life ones, we can just define what they do, and leave "third-party" middleware to deal with the specification problem?

bbatsov:
The middlewares don't actually decide what type of data to return - that's entire up to the transport. In practice all middlewares kept to the simply data structures supported by bencode (namely lists and maps of primitive data types like strings and numbers) and we can't really change this, although we can have some middleware returning more complex data structures when they using the EDN transport. 

We will decide this down the road, but for starters I'd just replicate what we have right now. Bencode requests and responses couldn't contain keywords because bencode can't represent them, but with EDN that's obviously not a problem, so I'd just move the current stringification to the bencode transport implementation (as opposed to being global as it is now). Then every transport can individually decide what structure of requests and responses it requires. I doubt any client would aim to support different transports (as this is a bit pointless), so this differences won't really be felt most of the time. 

Does this make sense to you?

bbatsov:
This is where we messed up in the past https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/transport.clj#L40 It was marked to be fixed a long time ago, but with only one transport it wasn't that big of a deal anyways.

src/clojure/nrepl/core.clj
shen-tian: Ok. So swap those two around? Will do.

src/clojure/nrepl/core.clj
shen-tian: tbh, not sure. This is from one of @dotemacs's commits.

src/clojure/nrepl/misc.clj
shen-tian: I suspect they can be all moved into `core-test`, where it's mostly used.

src/clojure/nrepl/transport.clj
shen-tian: The last two comments will be impacted by the protocol story.

shen-tian:
> The middlewares don't actually decide what type of data to return - that's entirely up to the transport. In practice all middlewares kept to the simply data structures supported by bencode (namely lists and maps of primitive data types like strings and numbers) and we can't really change this, although we can have some middleware returning more complex data structures when they using the EDN transport.

Technically true, yes, but the method you favour, if I understand, is for the EDN transport to be transparent? This would expose (previously unexposed) workings of the middleware to the outside. This was hidden by the limitation of bencode previously. This is, in practice, making this canonical. 

For example, I think as it stands, `{:op :eval :code "(+ 2 2)"}` is not actually valid with a transparent transport. The version that would work is `{:op "eval" :code "(+ 2 2)"}`. This was previously discussed [here](https://github.com/nrepl/nrepl/pull/67#issuecomment-445783776) but not resolved.

This contradicts the published [documents](https://nrepl.org/nrepl/0.6.0/ops.html) but didn't matter, because the bencode stripping and keywordizing at the end hid that detail before.

src/clojure/nrepl/core.clj
bbatsov: Well, I think you should check if this actually breaks something as it looks weird to me.

src/clojure/nrepl/core.clj
bbatsov: 👍

bbatsov:
> For example, I think as it stands, {:op :eval :code "(+ 2 2)"} is not actually valid with a transparent transport. The version that would work is {:op "eval" :code "(+ 2 2)"}. This was previously discussed here but not resolved.

Yeah, this is a good point. Probably we should make the EDN transport permissive as well on the request level, but the conversion of keywords should be done only in the transport itself and we should rely on the top-level conversion the way the bencode transport does it. Btw, I do think that internally we should be converting everything to keywords, not the other way around.

shen-tian:
I hear you on changing all the keywordize/stringify into transports. Should we tackles it in this PR?

We can be permissive on the request in side... but would still need to decide what we emit though. I can go at sketching out a (cloujure.)spec in spirit of what you are describing. (btw, how do you feel about having a spec for response + requests? useful?)

Do you think there's room to release this feature as an "alpha/beta/experimental" feature, i.e. might change the format a bit? I'm a bit weary of a yak shaving mission where we have to be way more sure of the format than we need to be.

bbatsov:
> I hear you on changing all the keywordize/stringify into transports. Should we tackles it in this PR?

No, we can do this later.

> We can be permissive on the request in side... but would still need to decide what we emit though. I can go at sketching out a (cloujure.)spec in spirit of what you are describing. (btw, how do you feel about having a spec for response + requests? useful?)

That's a pretty cool idea! Specing the requests and responses would be super useful in the long run. 

> Do you think there's room to release this feature as an "alpha/beta/experimental" feature, i.e. might change the format a bit? I'm a bit weary of a yak shaving mission where we have to be way more sure of the format than we need to be.

Certainly. I don't want us to overdo things right away. I think that releasing something that works would be a great start and we can polish it later. Makes sense to stick to the current request/response format for the first release. Then a next step would to consider a simple mechanism to supply different responses based on the transport, but that's orthogonal to creating the transport.

src/clojure/nrepl/core.clj
shen-tian: Reverted. Everything seems to still work.

shen-tian:
Ok! Progress. Addressed all the comments I think, and we have some resolution on the question of the EDN protocol spec. It's the same as the internals, except I've put a shim in to make `:op` keywords, rather than strings. We can change the internals to actually use keywords... later. Once that's done, the transport can actually be transparent. I think this satisfies the need to get the transport out there. The shim ensures it doesn't contradict the docs, at least.

As long as we reserve the rights to adjust how the transport works later, I'm happy with that aspect for now.

In the course of this, I've moved the keywordizing/stringfication out of `FnTransport`. This might be a breaking change for anything that depended on it? But should be easy enough to fix at this stage. 

I'll do some manual testing of the transport next..

src/clojure/nrepl/misc.clj
shen-tian: Removed most usages of this. Then rolled into `clean-response` in `core-test`

test/clojure/nrepl/core_test.clj
```diff
@@ -70,7 +72,8 @@
 
 (defmacro def-repl-test
   [name & body]
-  `(deftest ~(with-meta name (merge {:private true} (meta name)))
+  ;; TODO: decide what to do with the private tags before merging
+  `(deftest ~(with-meta name (merge #_{:private true} (meta name)))
```
shen-tian: Why do we make the repl tests private? It stops one from using test selectors like this from working:

```
lein with-profile +1.10,+test test :only nrepl.core-test/transports-fail-on-disconnects
```

I'd like to remove the private bit if possible. This is also the one line that's generating eastwood warnings.

bbatsov: No idea. Likely some oversight. Feel free to remove this.

bbatsov:
Everything looks great to me at this point! I'll merge this once you're done with the manual testing.

shen-tian:
Just checking we are talking the same thing:

Did some manual testing per https://nrepl.org/nrepl/0.6.0/hacking_on_nrepl.html

Used variations of this code after running `lein repl`

```clj
(require '[nrepl.server :refer [start-server stop-server]])
(require '[nrepl.core :as nrepl])

(def server (start-server :port 7889 :transport-fn nrepl.transport/edn))
(with-open [conn (nrepl/connect :port 7889 :transport-fn nrepl.transport/edn)]
           (let [resp (-> (nrepl/client conn 1000)
                          (nrepl/message {:op :eval :code "(+ 2 2)"}))]
             (->> resp
                  (map nrepl/read-response-value)
                  nrepl/combine-responses)))
```

Seems to work as expected.

test/clojure/nrepl/core_test.clj
shen-tian: 👍

bbatsov:
Yeah, that's exactly what I meant. Seems we're in business then! 

I see there's some linting failure and some failing test `nrepl.core-test/test-url-connect` (I assume this happens randomly). Perhaps some race condition in the tests or something like this?

shen-tian:
I've made a list of tests that randomly fail above. I don't think they are EDN transport related, but rather what you see if you run through the test suite 250+ times :)

Would still like to fix those though... but later.

You are not worried about having to set the `*print-readably*` to make the conversion work? If not, I'm done on this one.

bbatsov:
I had forgotten about this. Seems it's true by default, so I'm not worried about it https://clojuredocs.org/clojure.core/*print-readably*

shen-tian:
> I had forgotten about this. Seems it's true by default, so I'm not worried about it https://clojuredocs.org/clojure.core/*print-readably*. Just add some todo in the code, and perhaps a ticket for @cichli to take a look into this problem when he has some time.

Left a note :)

PR #136, opened 2019-03-25
Fixing the stream-printing test

shen-tian:
Think I've isolated the sub-test that fails for #132. Using this PR to trigger some re-runs to see if a longer wait makes it more robust.


shen-tian:
Closing this, as just the simple solution doesn't work..

PR #137, opened 2019-03-31
Moving all stringification into bencode encoder itself.

shen-tian:
This extends the bencode encoder to handle maps with either keywords or symbols as keywords. Currently, this causes the `write-bencode` function to fail.

This allows us to further simplify the bencode transport, by relieving it of stringification duties.

---

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,13 @@
 
 ### New features
 
-* [#60](https://github.com/nrepl/nrepl/issues/60): Implemented EDN transport.
+* [#60](https://github.com/nrepl/nrepl/issues/60): Implemented EDN transport
+
+### Changes
+
+* [#137](https://github.com/nrepl/nrepl/pull/137): Expanded Bencode writer to work with
+  maps that has keywords or symbols as keys. This allowed a simplification of the
+  Bencode transport itself.
 
 ## 0.6.0 (2019-02-05)
 
--- a/src/clojure/nrepl/bencode.clj
+++ b/src/clojure/nrepl/bencode.clj
@@ -389,9 +389,23 @@
 
 (declare lexicographically)
 
+(defn #^{:private true} thing>string
+  [thing]
+  (cond
+    (string? thing)
+    thing
+    (or (keyword? thing)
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,11 @@
 
 ### New features
 
-* [#60](https://github.com/nrepl/nrepl/issues/60): Implemented EDN transport.
+* [#60](https://github.com/nrepl/nrepl/issues/60): Implemented EDN transport
+
+### Changes
+
+* [#137](https://github.com/nrepl/nrepl/pull/137): Expanded Encode transport to allow more key types
```
bbatsov: I think you meant to say Bencode here. :-)

bbatsov: I think you can also mention that the key types are symbol and keywords here.

Probably you'd be a good idea to add a couple of examples about the conversation of requests from external to internal representation to the bencode and edn sections of the manual.

src/clojure/nrepl/bencode.clj
```diff
@@ -389,9 +389,23 @@
 
 (declare lexicographically)
 
+(defn #^{:private true} thing>string
```
bbatsov: Isn't it more idiomatic to use naming like `thing->string`? And `defn-`?

bbatsov: Ah, I see you just kept the to style in this file. That makes sense. 

Btw, you'll also have to apply this change to the standalone `nrepl/bencode` project.

shen-tian: I'll update the standalone project once done.

CHANGELOG.md
shen-tian: Will correct. I've also been working on internal v.s. external representations. Would like to jot down some notes (as an issue, probably) to discuss before getting to documenting.

bbatsov: 👍

shen-tian: Fixed the changelog

PR #139, opened 2019-04-05
Stringification of all ops internally

### LINKED ISSUE #138: Are op names keywords or strings?
### Why does it matter

Stability and backwards compatibility is NB for nREPL. Right now, the string/keyword specification of messages is largely implied by the limitations of bencode transport. With the EDN transport (and transit one too?), we get the chance to be more exact about what messages look like. But also, we need to take some care to define a stable default behaviour. 

For example, all four of these messages look the same once ran through the bencode write/read and keywordizing cycle:

```clj
{:op :eval :code "(+ 2 2)"}
{:op "eval" :code "(+ 2 2)"}
{"op" "eval" :code "(+ 2 2)"}
{"op" "eval" "code" "(+ 2 2)"}
```

Do we want to accept all four going forward? Even if we all requests to be permissive, we should still give some guarantees that responses follow a consistent format. (I'm thinking of `describe` and `unkown-op`, which need to provide `:op` values).

As it sits in master, the EDN transport accepts only the first two, but responds with keywords.

### What do we think it should be

This is mixed, but lean towards keywords: this is what [nrepl.org](https://nrepl.org/nrepl/ops.html) refers to and what's used in @bbatsov's [blog posts](https://metaredux.com/posts/2019/03/29/nrepl-0-6.html). Internally, all the unit tests (so far as they can be used as examples) use keywords.

However, this is not [done consistently](https://metaredux.com/posts/2019/01/12/nrepl-beyond-clojure.html). Other texts discussing nrepl also refer to ops as strings [e.g.](https://lambdaisland.com/guides/clojure-repls/clojure-repls#org1355f8e). 

### What happens under the hood.

It's all strings. Both this project but also all the middlewares, first and third party.

shen-tian:
This follows on discussions from, and closes #138.

- Updates the handful of `:op` as keywords in the main body of the code.
- Updates most of the tests to use string ops. It's telling that most of what needed changing was the test suite. Also modified the test suite to be stricter on EDN transports (we'll need to update fastlane)
- The EDN transport is now entirely transparent! `edn/read` and `str` are sufficient.
- Added a `liberally-accept` wrapper to incoming messages, so we still accept keyword ops.
- Updated all documentation where I could find it to be consistent.
- Added the missing `uri-scheme`, left out of the original PR.
- Expanded the documentation section on EDN transport.

---

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/doc/modules/ROOT/pages/design/transports.adoc
+++ b/doc/modules/ROOT/pages/design/transports.adoc
@@ -24,7 +24,7 @@ link:https://github.com/nrepl/nrepl/wiki/Extensions[Other nREPL transports are p
 
 == Bencode Transport
 
-Bencode is a relatively simple format, and comprises of of two scalar types (byte strings and integer) and two collection types (dictionaries and lists).
+This is, and will mostly likely remain, nREPL's default and primary transport. It is pronounced B-Encode, and is a relatively simple format, comprising of only two scalar types (byte strings and integer) and two collection types (dictionaries and lists).
 
 A simple nREPL message, to run the operation `eval` on the code `(+ 2 2)` would look like this in bencode (linebreaks added for clarity):
 
@@ -44,11 +44,29 @@ In documentation, this message would typically be expressed in EDN format:
 {:op "eval" :code "(+ 2 2)"}
 ----
 
-which, once encoded, would yield the above bencode message. Since bencode does not have a keyword type, Clojure keywords are encoded and transmitted as strings. nREPL server performs a `keywordize-keys` operation on the decoded map. Thus, the use of keywords in writing nREPL message is mainly for clarity. The following messages, when encoded using nREPL's bencode encoder and received by nREPL, are eqivalent:
+Given that many, if not most, nREPL client do not use EDN as their native data formats, it's best to think of this as the format of a message once received by nREPL. Correpondingly, responses are in the form they take just before being sent through the transport.
+
+nREPL's bencode encoder/decoder performs the following type conversions:
+
+.How nrepl.bencode handles types
+|=======================================
+| EDN type > | Bencode type > | EDN type
+
+| String     | String         | String
+| Keyword    | String         | String
+| Symbol     | String         | String
+| Integer    | Integer        | Integer
+| Map        | Dictionary     | Map
+| Vector     | List           | Vector 
+| List       | List           | Vector
+| Set        | List           | Vector
+|=======================================
+
+In addition, nREPL server performs a `keywordize-keys` operation on the recieved map, restoring the likes of `:op` and `:code` to keywords. Thus, the use of keywords in writing nREPL message is mainly for clarity. The following messages, when encoded using nREPL's bencode encoder and received by nREPL, are eqivalent:
 
 [source,clojure]
 ----
- {:op "eval"  ...}
+ {:op "eval"  ...} ;; this is also the internal representation
  {:op :eval ...}
  {"op" "eval" ...}
  {"op" :eval ...}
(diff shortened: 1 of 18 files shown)


doc/modules/ROOT/pages/ops.adoc
```diff
@@ -4,11 +4,11 @@ This file is _generated_ by #'nrepl.impl.docs/-main
 ////
 = Supported nREPL operations
 
-[small]#generated from a verbose 'describe' response (nREPL v0.6.1-SNAPSHOT)#
+[small]#generated from a verbose 'describe' response (nREPL v0.7.0-SNAPSHOT)#
 
 == Operations
 
-=== `:clone`
+=== `"clone"`
```
bbatsov: I think here we shouldn't print them with quotes, as that looks a bit weird. `clone` would be fine.

doc/modules/ROOT/pages/design/transports.adoc
```diff
@@ -91,6 +91,8 @@ $ clj -R:nrepl -m nrepl.cmdline -t nrepl.transport/edn
...
-the data format changes.
+The main difference between the bencode transport and the EDN one is that instead of sending and receiving bencode dictionaries and lists, and being limited to integer and byte types, you'd be sending and receiving full EDN structures and types, including maps, vectors, lists, sets, strings, keywords and symbols. The structures of the messages is similar, only the data format changes.
+
+This may be useful in a couple of usecases: In some clients, including ClojureScript ones, where EDN is easier to handle than bencode. Furthermore, because this transport exposes more of the richer data types in the internals of nREPL and its middlewares, it may support more complex usecases.
+
+The introduction of the EDN transport does require more specification of the message format, whereby some vagueness was allowed for by the limitatioms of bencode. For example, `:op` values are strings, and `:status` is either a keyword, or a set of keywords. This may cause some surprised, unintended consequences. Thus tweaks to the EDN message format may be possible until the transport is declared mature. Documentation of this using `clojure.spec` is planned.
```
bbatsov: Probably we should add a couple of code examples here and in the bencode section, so it'd be easier for people to understand the differences.

doc/modules/ROOT/pages/ops.adoc
shen-tian: Done.

doc/modules/ROOT/pages/design/transports.adoc
shen-tian: I've added more details to the section now. Let me know if it's in the right direction? I feel the next step in documentation is to actually write a section on *messages*. We can probably more formally adopt the convention that in documentation, we express messages in EDN, and the messages we send/receive over the wire is either whatever `bencode/write-bencode` produces or the EDN as given.

bbatsov: > I've added more details to the section now. Let me know if it's in the right direction? I feel the next step in documentation is to actually write a section on messages. We can probably more formally adopt the convention that in documentation, we express messages in EDN, and the messages we send/receive over the wire is either whatever bencode/write-bencode produces or the EDN as given.

You're definitely moving in the right direction with the latest round of updates. I like your idea. 

> I've done most of a spec for messages, which is sitting in another branch. I suspect that would serve as message specs more effectively than written documentation, given the intended audience of tool authors.

 I generally see the two approaches as complimentary, as it's kind of weird to just point people to some specs without providing them any explanations. But I agree that for people working a lot with nREPL probably the specs are going to be the single source of truth.

src/clojure/nrepl/server.clj
```diff
@@ -19,11 +19,19 @@
     (catch Throwable t
       (log t "Unhandled REPL handler exception processing message" msg))))
 
+(defn- liberally-accept
+  "Accept messages that are not quite in spec. This comes into effect with
+   The EDN transport, and other transports that allow more types/data structures
+   than bencode, as there's more oppertunity to be out of specification."
```
bbatsov: opportunity

src/clojure/nrepl/server.clj
```diff
@@ -19,11 +19,19 @@
     (catch Throwable t
       (log t "Unhandled REPL handler exception processing message" msg))))
 
+(defn- liberally-accept
```
bbatsov: I think this name is a bit misleading, as the function doesn't really accept anything - it just normalizes the message map. Perhaps something like `normalize-msg` or whatever would be a better name?

bbatsov:
Apart from my small feedback I just added, I think this PR is pretty much ready for merging in my opinion.

doc/modules/ROOT/pages/design/transports.adoc
shen-tian: I've updated the Message section to refer to the transport section for details, and elaborated a bit more on how the bencode transport works.

src/clojure/nrepl/server.clj
shen-tian: Fixed.

src/clojure/nrepl/server.clj
shen-tian: Let's go with `normalize-msg`. I was alluding to Postel's law with "liberally accept", but normalize is probably more accurate.

shen-tian:
> Apart from my small feedback I just added, I think this PR is pretty much ready for merging in my opinion.

I added a bit more documentation since your review. Am happy with where this is now.

bbatsov:
Thanks! Great work, overall!

Btw, I wonder if we should also add something about dealing with Boolean and float/double in the docs. And the encoding of nil. Probably those are some of the most common points of confusions for newcomers.

shen-tian:
> Thanks! Great work, overall!
> 
> Btw, I wonder if we should also add something about dealing with Boolean and float/double in the docs. And the encoding of nil. Probably those are some of the most common points of confusions for newcomers.

Hmm... haven't even thought how those worked. `nil` encodes to empty array, right?

bbatsov:
I think so. But in practice we don't use booleans and nils anywhere due to their fuzzy semantics.

Issue #152, opened 2019-08-04
nREPL server doesn't clean up threads created by 'clone'


zcaudate:
I found that nrepl creates a thread after every 'clone' operation. However, the thread remains even when the 'close' operation is called. I don't know how to kill it.

## Steps to reproduce the problem

Open an nrepl server in one project (tried with versions 0.6.0 and 0.7.0-alpha1), the port I've set to 51311.

On another project, run this code to connect to the server couple of times to see the thread count increase:

```
(do (def -conn- (nrepl.core/connect :port 51311))
      (def -c1- (nrepl.core/client -conn- 10000))
      (doall (nrepl.core/message -c1- {:op "clone"}))
      (doall (nrepl.core/message -c1- {:op "clone"}))
      (.close -conn-))
```

This next code kills all the sessions, but doesn't kill the threads;

```
(do (def -conn- (nrepl.core/connect :port 51311))
      (def -c1- (nrepl.core/client -conn- 10000))
      (def sessions (first (nrepl.core/message -c1- {:op "ls-sessions"})))
      (doseq [session (:sessions sessions)]
        (nrepl.core/message -c1- {:op "close"
                                  :session session}))
      (.close -conn-))
```

### Clojure version

1.9.0

### Java version

11

### Operating system

MacOS

zcaudate:
the crucial thing is to call the  `(close)` function, defined here:
https://github.com/zcaudate-me/nrepl/blob/15c95cb7dba7643c0ba0e3cc5666bbe8478cf4a4/src/clojure/nrepl/middleware/session.clj#L193

the additional try/catch-loop forms: https://github.com/zcaudate-me/nrepl/commit/f2497f3ccad01c32b32a89afd1aff2f3e21665a9#diff-e28a7dd6eef2522ca6b471c473c41ea0R168-R169 is to suppress the exception:

```
Exception in thread "nRepl-session-4481681a-fd89-4a2f-9a84-0af8607ee38e" java.lang.InterruptedException
#object[nrepl.transport.FnTransport 0x21c12fe1 "nrepl.transport.FnTransport@21c12fe1"]
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.reportInterruptAfterWait(AbstractQueuedSynchronizer.java:2056)
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2090)
[...]

PR #153, opened 2019-08-04
fix for #152

zcaudate:

- TESTED You've added tests to cover your change(s)
- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#152](https://github.com/nrepl/nrepl/issues/60): Bug fix for 'close' operation to kill session thread
 * [#60](https://github.com/nrepl/nrepl/issues/60): Implemented EDN transport.
 
 ### Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bugs fixed
 
-* [#152](https://github.com/nrepl/nrepl/issues/152): Bug fix for 'close' operation to kill session thread
+* [#152](https://github.com/nrepl/nrepl/issues/152): Bug fix to kill session threads when closing sessions
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,11 @@
 
 ### New features
 
-* [#152](https://github.com/nrepl/nrepl/issues/60): Bug fix for 'close' operation to kill session thread
 * [#60](https://github.com/nrepl/nrepl/issues/60): Implemented EDN transport.
+
+### Bugs fixed
+
+* [#152](https://github.com/nrepl/nrepl/issues/152): Bug fix for 'close' operation to kill session thread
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#152](https://github.com/nrepl/nrepl/issues/60): Bug fix for 'close' operation to kill session thread
```
bbatsov: Seems you forgot the update that link to point to the actual issue. Bug fixes should go under "Bugs fixed", not under new features.

bbatsov:
The suggested fix seems reasonable to me, but I'd like to hear what the other members of @nrepl/nrepl think about it. 

It might be a good idea to add a few inline comments about the session threads lifecycle.

zcaudate:
I've added in a comment regarding `session-exec` and fixed the CHANGELOG

src/clojure/nrepl/middleware/session.clj
```diff
@@ -165,26 +165,32 @@
...
+    ;; This map is added to the meta of the session object by `register-session`,
+    ;; it contains functions that are accessed by `interrupt-session` and `close-session`.
+    {:interrupt (fn [exec-id]
+                  ;; nil means interrupt whatever is running
+                  ;; returns :idle, interrupted id or nil
                   (let [current @running]
                     (cond
                       (nil? current) :idle
-                      (and (or (nil? exec-id) (= current exec-id)) ; cas only checks identity, so check equality first
+                      (and (or (nil? exec-id) (= current exec-id)) ;; cas only checks identity, so check equality first
```
bbatsov: This should be just `;`, as it's an inline comment.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 * [#60](https://github.com/nrepl/nrepl/issues/60): Implemented EDN transport.
 
+### Bugs fixed
+
+* [#152](https://github.com/nrepl/nrepl/issues/152): Bug fix for 'close' operation to kill session thread
```
bbatsov: I guess a better wording would be "Kill session threads when closing sessions." (or something like this).

bbatsov:
The patch looks good to me. @cgrand @cichli @SevereOverfl0w Any thoughts from you?

SevereOverfl0w:
Just had a quick look at the code, and calling `(close)` seems like the right approach here based on my understanding that each session has a persisting thread that backs it.

zcaudate:
what's the progress for the PR?

bbatsov:
@zcaudate I had left a couple of tiny remarks and I was just waiting for them to be addressed. The PR is pretty much good to go.

bbatsov:
@zcaudate ping :-)

zcaudate:
@bbatsov. I didn’t realise you were waiting on my end. Will fix the changes asap.

bbatsov:
Thanks!

PR #160, opened 2019-09-18
Skip the interrupt on stream printing test until it's reliable.

shen-tian:
#132 is making CI fail almost all the time. Given there's no quick fix for it, (and we do think it's a real bug) can I suggest we skip the failing test so we can expect CI to be green?

As far as I can gather, the bug is not a big issue (given how long 0.6.0 has been in the wild).

If we are worried about just forgetting the issue, we can open a PR to add the test back in, and try to fix that? Feels like a better alternative than having the test fail by default?


bbatsov:
Yeah, I totally agree. A flaky test adds no value. Hopefully we'll figure out what's wrong with it eventually.

PR #161, opened 2019-09-21
Adding spec for messages

### LINKED ISSUE #140: Use clojure.spec to describe the request/response format
That's something we've discussed a while ago with @shen-tian. I'm just adding the ticket here, so we won't forget about this. 

As nREPL's requests and responses are just plain old maps they are perfect candidates to get some spec love.

shen-tian:
Closes #140. 

Here, we define the spec for a message, and for most of the end-to-end tests (`core-test`), check the responses against this. This is only enabled for the Clojure 1.10 test profile. We could enable it for 1.9 too, but not sure there's a point to that.

This feels like quite an un-intrusive way to start being more explicit about message formats for the sake of EDN transports (including fastlane).


bbatsov:
The PR looks good to me overall. @nrepl/nrepl any input/feedback?

src/spec/nrepl/spec.clj
```diff
@@ -0,0 +1,104 @@
...
+
+(s/def ::uuid-str #(and (string? %)
+                        (re-matches #"[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}"
+                                    (str/lower-case %))))
+
+;; should be uuid strings
+(s/def ::session (s/or :single ::uuid-str
+                       :multi  (s/coll-of ::uuid-str :kind? set?)))
+
+;; Clone
```
bbatsov: Might be a good idea to have a separate section for the request and response keys for every op. That'd be easier to read IMO.

shen-tian: Do you mean we should organise this source file better? If so, I've tried to roughly do that, but some keys are used in more than one op.

bbatsov: I was thinking we can start with a section "Common" for things like `id`, `session`, etc. 

`eval` and `load-file` can also be grouped together.

shen-tian: Can do.

bbatsov:
I guess this will do for now. Down the road we can do something similar for `cider-nrepl` as well.

bbatsov:
Thanks!

PR #162, opened 2019-09-24
Sideloader (take 2)

### LINKED ISSUE #97: Sideloader

shen-tian:
Closes #97 

This rebases @cgrand's PR #109, and fixes the test. The sideloader itself is from unrepl, so I'm guessing should work well. The bits I'd like to get done still are:

- [x] Add more tests, at least covering the cases of loading classes and resources.
- [x] Resolve the question around `clojure.lang.Compiler/LOADER` a bit. See note later.
- [x] Add documentation.

## How is this meant to be used?

As it stands, the sideloader doesn't work if we `(require 'foo.bar)` where `foo.bar` is missing on the nREPL classpath, because it's not hooked into the compiler classloader. See this [SO post](https://stackoverflow.com/questions/10941330/class-forname-in-clojure-not-respecting-contextclassloader). So to benefit from the classloader, the code being ran in the client would need to make explicit effort to trigger the sideloader (by explicitly using `Thread.currentThread.getContextClassLoader()` or similar). This probably works for a tooling "we are going to sideload these classes now" usecase, but not for a general "it just works" one. 

unrepl deals with this by wrapping the eval thread with a binding:

https://github.com/Unrepl/unrepl/blob/fa946eef88b0516dab81c8a9b3d8f9fcff06f44b/src/unrepl/repl.clj#L396-L397

I'm not sure it would be safe to do the same in `interruptible-eval`? This feels like one of those things that might have unforeseen consequences even if the client is not using the sideloader?

If the objective is to get this into nREPL so we can see what we can do with it, then we probably don't need to resolve this now. But some insights from people who's been thinking about this moer would be helpful.


bbatsov:
> If the objective is to get this into nREPL so we can see what we can do with it, then we probably don't need to resolve this now. But some insights from people who's been thinking about this moer would be helpful.

More or less. I've never envisioned us to auto-sideload stuff on `require` - the main usefulness I see for this is two-fold:

* auto-inject some libraries from nREPL clients - e.g. CIDER can auto-inject orchard if it's sees that `cider-nrepl` is not currently present and fallback to evaluation of code from orchard to power some of its command. Something like REPLy can sideload a completion library like `compliment` and so on.
* auto-injection of some middleware by clients - this is the dream, but it's going to be trickier as we'll also have to recalculate the middleware dependencies between each other after loading some extra middleware. 

At any rate - I was only thinking of using sideloading explicitly, without touching the existing evaluation logic. Not sure if @cgrand had a different idea about this.

bbatsov:
All that being said - I can certain envision many benefits if that's working transparently like in unrepl, so the client won't have to do extra checks before deciding if they'd like to sideload something. Then client can simply eval away and missing things would be loaded if possible. 

> So to benefit from the classloader, the code being ran in the client would need to make explicit effort to trigger the sideloader (by explicitly using Thread.currentThread.getContextClassLoader() or similar).

I can't quite imagine how this is going to look from the client's point of view. Might be easier to just have an explicit `sideload` message and dispense with the auto-lookup if we have to go down this route. 

Another small thing that bothers me with the current design is that the initial message `sideloader-start` doesn't return done. I think all message should return done. I'm also wondering if it makes sense to have a message to explictly disable sideloading for some session - e.g. after the client has happily loaded everything it needs, so we can avoid potential surprises for the end users related to sideloading.

shen-tian:
The (unrepl) API proposed here is very much geared at the "it just works" side loading. The cost is that we'd always need to keep on handle `sideloader-lookups` on the client side, as once we've started to sideload, we won't know when it will be done. Maybe `sideloader-start` never returns `done` is to reflect that?

Maybe a simpler API would be to do an explicit sideload:

```clj
;; ->
{:op      "sideload"
 :id      "1"
 :session "x"
 :code    (quote (require '[orchard.eldoc :as eldoc]
                          '[orchard.info :as info]))}
;; <-
{:id      "1"
 :session "x"
 :status  :sideloader-lookup
 :type    "resource"
 :name    "orchard/eldoc.clj"}
;; ->
{:id "1"
 :session "x"
 :type "resource"
 :name "orchard/eldoc.clj"
 :content "<base64 package>"}
;; <-
{:id      "1"
 :session "x"
 :status  :sideloader-lookup
 :type    "resource"
 :name    "orchard/info.clj"}
;; ->
{:id "1"
 :session "x"
 :type "resource"
 :name "orchard/info.clj"
 :content "<base64 package>"}
;; <-
{:id      "1"
 :session "x"
 :value   "nil"
 :status  :done}
```

Once this is done, we can go back to sending `eval`. Thoughts?

SevereOverfl0w:
The idea of unrepl's approach was that you could just send code and unrepl would call you if needed.

That way a client can drop in and out as necessary (eg fireplace did/does this).

shen-tian:
I think I found quite a safe way to deal with the `clojure.lang.Compiler/LOADER` issue: it only re-binds it if you've already called `sideloader-start`, so should have a low risk of unexpected breaking changes.

I'm quite happy with where the PR sits, other than documentation. Only other thing @bbatsov suggested (over slack) was a `sideloader-stop` op, so a client can "safe" nREPL after an initial setup.

bbatsov:
> I think I found quite a safe way to deal with the clojure.lang.Compiler/LOADER issue: it only re-binds it if you've already called sideloader-start, so should have a low risk of unexpected breaking changes.

Excellent! Let's proceed with this approach then. 

> I'm quite happy with where the PR sits, other than documentation.

Same here.

src/clojure/nrepl/middleware/sideloader.clj
```diff
@@ -0,0 +1,100 @@
+(ns nrepl.middleware.sideloader
```
bbatsov: It'd be nice to document the high-level idea here as well and add "0.7" as the version where the ns was introduced.

src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -52,6 +53,24 @@
...
+(defn classloader
+  "Creates a classloader that obey standard delegating policy.
+   Takes two arguments: a parent classloader and a function which
+   takes a keyword (:resource or :class) and a string (a resource or a class name) and returns an array of bytes
+   or nil."
+  [parent f]
+  (proxy [clojure.lang.DynamicClassLoader] [parent]
+    (findResource [name]
+      (when-some  [bytes (f "resource" name)]
+        (let [file (doto (java.io.File/createTempFile "unrepl-sideload-" (str "-" (re-find #"[^/]*$" name)))
```
bbatsov: unrepl -> nrepl :-)

bbatsov:
I've been thinking a bit about the docs for the sideloader and I think we need the following:

* simple overview under Design/Middleware
* some instructions for tool authors on how to use this. Probably we'll need to put this under a new section - e.g. "Building Clients" or something along those lines.

shen-tian:
> I've been thinking a bit about the docs for the sideloader and I think we need the following:
> 
> * simple overview under Design/Middleware
> * some instructions for tool authors on how to use this. Probably we'll need to put this under a new section - e.g. "Building Clients" or something along those lines.

Makes sense. I know you've been keen for a "guide for client authors" section going. Let's get that started, though I assume that's going to take a while to build up.

bbatsov:
Yep. I plan to tackle the rest of this somewhere in October. It has been on my TODO for quite a while already.

shen-tian:
I've added some documentation. What should we still do before this can be merged? I feel like the next step for this feature is for a client to try and implement it.

bbatsov:
Yeah, I agree with you. I guess that should be CIDER. :)

So, I'll just have to send from the client jars or clj files that are Base10 encoded, right? 

I also finally figured out why @cgrand decided not to send `done` - I see all the sideloader messages have the same id and I expected they'd have different ids. In the current arrangement the `done` response definitely doesn't make a lot of sense. Anyways just squash the related commits together and I'll merge this and start playing with it.

shen-tian:
I can remove the `:done` response before we merge? If we ever add a `sideloader-stop` op, that could be what triggers the `:done` response for the initial message.

bbatsov:
I like the idea! Let's do this.

liquidz:
@shen-tian Your work is awesome!
Just a quick question.
What response should I send in `sideloader-provide` when client didn't have a way to provide lookuped `type` and `name`?

`sideloader-stop` will also be usable for aborting `sideloader-lookup`?

liquidz:
> An empty response indicates "resource/class not found.

Oh, I missed this!
@shen-tian Sorry, please ignore the above question.

shen-tian:
Ok. The `done` response removed again, and commit history cleaned up. :)

shen-tian:
> @shen-tian Your work is awesome!
> Just a quick question.
> What response should I send in `sideloader-provide` when client didn't have a way to provide lookuped `type` and `name`?
> 
> `sideloader-stop` will also be usable for aborting `sideloader-lookup`?

To be clear, we are not implementing `sideloader-stop` for now, as it's not clear whether we'll need it. The circumstance I can imagine it being useful for is if we find sideloader mainly used to do a once-off setup on nREPL in a boot phase of a client, but the altered classloading behaviour causes other difficulties after that. Then there would be a case for reverting to the default classloader.

bbatsov:
Thanks! I'll cut a new alpha right away!

PR #163, opened 2019-10-04
First interrupt a thread. Stop it if that doesn't work

### LINKED ISSUE #158: Confused about interrupt
(Not sure if this is the right place to ask this question, please refer me to the right place if it's not).

I was running into some issues with interrupting evaluation from CIDER, and decided to dive into the code a bit for what's going on (great work BTW, was very readable!). There I found [this particular line](https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/middleware/session.clj#L190), which (I think) is used for actually interrupting the thread:

```clj
(doto ^Thread @thread .interrupt .stop)
```

Now, I think (haven't investigated further) my issue is caused by a race between `.stop` and `.interrupt`, but that doesn't really matter for my questions: 

* Why is `.stop` used here? AFAIK using this is [inherently unsafe](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#stop--), and I've actually ran into issues where I needed to restart the REPL.
* If `.stop`ing _is_ intended, I think the naming is confusing here. I think someone coming from a Java background would expect `.interrupt`, but not a `.stop` also. What do you think?

Thanks in advance!

shen-tian:
Fixes #158 

Right now, an interrupt first calls `Thread.interrupt()`. It waits 100ms for the thread to terminate itself. Failing that, it calls `Thread.stop()`. This gives code that respects the interrupt flag a chance to exit in a more safe manner.

A few things to note:

- `Thread.sleep()` throws an `InterruptedException` when interrupted. This causes an `eval-error` if not otherwise handled. One test has been updated to reflect this. This does change nREPL's behaviour. I'm not sure how much of an issue this could be?
- The 100ms waiting time was arbitrary. Guess the "correct" value depends on how often the thread being interrupted checks the flag. I have no idea what's the typical interval, but can imagine a situation where the client would want to set the time to be longer? I also feel like the default wait time shouldn't be too long. @FreekPaans: just to get one more data point, what would be the "right" timeout for you?

I think this provides us a way to deal with the #132, where a an interrupted eval with streamed printing can generate malformed bencode. Not too sure how that should be done though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,6 @@
 * [#140](https://github.com/nrepl/nrepl/issues/140): Added initial version of spec for message responses. These are used during Clojure 1.10 tests.
 * [#97](https://github.com/nrepl/nrepl/issues/97): Added a sideloader, a network classloader that allows dependencies to be added even the source/class files are not available on the server JVM's classpath.
 
-### Changes
-
-* [#158](https://github.com/nrepl/nrepl/issues/158): Interrupt now runs in three stages: calls `interrupt` on the thread, waits 100ms for the thread to respond and return messages, then waits 5000ms for the thread to terminate itself. A hard `.stop` is only called if it fails to do so.
-
 ### Bugs fixed
 
 * [#152](https://github.com/nrepl/nrepl/issues/152): Bug fix to kill session threads when closing sessions
@@ -21,6 +17,7 @@
 * [#137](https://github.com/nrepl/nrepl/pull/137): Expanded Bencode writer to work with
   maps that have keywords or symbols as keys. This allowed a simplification of the
   Bencode transport itself.
+* [#158](https://github.com/nrepl/nrepl/issues/158): Interrupt now runs in three stages: calls `interrupt` on the thread, waits 100ms for the thread to respond and return messages, then waits 5000ms for the thread to terminate itself. A hard `.stop` is only called if it fails to do so.
 
 ## 0.6.0 (2019-02-05)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 * [#140](https://github.com/nrepl/nrepl/issues/140): Added initial version of spec for message responses. These are used during Clojure 1.10 tests.
 * [#97](https://github.com/nrepl/nrepl/issues/97): Added a sideloader, a network classloader that allows dependencies to be added even the source/class files are not available on the server JVM's classpath.
 
+### Changes
+
+* [#158](https://github.com/nrepl/nrepl/issues/158): Interrupt now runs in three stages: calls `interrupt` on the thread, waits 100ms for the thread to respond and return messages, then waits 5000ms for the thread to terminate itself. A hard `.stop` is only called if it fails to do so.
+
 ### Bugs fixed
 
 * [#152](https://github.com/nrepl/nrepl/issues/152): Bug fix to kill session threads when closing sessions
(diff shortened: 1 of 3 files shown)


FreekPaans:
Hi. Nice work! A default of 100ms seems a little bit on the low end to me. From the perspective of a UI client (like CIDER), you would typically interrupt when something takes too long, right? So I guess an additional second or 2 of waiting wouldn't be a problem in that case. For my use case, 5 seconds would be fine.

Just another thought: how hard would be to encode the timeout in the actual message (sorry, not too familiar with nrepl)? In that case you can let the client decide.

shen-tian:
> Hi. Nice work! A default of 100ms seems a little bit on the low end to me. From the perspective of a UI client (like CIDER), you would typically interrupt when something takes too long, right? So I guess an additional second or 2 of waiting wouldn't be a problem in that case. For my use case, 5 seconds would be fine.
> 
> Just another thought: how hard would be to encode the timeout in the actual message (sorry, not too familiar with nrepl)? In that case you can let the client decide.

I also realised we don't actually need to wait for the thread to terminate or stop before returning on the interrupt call, so something like a 5s wait time wouldn't slow things down on the client side. 

The timeout is really "what's the cap on CPU time wasted on an eval after you interrupt it".

@bbatsov : your thoughts on adding this timeout as something the client can specify?

bbatsov:
> @bbatsov : your thoughts on adding this timeout as something the client can specify?

I'm on the fence about this. It really depends on whether we can agree that some default is good enough for most cases.

shen-tian:
I'm fiddling a version that does the wait-and-stop asynchronously.

If we can get that to work with say, a 5s timeout, that might provide a good enough default behaviour that we won't need to add a customisable parameter for now?

shen-tian:
Ok. Updated this to work as follows:

1. Calls interrupt
2. Wait 100ms. This is mainly to allow thread that respond quickly to interrupts to send a message back in response to the interrupt. Significantly, this includes an exception thrown by `Thread/sleep`.
3. Asynchronously: wait another 5000ms for the thread to cleanly terminate. Only calls `.stop` if it fails to do so.

I think this set of behaviours strikes a balance between allowing a thread to respond to an interrupt, but also ensuring we actually kill runaway processes.

The build is failing because of a network issue with the master build of clojure.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -192,8 +192,16 @@
                       (nil? current) :idle
                       (and (or (nil? exec-id) (= current exec-id)) ; cas only checks identity, so check equality first
                            (compare-and-set! running current nil))
-                      (do
-                        (doto ^Thread @thread .interrupt .stop)
+                      (let [t ^Thread @thread]
+                        ;; First interrupt the thread. Then wait for a timeout
+                        ;; if not terminated, do a hard stop
+                        (.interrupt t)
+                        (Thread/sleep 100)
```
bbatsov: Probably it's makes sense to duplicate the explanations from your last comment here as well.

shen-tian: Spun this out into its own fn, with docs added.

Also updated `close` to use the same logic. Think it makes sense.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 * [#60](https://github.com/nrepl/nrepl/issues/60): Implemented EDN transport.
 * [#140](https://github.com/nrepl/nrepl/issues/140): Added initial version of spec for message responses. These are used during Clojure 1.10 tests.
+
+### Changes
```
bbatsov: There's already a section "Changes" for this release.

shen-tian: oops. merged.

bbatsov:
Thanks!

PR #164, opened 2019-10-09
Updating the changelog, and making a correction in the docs

shen-tian:
Forgot to update the changelog with the last PR.

Also, we removed the `done` response to `sideloader-start` late in the PR, but didn't update the docs to reflect.


bbatsov:
Thanks!

Issue #167, opened 2019-10-14
Allow suppressing ack message when using nrepl.cmdline.


cursive-ide:
When using the `--ack` option to `nrepl.cmdline`, it currently prints a message:
```
Starting nREPL server...
/Users/colin/jdks/jbsdk8u152b1293.1_osx_x64/jdk/Contents/Home/bin/java etc etc
ack'ing my port 52838 to other server running on port 52833 nil
Connecting to local nREPL server...
nREPL server started on port 52838 on host localhost - nrepl://localhost:52838
Clojure 1.10.1
```
In the case of Cursive, and probably for other tooling, the two ports will almost certainly be arbitrary and the message is mostly useless to users. It would be nice to be able to hide this.

PR #168, opened 2019-10-17
Allow suppressing ack message when using nrepl.cmdline

cursive-ide:
This commit adds a `--verbose` command line option, and only displays the ack
message if this flag is passed. It also prints the message to `*out*` rather
than `*err*`, which didn't seem to make sense since it's not an erroneous
condition. Finally, the previous message also included the result of calling
`send-ack`, which always returns `nil`.

Fixes https://github.com/nrepl/nrepl/issues/167


bbatsov:
Nicely done! Thanks!

Issue #126, opened 2019-01-27
The built-in client hangs when using tty transfer


holyjak:
I am sure it is me who does something wrong, but I'd like to understand what and 

## Expected behavior

Starting the server with the tty transfer:

```clj
;; $  clj -R:nrepl
user=> (require '[nrepl.transport :as transport])
nil
user=> (require '[nrepl.server :refer [start-server stop-server]])
nil
user=> (start-server :port 55555 :transport-fn transport/tty)
```

and connecting with the client using the same transfer:

```sh
$ clj -R:nrepl -m nrepl.cmdline --connect --host 127.0.0.1 --port 55555 -t nrepl.transport/tty
```

I expect to see the prompt and be able to interact with it.

## Actual behavior

The client blocks, there is no output and it has to be ^C stopped. At that moment, the server prints `Exception in thread "nREPL-worker-0" java.lang.Error: java.net.SocketException: Socket closed ... at nrepl.transport$tty$write__427.invoke(transport.clj:136)`.

Beware: I see the same behavior when I omit the `-t nrepl.transport/tty` option so it seems it somehow is not processed correctly?

### Clojure version

1.10.0

### Java version

1.8

### Operating system

OSX

### nREPL

0.5.3

shen-tian:
Can confirm that this hangs:

`clj -Sdeps '{:deps {nrepl {:mvn/version "0.6.0"}}}' -m nrepl.cmdline --transport nrepl.transport/tty --interactive`

So this has been broken for a while now.

Looking at the changelog, the `--transport` flag was only fixed in `0.6.0`, so not sure there was ever a command line version of this that worked.

Gonna do some digging on when/how this last worked, but suspect this feature is pretty end of line, given the Socket REPL? Do we know anyone that uses this?

bbatsov:
I think your suggestion is quite reasonable. It's quite unlikely someone will want to use a `tty` server with the default client.

PR #169, opened 2019-10-19
Fixing the telnet server functionality

shen-tian:
While looking at #126, I noticed that the TTY server is broken. Turns out this happened in #135 (my bad), and was further impacted by #153.

To test this:

```
lein install
clj -Sdeps '{:deps {nrepl {:mvn/version "0.7.0-alpha2"}}}' -m nrepl.cmdline --transport nrepl.transport/tty --port 7888 --headless
```

Then, in a different terminal

```
telnet localhost 7888
```

I had to install telnet on my mac using `brew install telnet`.

Feels like we should have a regression test for this kinda thing... not too sure what that would look like though.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -29,7 +29,8 @@
                                     :sign-releases false}]]
 
   :profiles {:fastlane {:dependencies [[nrepl/fastlane "0.1.0"]]}
-             :test {:dependencies [[com.hypirion/io "0.3.1"]]
+             :test {:dependencies [[com.hypirion/io "0.3.1"]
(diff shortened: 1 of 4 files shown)


bbatsov:
> Feels like we should have a regression test for this kinda thing... not too sure what that would look like though.

Agreed. Probably the test should just spin a server, connect to it and send something for evaluation. Simple integration test.

shen-tian:
Added an extra testing dep, but seems to work as expected.

PR #170, opened 2019-10-20
Adding tests for openjdk 13

shen-tian:
Clojure adding official images for openjdk13 and 14 EA, so adding java 13 tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,9 +25,9 @@ executors:
     docker:
       - image: circleci/clojure:openjdk-11-lein-2.9.1
     <<: *defaults
-  openjdk14:
+  openjdk13:
     docker:
-      - image: circleci/clojure:openjdk-14-lein-2.9.1-buster
+      - image: circleci/clojure:openjdk-13-lein-2.9.1-buster
     <<: *defaults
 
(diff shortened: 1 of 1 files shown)


bbatsov:
I thought the latest version is 13. :-)

shen-tian:
Didn't mean to open the PR, but might as well now: Clojure (and thus CircleCI) does provide 13 and 14-ea OpenJDK images now.

Guess the question is, do we want to add another 5 builds into the CI matrix?

bbatsov:
I guess it makes sense to test on all supported versions which are currently 8, 11 and 13. Once 14 is out we'll replace 14 with it.

shen-tian:
Ok. Looks easy enough. I'll re-open the PR on that basis.

bbatsov:
Just remembered we even had some documentation about this - https://nrepl.org/nrepl/about/compatibility.html

shen-tian:
I was looking for that. Thanks. FWIW, the tests run fine on Java 14 EA.

shen-tian:
@bbatsov: pretty straightforward. We'll move from 13 -> 14 when the time comes.

bbatsov:
Thanks!

PR #173, opened 2019-11-10
Some housekeeping on middleware code

shen-tian:
- Removed some unused code from `interruptable-eval`, from back when it handled interrupts. Added doc string for the ns, including a note about this.
- Consolidated all the sideloader code into its own ns. The only interface to the rest of the codebase is just a `:classloader` key on the session.
- Renamed `sideloader` to `wrap-sideloader`, following preferred pattern for new middlewares.


bbatsov:
Nicely done! 🙇 

> Removed some unused code from interruptable-eval, from back when it handled interrupts. Added doc string for the ns, including a note about this.

Probably at some point we'll have to find a way to gracefully rename the middleware and its ns for consistency's sake. Duplicating the ns and just importing the public stuff there would be a simple way to do this I guess. 

> Renamed sideloader to wrap-sideloader, following preferred pattern for new middlewares.

This reminded me that we have to document this in the best practices or something.

PR #178, opened 2020-02-11
Add :read-cond option for `eval`.

FiV0:
Allow to pass reader-conditional options, so that reader
conditionals can be processed in higher middlerware layers.

- [ ] All tests are passing (No a bencode one fails)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   Bencode transport itself.
 * [#158](https://github.com/nrepl/nrepl/issues/158): Interrupt now runs in three stages: calls `interrupt` on the thread, waits 100ms for the thread to respond and return messages, then waits 5000ms for the thread to terminate itself. A hard `.stop` is only called if it fails to do so.
 * [#167](https://github.com/nrepl/nrepl/issues/167): Allow suppressing ack message when using nrepl.cmdline
+* [#178](https://github.com/nrepl/nrepl/pull/178): Allow `:read-cond` option when evaluating code.
 
 ## 0.6.0 (2019-02-05)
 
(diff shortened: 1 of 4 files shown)


FiV0:
So this is half of the fix for https://github.com/clojure-emacs/cider-nrepl/issues/437.

One test fails for the Bencode Transport layer with "Syntax error reading source at (REPL:1:3).\nConditional read not allowed\n". I don't know why that is the case. Maybe you can give me a pointer.

bbatsov:
Where is this failure exactly? I opened some of the failed builds and I see the only test failing is the one added in the PR? I'm guessing the error you got originated somewhere in `clojure.repl`, so it's probably best to check the code there which generated it.

shen-tian:
I'm guessing the bencode transport stringizes keywords in value slots, including `:preserve`. If I try to run this line in a REPL, I get the "Conditional read not allowed" error:

```
(read-string {:read-cond "preserve"} "#?(:clj (Math/random))")
```

So try this:

1. Either decide to send it as a string, and keywordize in `interruptable-eval` OR

2. modify this fn: to noramalize anything in `:read-cond` slot to keyword?

https://github.com/nrepl/nrepl/blob/151102df3088f7da6702bd162fc56af99dd7c5e6/src/clojure/nrepl/server.clj#L23-L29

I'd suggest the first option, as we typically don't use keywords as values in other requests, and it keeps the logic concentrated.

shen-tian:
I'm guessing the bencode transport stringizes keywords in value slots, including `:preserve`. If I try to run this line in a REPL, I get the "Conditional read not allowed" error:

```
(read-string {:read-cond "preserve"} "#?(:clj (Math/random))")
```

So try this:

1. Either decide to send it as a string, and keywordize in `interruptable-eval` OR

2. modify this fn: to noramalize anything in `:read-cond` slot to keyword?

https://github.com/nrepl/nrepl/blob/151102df3088f7da6702bd162fc56af99dd7c5e6/src/clojure/nrepl/server.clj#L23-L29

I'd suggest the first option, as we typically don't use keywords as values in other requests, and it keeps the logic concentrated.

bbatsov:
@FiV0 ping :-)

FiV0:
Seen the reply. Will look at this on the weekend.

bbatsov:
@FiV0 Just in time. :-) Seems we're one small code style fix from merging this. I'll probably cut the new nREPL tomorrow.

Issue #113, opened 2018-12-30
Hot-loading dependencies sample doesn't work in lein


jc776:
## Expected behavior
The pomegranate [hot-loading dependencies](https://nrepl.org/nrepl/usage/misc.html) sample (for `clj`) should also work in `lein`.

## Actual behavior
The dependency gets added without error to some classloader, but isn't available to `require`. It's visible as a resource.
```
PS C:\Users\Jack\SSD\Projects\classloaders-201> lein run -m nrepl.cmdline --interactive
nREPL server started on port 53374 on host 127.0.0.1 - nrepl://127.0.0.1:53374
nREPL 0.5.3
Clojure 1.10.0
Java HotSpot(TM) 64-Bit Server VM 10.0.1+10
Interrupt: Control+C
Exit:      Control+D or (exit) or (quit)
user=> (require 'bidi.bidi)
Execution error (FileNotFoundException) at user/eval1282 (form-init11955807739262716208.clj:1).
Could not locate bidi/bidi__init.class, bidi/bidi.clj or bidi/bidi.cljc on classpath.
user=> (require '[cemerick.pomegranate :as pg])
nil
user=> (pg/add-dependencies :coordinates '[[bidi "2.1.5"]]
                     :repositories (merge cemerick.pomegranate.aether/maven-central {"clojars" "https://clojars.org/repo"}))
{[bidi "2.1.5"] #{[prismatic/schema "1.1.7" :exclusions [[org.clojure/tools.reader]]]}, [prismatic/schema "1.1.7" :exclusions [[org.clojure/tools.reader]]] nil}
user=> (require 'bidi.bidi)
Execution error (FileNotFoundException) at user/eval1756 (form-init11955807739262716208.clj:1).
Could not locate bidi/bidi__init.class, bidi/bidi.clj or bidi/bidi.cljc on classpath.
user=> (.getResource (.getContextClassLoader (Thread/currentThread)) "bidi/bidi.cljc")
#object[java.net.URL 0x93a1078 "jar:file:/C:/Users/Jack/.m2/repository/bidi/bidi/2.1.5/bidi-2.1.5.jar!/bidi/bidi.cljc"]
user=>
```

## Steps to reproduce the problem
Example project: https://github.com/jc776/classloaders-201, then follow REPL session above
It is possible to get the dependency to load by bringing back some additional behaviour that was in cemerick/pomegranate#102 but never merged, which was closed in favour of nrepl/nrepl#35 - see project's README.

[...]

PR #185, opened 2020-05-01
Dynamic loading of middleware

### LINKED ISSUE #143: Consider adding some mechanism for deferred/dynamic middleware loading
## Deferred loading

Some middlewares might end up requiring a lot of code which results in a significant slowdown when booting nREPL. cider-nrepl really struggled with this in its early versions and eventually we started deferring requires in middlewares, before @vspinu came up with a way to defer loading middlewares until their first usage. 

The implementation for this lives [here](https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl.clj#L42). The usage looks like:

``` clojure
(def-wrapper wrap-xref cider.nrepl.middleware.xref/handle-xref
  {:doc "Middleware that provides find references functionality."
   :handles {"fn-refs"
             {:doc "Look up functions that reference a particular function."
              :requires {"symbol" "The symbol to lookup"
                         "ns" "The current namespace"}
              :returns {"fn-refs" "A list of function references."
                        "status" "done"}}
             "fn-deps"
             {:doc "Look up the function dependencies of particular function."
              :requires {"symbol" "The symbol to lookup"
                         "ns" "The current namespace"}
              :returns {"fn-deps" "A list of function deps."
                        "status" "done"}}}})
```

I'm reasonably certain we can come up with a better API for this if we simply make it part of middleware descriptor (for instance). There are two things to note:

* You won't be able to put the descriptors in the same file as the middleware definition, as you want to have access to it before the middleware has been loaded
* There are some unsolved issues with the current approach when it comes parallel loading of multiple namespaces (https://github.com/clojure-emacs/cider-nrepl/issues/447)

## Alternative approach (dynamic loading)

In https://github.com/clojure-emacs/cider-nrepl/pull/438 Chas posted his perspective in all of this which is quite interesting:

```
Just to define some terms as I'll use them (and as they've been used by a couple of others that've worked on things similar to this):

* _deferred_ loading is where portions of a middleware stack are activated either via explicit demand, or, as here, based on some ambient criteria (e.g. receiving the first message of a particular type)
* _dynamic_ loading is where the middleware stack can be altered at any time, including adding, removing, and reordering it to satisfy user/client demands. 

This PR looks to accomplish the former for a cider stack.

To answer @bbatsov's question, dynamic loading will definitely be part of nREPL. Phil has worked on this more than I have and has a couple of concrete proposals, so the particular approach isn't set in stone, but the result would also satisfy the objectives of any deferred-loading change. (Since a dynamically-loaded middleware wouldn't be `require`d at all until a client/user requested it.)

To be clear, I am _not_ saying "don't merge this". If loading the code that supports cider's middleware stack is determined to be a serious UX problem for the community here, then godspeed.```

I think this approach ties nicely with @cgrand's idea about injecting dependencies from clients. 

## Summary

Generally speaking it seems wiser to work in the direction of dynamic loading, but on the other hand we already have a working implementation for deferred loading. We can either try to integrate this better with nREPL or simply ship it as some helper which people may chose to use or not. 

I don't think that @technomancy has been working on his ideas for the dynamic loading recently, but it'd be great if he can share them with us. Perhaps someone would be willing to pick up his work. 

I'd love to hear what the members of @nrepl/nrepl think about this, and all nREPL client maintainers, of course.

### PR DESCRIPTION
**Pull request by shen-tian**
This add a new middleware, `dynamic-loader` (open to other suggestions), that allows us to modify the middleware stack on the fly. When combined with sideloading, this would allow us to inject middleware that the host JVM is not even aware of at startup.

A possible use case, e.g. CIDER, to connect to an already running nREPL instance, check what middleware are loaded, and if missing `cider-nrepl`, load it.

Closes #143 

## API

The new middleware defines three new ops:

- `ls-middleware`: returns a list of middleware currently loaded.
- `add-middleware`: add a collection of middleware to the existing stack. This will re-sort all the existing middleware as well as the new ones.
- `swap-middleware`: replaces the existing stack with a new set of middleware.

## Implementation

This is done via a new middleware, `dnyamic-loader/wrap-dynamic-loader`. There are two interesting bits

### Passing of state

An earlier design had the `dynamic-loader` being a specialised handler, and not a middleware. It was initialised with a list of middleware. This worked quite well, and the handler held its own state of what middleware are active. However it had to be the top most layer, so we can't use any upstream middleware in this handler. In particular, not having access to `session` made sideloading difficult.

Thus, the current design used an external state atom that stores `:handler`, the active handler fn built out of all the middleware and `:stack`, the list of individual active middleware. In `server/default-handler`, we create this atom, and close over it to return the handler.

We the bind a `*state*` var in the `dynamic-loader` ns to this atom when calling the handler. This allows the middleware to modify the global(ish) handler/stack state. (The state is not actually global. It will be created each time we create a new server via `default-handler`).
This allows us to preserve `wrap-dynamic-handler` as a 1-arity fn that works well with `set-descriptor!` and `linearize-middleware-stack` etc., while effectively passing the state in.

Here's the key lines of code, showing how we create state, bootstrap the stack, and close over the state in the resultant handler:

```clojure
(defn default-handler
  [& additional-middlewares]
  (let [initial-handler (dynamic-loader/wrap-dynamic-loader unknown-op)
        state           (atom {:handler initial-handler
                               :stack   [#'nrepl.middleware.dynamic-loader/wrap-dynamic-loader]})]
    (binding [dynamic-loader/*state* state]
      (initial-handler {:op          "swap-middlewares"
                        :state       state
                        :middlewares (concat default-middlewares additional-middlewares)}))
    (fn [msg]
      (binding [dynamic-loader/*state* state]
        ((:handler @state) msg)))))
```

Open to suggestions on this approach. It's a bit loopy, but the weirdness is concentrated in very few places, and makes it easy to reason able this middleware like any other, most of the time.

### Working with the sideloader

When updating middleware, this will look to the same `(:classloader (meta session))` value that the sideloader uses, just like `interruptable-eval`. You'll want to enable the sideloader in the "nREPL config" session, update the middleware, and be free to use the middleware without having to deal the sideloader during regular use.

Note that the `dynamic-loader` doesn't interact directly with `sideloader`, thus can be before or after it on the stack. However, it does require `session` to come before it, just for the sideloading case.

## Other notes

- The PR is not ready to merge, but does have working tests, including the all important case where we combine this with the sideloader. I'd be keen for comments on the design.
- I'd like to do more on error handling/documentation/tests
- Once we have this in, a path for deferred middleware loading, e.g. the current cider ones, can be simplified a bit: 
  - Instead of loading all middleware, and have them manage their own deferred loading, have an "entry point" middleware, which doesn't depend on any of the others, but has a registry of which opts correspond to which middleware. 
  - When one of those ops is first called, use the dynamic loader to add the relevant middleware to the stack, then handle the original message.

### DIFF
--- a/doc/modules/ROOT/pages/building_clients.adoc
+++ b/doc/modules/ROOT/pages/building_clients.adoc
@@ -116,6 +116,58 @@ This unblocks the `eval` command, which returns the value `Qaz` in the final mes
 
 It's thus important, once sideloading has started, to asynchronously handle lookup requests, without these being blocked by waiting on a response from another message.
 
+== Modifying middleware
+
+To add a middleware that's already available on the server's classpath, it's as
+simple as sending the message
+
+[source,clojure]
+----
+{:op "add-middleware"
+ :middleware ["cider.nrepl.middleware/wrap-version"]}
+----
+
+However, if the middleware is loaded, it's very likely that it hasn't been included
+as a dependency on the server, and thus unavailable on its classpath. Furthermore.
+In this case, we can use the `dynamic-loader` in conjunction with the `sideloader`:
+
+[source,clojure]
+----
+{:op "sideloader-start"}
+;; handle sideloading separatedly...
+;; now we add the middleware
+{:op "add-middleware"
+ :middleware ["cider.nrepl.middleware/wrap-version"]}
+;; confirm it's being loaded..
+{:op "ls-middleware"}
+;; and we should get something like...
+{:status #{:done}
+ :middleware [... "#'cider.nrepl.middleware/wrap-version" ...]}
+----
+
+However, if we tried to use the middleware with an `cider-version` op, we'd get an
+error, because the middleware is implemented in a different namespace, which is
+only loaded on the first use of the `cider-version` op. This is a practice in
+many middleware to improve startup performance. One method of getting around this
+is to request the extra namespace to be loaded at `add-middleware` time too:
+
+[source,clojure]
+----
+;; after starting the sideloader...
+{:op "add-middleware"
+ :middleware ["cider.nrepl.middleware/wrap-version"]
+ :extra-namespaces ["cider.nrepl.middleware.version"]}
+ ;; now, the following shoud work
+ {:op "cider-version"}
+----
+
+There is no operation to remove a single middleware, but it's possible to reset
+the stack to a baseline with the `swap-middleware` operation. If the goal is to
+simply reset the middleware stack, use this in conjuction with
+`nrepl.server/default-middleware`.
+
+Also note that updating the middleware stack may also destroy/re-create middleware state. As an example, sideloading would need to be re-started. The impact on each middleware differs, however, as some of them, e.g. `session` holds their state globally.
+
 == Additional Resources
 
 * https://mauricio.szabo.link/blog/2020/04/04/implementing-a-nrepl-client/
--- a/doc/modules/ROOT/pages/design/middleware.adoc
+++ b/doc/modules/ROOT/pages/design/middleware.adoc
@@ -28,13 +28,13 @@ All of nREPL's provided default functionality is implemented in terms of
 middleware, even foundational bits like session and eval support.  This default
 middleware "stack" aims to match and exceed the functionality offered by the
 standard Clojure REPL, and is available at
-`nrepl.server/default-middlewares`.  Concretely, it consists of a
+`nrepl.server/default-middleware`.  Concretely, it consists of a
 number of middleware functions' vars that are implicitly merged with any
 user-specified middleware provided to
 `nrepl.server/default-handler`.  To understand how that implicit
 merge works, we'll first need to talk about middleware "descriptors".
 
-link:https://github.com/nrepl/nrepl/wiki/Extensions[Other nREPL middlewares are provided by the community].
+link:https://github.com/nrepl/nrepl/wiki/Extensions[Other nREPL middleware are provided by the community].
 
 (See xref:ops.adoc[this documentation listing] for
 details as to the operations implemented by nREPL's default middleware stack,
@@ -125,7 +125,7 @@ Clojure's default stream-based REPL implementation).
 
 The specific contents of a middleware's descriptor depends entirely on its
 objectives: which operations it is to implement/define, how it is to modify
-incoming request messages, and which higher- and lower-level middlewares are to
+incoming request messages, and which higher- and lower-level middleware are to
 aid in accomplishing its aims.
 
 nREPL uses the dependency information in descriptors in order to produce a
@@ -268,7 +268,7 @@ specified in both). The following options are supported:
  :nrepl.middleware.print/quota 8096}
 ----
 
-The functionality of the `print` middleware is reusable by other middlewares. If
+The functionality of the `print` middleware is reusable by other middleware. If
 a middleware descriptor's `:requires` set contains
 `#'nrepl.middleware.print/wrap-print`, then it can expect:
 
@@ -316,7 +316,7 @@ any options are specified in both). The following options are supported:
  :nrepl.middleware.caught/print? true}
 ----
 
-The functionality of the `caught` middleware is reusable by other middlewares.
+The functionality of the `caught` middleware is reusable by other middleware.
 If a middleware descriptor's `:requires` set contains
 `#'nrepl.middleware.caught/wrap-caught`, then it can expect:
 
@@ -359,3 +359,19 @@ See link:../building_clients.html[Building Clients] on how to implement the serv
 Once the sideloader has been started, it can be triggered by an ordinary `eval` or `load-file` operation. The nREPL server will first try to find a resource/class on the classpath of its own JVM. Failing this, it will attempt to request it from the nREPL client by the above described mechanism. Should this still fail, expect a `ClassNotFoundException` as usual.
 
 TIP: Once a class has been loaded, it will become available to all sessions on nREPL server.
+
+== Dynamic middleware loading
+
+NOTE: Dynamic middleware loading support was added in nREPL 0.8 and the API is still
+considered experimental, and may change.
+
+nREPL includes a `dynamic-loader` middleware, which can be used, at runtime,
+query and change the middleware stack the nREPL server is using. This is especially
+powerful when combined with sideloading, as it allows a client to configure the
+server after connecting, and provides an alternative to having to specify the middleware required by the client as startup time.
+
+This introduces three new operations:
+
+* `ls-middleware`, to return a list of active middleware, ordered from inside outwards.
+* `add-middleware`, which adds a middleware to the stack. Optionally, a list of `extra-namespaces` could be provided for loading. This is useful when adding middleware that implement some form of deferred loading. Examples include `cider-nrepl` and `refactor-nrepl`. In these cases, some of the required namespaces might only be loaded upon first use, which may occur outside of a sideloading session, and thus fail. This feature allows us to pre-load namespaces when we add a middleware. If loading of any particular middleware fails, the stack will be unchanged.
+* `swap-middleware`, similar to `add-`, but replaces all existing middleware. Note that this _may_ remove the `dynamic-loader` itself.
--- a/src/clojure/nrepl/middleware/dynamic_loader.clj
+++ b/src/clojure/nrepl/middleware/dynamic_loader.clj
@@ -3,9 +3,14 @@
   nREPL server. This can be used by tools to configure an existing instance of
   an environment after connection.
 
+  It can also be used to load extra namespaces, in addition to the ones that new
+  middleware are defined in, to handle existing middleware that performs
+  deferred loading.
+
   When combined with the sideloader, this could be used to inject middleware
   that are unknown to the server prior to connection."
-  {:author "Shen Tian", :added "0.8"}
+  {:author "Shen Tian"
+   :added  "0.8"}
   (:require [clojure.string :as str]
             [nrepl.middleware :refer [linearize-middleware-stack set-descriptor!]]
             [nrepl.middleware.session :as middleware.session]
@@ -34,8 +39,7 @@
       (if (every? some? resolved)
         (reset! *state* {:handler ((apply comp (reverse stack)) unknown-op)
                          :stack   stack})
-        {:error      "Unable to resolve all middleware"
-         :unresolved (keep (fn [[m resolved]]
+        {:unresolved (keep (fn [[m resolved]]
                              (when (nil? resolved) m))
                            (zipmap middleware resolved))}))))
 
@@ -63,36 +67,40 @@
   [h]
   (fn [{:keys [op transport session middleware extra-namespaces] :as msg}]
     (when-not (instance? clojure.lang.IAtom *state*)
-      (throw (ex-info "dynamic-loader/*state* is not bond to an atom. This is likely a mistake" nil)))
+      (throw (ex-info "dynamic-loader/*state* is not bond to an atom. This is likely a bug" nil)))
     (case op
       "add-middleware"
       (do
         (require-namespaces session extra-namespaces)
-        (let [{:keys [error unresolved]}
+        (let [{:keys [unresolved]}
               (update-stack! session (concat middleware (:stack @*state*)))]
-          (if-not error
+          (if-not unresolved
             (t/send transport (response-for msg {:status :done}))
             (t/send transport (response-for msg {:status                #{:done :error}
-                                                 :error                 error
                                                  :unresolved-middleware unresolved})))))
 
       "swap-middleware"
       (do
         (require-namespaces session extra-namespaces)
-        (let [{:keys [error unresolved]} (update-stack! session middleware)]
+        (let [{:keys [unresolved]} (update-stack! session middleware)]
           (when transport
-            (if-not error
+            (if-not unresolved
               (t/send transport (response-for msg {:status :done}))
               (t/send transport (response-for msg {:status                #{:done :error}
-                                                   :error                 error
                                                    :unresolved-middleware unresolved}))))))
 
       "ls-middleware"
-      (do
-        (t/send transport (response-for msg :middleware (mapv str (:stack @*state*))))
-        (t/send transport (response-for msg {:status :done})))
+      (t/send transport (response-for msg
+                                      :middleware (mapv str (:stack @*state*))
+                                      :status :done))
 
       (h msg))))
+
+(def ^{:private true} add-swap-ops
+  {:requires {"middleware" "a list of middleware"}
+   :optional {"extra-namespaces" "a list of extra namespaces to load. This is useful when the new middleware feature deferred loading"}
+   :returns  {"status"                "`done`, once done, and `error`, if there's any problems in loading a middleware"
+              "unresolved-middleware" "List of middleware that could not be resolved"}})
 
 (set-descriptor! #'wrap-dynamic-loader
                  {:requires #{#'middleware.session/session}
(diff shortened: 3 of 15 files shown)


bbatsov:
> Instead of loading all middlewares, and have them manage their own deferred loading, have an "entry point" middleware, which doesn't depend on any of the others, but has a registry of which opts correspond to which middleware.

Yeah, that's one idea I like very much. One big issue with the deferred loading approach taking currently by `cider-nrepl` is that it adds a lot of complexity - you have decouple descriptors from the middleware, middleware loading doesn't work the way it would normally work and so on.

src/clojure/nrepl/middleware/dynamic_loader.clj
```diff
@@ -0,0 +1,108 @@
...
+  When combined with the sideloader, this could be used to inject middlewares
+  that are unknown to the server prior to connection."
+  {:author "Shen Tian", :added "0.8"}
+  (:require [clojure.string :as str]
+            [nrepl.middleware :as middleware :refer [set-descriptor!]]
+            [nrepl.middleware.session :as middleware.session]
+            [nrepl.misc :refer [response-for]]
+            [nrepl.transport :as t]))
+
+(defn require-resolve
```
bbatsov, line 15: We already have `requiring-resolve` in `nrepl.misc`.

src/clojure/nrepl/middleware/dynamic_loader.clj
```diff
@@ -0,0 +1,108 @@
...
+      (do
+        (t/send transport (response-for msg :middlewares (mapv str (:stack @*state*))))
+        (t/send transport (response-for msg {:status :done})))
+
+      (h msg))))
+
+(set-descriptor! #'wrap-dynamic-loader
+                 {:requires #{#'middleware.session/session}
+                  :expects  #{}
+                  :handles  {"add-middlewares"
```
bbatsov, line 105: Seems you forgot to list the other ops here.

src/clojure/nrepl/middleware/dynamic_loader.clj
```diff
@@ -0,0 +1,108 @@
...
+(defn wrap-ping
+  [h]
+  (fn [{:keys [op transport] :as msg}]
+    (case op
+      "ping"
+      (t/send transport (response-for msg {:pong "pong"
+                                           :status :done}))
+      (h msg))))
+
+(set-descriptor! #'wrap-ping
```
bbatsov, line 34: I guess that's some test middleware, right?

src/clojure/nrepl/middleware/dynamic_loader.clj
```diff
@@ -0,0 +1,108 @@
...
+         :alt-cl?      (boolean alt-cl)
+         :before-stack (map str (:stack @*state*))})
+      (finally
+        (.setContextClassLoader (Thread/currentThread) ctxcl)))))
+
+(defn wrap-dynamic-loader
+  [h]
+  (fn [{:keys [op transport session middlewares] :as msg}]
+    (case op
+      "add-middlewares"
```
bbatsov, line 78: The plural of `middleware` is `middleware`. :-) https://english.stackexchange.com/questions/257120/middleware-vs-middlewares

bbatsov:
I'm impressed with the simplicity of the solution! Great work!

I'm also tempted to add a ping middleware to nREPL, as a simple example for a middleware. :D Reminds me of SLIME and SWANK. :D 

It'd be nice if you incorporated more of your descriptions from the ticket in the middleware's implementation as docstrings and comments.

src/clojure/nrepl/middleware/dynamic_loader.clj
shen-tian: Yes. Will add these in when I do the documentation

src/clojure/nrepl/middleware/dynamic_loader.clj
shen-tian: Yes. Will remove before asking for a full review

src/clojure/nrepl/middleware/dynamic_loader.clj
shen-tian: Will change :)

shen-tian: Oh dear. Happy to change in new code, but "middlewares" is used quite liberally in the codebase + docs already.

In particular, `nrepl.server/default-middlewares` is an important part of the public API.

So it might be worth conceding to `middlewares` for nREPL purposes?

bbatsov: I'll think about this. We've been inconsistent with the wording indeed, but I generally prefer to stick to the rules of natural languages. We can always have an alias for `default-middlewares` if it's our only public API transgression.

src/clojure/nrepl/middleware/dynamic_loader.clj
shen-tian: Switched to that

src/clojure/nrepl/middleware/dynamic_loader.clj
shen-tian: I'd be supportive of that. I'll stick the the plural "middleware" in this PR, and we can switch everything else in a separate one for sake of commit history?

shen-tian: Done

shen-tian:
Thanks for the comments! Made some quick changes in response. I take it you are, in principle, supportive of the design?

I'd like to do a bit more testing of the core capabilities, while I finish off the PR. More comments from the @nrepl team welcome

shen-tian:
Just hit a good milestone:

1. Started a plain nREPL
2. Connected to it using a client that supported dynamic loading and sideloading
3. Injected `cider-nrepl` and `refactor-nrepl` and disconnected
4. `cider-connect` from Emacs and it was ready to go.

A few new realisations/discoveries:

- Ironically(?) deferred loading in both `cider-nrepl` and `refactor-nrepl` made this harder, as the said deferred loading happened outside of an `eval` or `add|swap-middleware`, so didn't trigger the sideloader. You can work around this by pre-loading the namespaces though (thus defeating the point of deferred loading). I can think of two ways to make this less hacky: 
  1. Introduce some sort of `while-sideloading` macro, and update relevant middleware to run the deferred loading inside of it. This should be simple (and non-intrusive) enough to do, but raising the question of whether it's desirable to run the sideloader all the time. Deferred loading would most likely happen at an unpredictable time.
  2. Formalise the hack by add an optional `:extra-ns` slot to `add|swap-middleware` messages. The client would have to know which (deferred) namespaces goes with which middleware, but it's not hard to figure out.
- Changing the middleware setup destroys any local states contained within the middleware themselves. Only the sessions remain. I came across this because it broke the sideloader's pending atom. I assume we have no guarantee of how stateful middleware in general are, so this might mean we need to be careful with the boot sequence. (This does feel like a disadvantage of the middleware architecture, v.s. an interceptor one: that you can re-assemble interceptors, but the composition of middleware fn's kinda bakes them all together?) 
- With all this, it's likely the dynamic loader doesn't remove the need for deferred loading. But it does clarify it's role (optimise startup time) v.s. dynamic loader (post startup configuration). 

Will ponder more about this, but quite glad to see this work mostly as envisioned.

doc/modules/ROOT/pages/design/middleware.adoc
```diff
@@ -359,3 +359,21 @@ See link:../building_clients.html[Building Clients] on how to implement the serv
 Once the sideloader has been started, it can be triggered by an ordinary `eval` or `load-file` operation. The nREPL server will first try to find a resource/class on the classpath of its own JVM. Failing this, it will attempt to request it from the nREPL client by the above described mechanism. Should this still fail, expect a `ClassNotFoundException` as usual.
 
 TIP: Once a class has been loaded, it will become available to all sessions on nREPL server.
+
+== Dynamic middleware loading
+
+NOTE: Sideloading support was added in nREPL 0.8 and the API is still
```
bbatsov, line 365: Sideloading -> dynamic loading :-)

doc/modules/ROOT/pages/design/middleware.adoc
```diff
@@ -359,3 +359,21 @@ See link:../building_clients.html[Building Clients] on how to implement the serv
...
+query and change the middleware stack the nREPL server is using. This is especially
+powerful when combined with sideloading, as it allows a client to configure the 
+server after connecting, and provides an alternative to having to specify the middleware required by the client as startup time.
+
+This introduces three new operations:
+
+* `ls-middleware`, to return a list of active middleware, ordered from inside outwards.
+* `add-middleware`, which adds a middleware to the stack. Optionally, a list of `extra-namespaces` could be provided for loading. This is useful when adding middleware that implement some form of deferred loading. Examples include `cider-nrepl` and `refactor-nrepl`. In these cases, some of the required namespaces might only be loaded upon first use, which may occur outside of a sideloading session, and thus fail. This feature allows us to pre-load namespaces when we add a middleware. If loading of any particular middleware fails, the stack will be unchanged.
+* `swap-middleware`, similar to `add-`, but replaces all existing middleware. Note that this _may_ remove the `dynamic-loader` itself.
```
bbatsov, line 377: I couple of examples of using `add/swap`-middleware would be useful I guess. I think that using `swap-middleware` will be particularly tricky for some people. I'm not sure if this should go here or in the building clients section.

src/clojure/nrepl/misc.clj
```diff
@@ -60,3 +60,20 @@
         (require (symbol (namespace sym)))
         (resolve sym)
         (catch Exception _))))
+
+(defmacro with-session-classloader
```
bbatsov, line 64: You'll have to make sure that hotloading still works. I'm always afraid we might break it when fiddling with the classloader, as it doesn't really have any tests.

src/clojure/nrepl/server.clj
```diff
@@ -82,27 +83,41 @@
 
 (def default-middlewares
   "Middleware vars that are implicitly merged with any additional
-   middlewares provided to nrepl.server/default-handler."
+   middleware provided to nrepl.server/default-handler.
 
+   Note: this is part of the de-facto API, thus the name of the var should be
```
bbatsov, line 88: I'd just add one more `def` with the legacy name and mention it exists only for backwards compatibility.

test-resources/ping.clj
```diff
@@ -0,0 +1,39 @@
+(ns ping
+  "This provides an example of a middleware, including a deferred handler
+  that's not loaded until called. This is representative of how cider-nrepl
+  and refactor-nrepl handles deferred loading"
```
bbatsov, line 4: Btw, I think `refactor-nrepl` doesn't really defer anything, although this might have changed. It actually provides all its ops in a single middleware, which is somewhat different from what nREPL and `cider-nrepl` do. End of random comment. :-)

test/clojure/nrepl/describe_test.clj
```diff
@@ -1,18 +1,20 @@
 (ns nrepl.describe-test
   {:author "Chas Emerick"}
-  (:require
-   [clojure.test :refer :all]
-   [nrepl.core :as nrepl]
-   [nrepl.core-test :refer [def-repl-test repl-server-fixture project-base-dir]]
-   [nrepl.middleware :as middleware]
-   [nrepl.version :as version]))
+  (:require [clojure.test :refer [is testing use-fixtures]]
```
bbatsov, line 3: I think you reformatted this accidentally - all the code sticks to the other style.

test/clojure/nrepl/describe_test.clj
```diff
@@ -1,18 +1,20 @@
...
+            [nrepl.core :as nrepl]
+            [nrepl.core-test :refer [def-repl-test repl-server-fixture]]
+            [nrepl.middleware :as middleware]
+            [nrepl.server :as server]
+            [nrepl.version :as version]))
 
 (use-fixtures :once repl-server-fixture)
 
 (def ^{:private true} op-names
```
bbatsov, line 12: This can even be some public API somewhere - it's useful to be able to tell apart built-in from third-party ops. I needed to do this in the docs generator of `cider-nrepl` (for instance).

bbatsov:
Looks solid to me! Apart from my small remarks, I think the biggest thing missing is more documentation for client authors.

src/clojure/nrepl/misc.clj
shen-tian: hotloading as in what Pomegranate does?

Dynamic loader capabilities is fairly separate from sideloading, if that's what you are concerned about. This macro is from the changes made in #162, which has its origins in unrepl. Both sideloading + dynamic-loader while sideloading are covered in tests.

I was looking at #113 : is that the kind of breakage you are concerned about? Maybe the the thing to do is to add more tests that we are concerned about?

test-resources/ping.clj
shen-tian: I found out the hard way what does/doesn't do deferred loading :)

This handles deferred loading for `clean-ns`, for example.

https://github.com/clojure-emacs/refactor-nrepl/blob/dbafd6e686402dc020d7026e21ee556812fc51da/src/refactor_nrepl/middleware.clj#L109-L111

not all the ops do this, but a bunch of them do.

test/clojure/nrepl/describe_test.clj
shen-tian: do you mean the `clojure.test :refer :all` part or the first spec on the line after `(:require .. `part?

test/clojure/nrepl/describe_test.clj
shen-tian: This should be just "the list of ops provided by default middleware"? It feels reasonably safe.

Are you saying other libs have a dependency on `describe-test`?

src/clojure/nrepl/misc.clj
bbatsov: > hotloading as in what Pomegranate does?

Yep. It didn't work with nREPL until we did some classloader modifications a while ago, so I'm always wary of us breaking this again. 

> I was looking at #113 : is that the kind of breakage you are concerned about? Maybe the the thing to do is to add more tests that we are concerned about?

Yeah, we should probably tackle this at some point. Initially we skipped the tests because of the many moving pieces there (hotloading library, build tool).

test-resources/ping.clj
bbatsov: Got it. I meant they don't use the same deferring that cider-nrepl does. Before we added the current macros, cider-nrepl was also just using `delay` for some requires.

test/clojure/nrepl/describe_test.clj
bbatsov: I mean having like-break after `:require`.

test/clojure/nrepl/describe_test.clj
bbatsov: See https://github.com/clojure-emacs/cider-nrepl/blob/master/maint/cider/nrepl/impl/docs.clj#L41

src/clojure/nrepl/server.clj
shen-tian: done

test-resources/ping.clj
shen-tian: Right.

test/clojure/nrepl/describe_test.clj
shen-tian: Reverted.

test/clojure/nrepl/describe_test.clj
shen-tian: Ah I see. So sounds like we should add `built-in-ops` to `nrepl.middleware`?

src/clojure/nrepl/misc.clj
shen-tian: I've made the macro a bit safer. If there's no `:classloader` under (meta session), then it's pretty much a `do` block.

Currently, only the sideloader sets `:classloader`, so it's quite unlikely to make any breaking changes. I'd be happy to chat more about tests/classloader etc. separately though?

doc/modules/ROOT/pages/design/middleware.adoc
shen-tian: Added section in "building clients"

bbatsov:
@shen-tian I think we're good to merge. Ideally you should clean up the commit history a bit by squashing related commits together (e.g. - it seems the first 9-10 commits can be reduced to one).

bbatsov:
Ah, actually you'll also have to run `lein docs` to regen the ops page.

doc/modules/ROOT/pages/building_clients.adoc
```diff
@@ -116,6 +116,58 @@ This unblocks the `eval` command, which returns the value `Qaz` in the final mes
...
+is to request the extra namespace to be loaded at `add-middleware` time too:
+
+[source,clojure]
+----
+;; after starting the sideloader...
+{:op "add-middleware"
+ :middleware ["cider.nrepl.middleware/wrap-version"]
+ :extra-namespaces ["cider.nrepl.middleware.version"]}
+ ;; now, the following shoud work
+ {:op "cider-version"}
```
liquidz, line 161:
```suggestion
;; now, the following shoud work
{:op "cider-version"}
```

shen-tian:
> @shen-tian I think we're good to merge. Ideally you should clean up the commit history a bit by squashing related commits together (e.g. - it seems the first 9-10 commits can be reduced to one).

Will do. Was assuming you'll squish :)

bbatsov:
From the web UI I can only squash all commits in 1. I think it'd better to do manual squashing in this case, as there 3-4 meaningful changes apart from the dynamic loader itself.

shen-tian:
@bbatsov Squished it down to 6 commits, including regen of the ops page.

bbatsov:
🎉

PR #187, opened 2020-05-12
Update tests to use Kaocha, and some other enhancements. 

shen-tian:
I've been using Kaocha locally while working on nREPL, and it's been a much better experience, especially with output capturing + watch. This PR does the following:

- Adding Kaocha, while maintaining compatibility for `lein test` for Clojure 1.7 and 1.8. The makefile has been updated to do the switch automatically. I've also added a hook for manually clearing sessions and killing their threads after a test run, as this was causing the JVM to crash fairly quickly in "watch" mode because we start hundreds of thread per test run. All this is documented, and I've expanded the "hacking on nREPL" section.
- Bye java 13 hi java 14.
- Fixed the `edn-test` that were failing on CI at times. Letting each test run pick its own port helps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,9 +25,9 @@ executors:
     docker:
       - image: circleci/clojure:openjdk-11-lein-2.9.1
     <<: *defaults
-  openjdk13:
+  openjdk14:
     docker:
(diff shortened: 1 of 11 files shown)


bbatsov:
Thanks!

PR #189, opened 2020-05-16
Adding some docs that weren't added properly from #173

shen-tian:
The `read-cond` docs were added to the ops.adoc directly, so we lost them in the next regen. This adds them to the descriptors.


bbatsov:
Thanks!

PR #191, opened 2020-05-17
Avoid writing partial messages in the bencode transport

### LINKED ISSUE #17: Transport/send hangs when sending non-bencodable information outside of :value slot
Situation: Writing a CIDER middleware, trying to `transport/send` a response message via the standard bencode transport. The response message has a non-bencodable item outside of the :value slot.

Expected behavior: `transport/send` will bubble up the IllegalArgumentException thrown in c.t.n.bencode/write-bencode due to the non-bencodable item.

Actual behavior: `transport/send` silently fails and hangs. No response is sent over the wire (verified by monitoring communication going across nREPL's port). No exception is bubbled up to the CIDER middleware. Subsequent calls to nREPL (to evaluate code for example) do not return. Once an interrupt is sent, the subsequent calls do return correct values, but the non-bencodable response never appears.

I'm guessing there's some relationship to NREPL-30 (http://dev.clojure.org/jira/browse/NREPL-30)? Please see https://github.com/clojure-emacs/cider-nrepl/issues/332 for further info as well. Let me know if there's anything I can do to debug this issue better.

Thanks for your help,
Sanjay

(from https://dev.clojure.org/jira/browse/NREPL-81)

shen-tian:
Fixes #17 

This adds a `write-bencode*` fn to the bencode ns. It first writes to a buffer, and only once successful, write to the provided output stream.

I don't see any reason we'd ever what to send half a message down bencode, so this feels like it's a reasonable default behaviour.

### DIFF
--- a/test/clojure/nrepl/bencode_test.clj
+++ b/test/clojure/nrepl/bencode_test.clj
@@ -8,7 +8,7 @@
 
 
 (ns nrepl.bencode-test
-  (:require [clojure.test :refer [are deftest is]]
+  (:require [clojure.test :refer [are deftest is testing]]
             [nrepl.bencode :as bencode :refer [read-bencode
                                                read-netstring
                                                write-bencode
@@ -196,3 +196,10 @@
                (decode :reader read-bencode)
                (get "data")
                seq)))))
+
+(deftest unwritable-values
+  (testing "write-bencode writes eagerly"
+    (let [out (ByteArrayOutputStream.)]
+      (is (thrown? IllegalArgumentException
+                   (write-bencode out {"obj" (Object.)})))
+      (is (= "d3:obj" (String. (.toByteArray out)))))))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#125](https://github.com/nrepl/nrepl/issues/125): The built-in client supports `greeting-fn`.
 * [#126](https://github.com/nrepl/nrepl/issues/126): The built-in client exits with an error message when the tty transport is selected. It used to fail silently. This was never supported.
 * [#113](https://github.com/nrepl/nrepl/issues/113): Fix an issue with hotloading using Pomegranate in Leiningen.
+* [#17](https://github.com/nrepl/nrepl/issues/17): It was possible for the bencode transport to write partial messages if a middleware tries to write something unencodable. This could cause the client or server to hang.
 
 ## 0.7.0 (2020-03-28)
 
--- a/src/clojure/nrepl/transport.clj
+++ b/src/clojure/nrepl/transport.clj
@@ -10,7 +10,7 @@
    nrepl.version)
   (:import
    clojure.lang.RT
-   [java.io EOFException PushbackInputStream PushbackReader]
+   [java.io ByteArrayOutputStream EOFException PushbackInputStream PushbackReader]
    [java.net Socket SocketException]
    [java.util.concurrent BlockingQueue LinkedBlockingQueue SynchronousQueue TimeUnit]))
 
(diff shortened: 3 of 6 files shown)


test/clojure/nrepl/bencode_test.clj
```diff
@@ -196,3 +197,15 @@
                (decode :reader read-bencode)
                (get "data")
                seq)))))
+
+(deftest unwritable-values
+  (testing "write-bencode writes eagerly"
+    (let [out (ByteArrayOutputStream.)]
+      (is (thrown? IllegalArgumentException
+                   (write-bencode out {"obj" (Object.)})))
+      (is (= "d3:obj" (String. (.toByteArray out))))))
```
bbatsov, line 206: I guess you can also use `byte-array`. It should behave the same way, right?

bbatsov:
> I don't see any reason we'd ever what to send half a message down bencode, so this feels like it's a reasonable default behaviour.

Same here. I'm not sure we need a separate function for this. We just have to draw a clear line between a partial message and a message streamed in chunks, as I think the current wording might confuse some people.  

> It first writes to a buffer, and only once successful, write to the provided output stream.

What's the performance overhead when dealing with bigger output?

test/clojure/nrepl/bencode_test.clj
shen-tian: Not sure if you can stream to a `byte-array`. Also seems the fixed size part might be a bad fit for what we need to do here?

shen-tian:
Are you suggesting that we change the default behaviour of `bencode/write-bencode` to verify a message first, then stream it? I don't see an easy way to do that while preserving the rather elegant API of that ns.

I guess we can move that logic inside of `transport/bencode`? 

In terms of performance, I can think up some synthetic tests, but I'd be surprised if it has any non-trivial performance impact for nREPL message. Furthermore, I assume the performance bottleneck will always be what we can print onto screen on the other side?

test/clojure/nrepl/bencode_test.clj
bbatsov: I meant Clojure has a `byte-array` function which I think does the same thing as `.toByteArray`. :-)

bbatsov:
> Are you suggesting that we change the default behaviour of bencode/write-bencode to verify a message first, then stream it? I don't see an easy way to do that while preserving the rather elegant API of that ns.

Well, when you put it like this it doesn't seem like a good idea anymore. 

> I guess we can move that logic inside of transport/bencode?

That's an option, but I guess we should aim for the simplest possible change, as so far the impact of this bug has been relatively small. 

> In terms of performance, I can think up some synthetic tests, but I'd be surprised if it has any non-trivial performance impact for nREPL message. Furthermore, I assume the performance bottleneck will always be what we can print onto screen on the other side?

Yeah, I guess you're right.

src/clojure/nrepl/bencode.clj
```diff
@@ -323,6 +323,16 @@
       (or (nil? thing) (coll? thing) (.isArray (class thing))) :list
       :else (type thing))))
 
+(defn write-bencode*
+  "Similar to `write-bencode`, except it will only write if the whole `thing`
+   is writable. In practice, it avoids sending partial messages, which can
```
bbatsov, line 328: We should expand on the definition of "writable" here - e.g. with some examples.

shen-tian:
> > Are you suggesting that we change the default behaviour of bencode/write-bencode to verify a message first, then stream it? I don't see an easy way to do that while preserving the rather elegant API of that ns.
> 
> Well, when you put it like this it doesn't seem like a good idea anymore.
> 
> > I guess we can move that logic inside of transport/bencode?
> 
> That's an option, but I guess we should aim for the simplest possible change, as so far the impact of this bug has been relatively small.

Sounds like leaving the bencode ns alone and adding this check/extra step to the bencode transport is probably the best approach then.

BTW, this problem doesn't have too many major day to day impacts, especially to end users, but is always comes up as a head scratcher to people working on it, because how it corrupts the transport, so I definitely think it's worth fixing, even if it means touching how the transport works.

bbatsov:
Agreed.

test/clojure/nrepl/bencode_test.clj
shen-tian: Tried, and afraid it doesn't. ByteArrayOutputStream doesn't implement ISeq.

shen-tian:
Done. I left a test in `bencode-test` to show the streaming behaviour.

bbatsov:
Thanks! Btw, does the EDN transport suffer from a similar problem with partial messages?

shen-tian:
> Thanks! Btw, does the EDN transport suffer from a similar problem with partial messages?

No. The printer ensures that everything will be converted to EDN (even if there's no readers for things like `#function` on the other side), so it becomes a read rather than write problem.

PR #192, opened 2020-05-23
Change print tests to not reply on eval and session

shen-tian:
Most of the printing specific tests in `core-test` can be performed without rest of the default middleware stack. This is an attempt to do so, and in a new print specific test ns.

There's a few notable things:

- This bypasses eval by placing the thing we want to print directly into the `:value` slot, and having an echo handler send it back for printing.
- We use a simple test-transport to collect return messages.
- Since we are not sending messages across the wire, the "session variable" tests can be simplified to just `binding`, which is more explicit, and doesn't depend on test order.
- We don't need to clean out session/message IDs, because those things don't exist in these tests.

The few print tests left in core test the interaction of multiple middleware.

I think we can do something similar to `caught` tests, if this is an improvement.


test/clojure/nrepl/middleware/print_test.clj
```diff
@@ -0,0 +1,221 @@
+(ns nrepl.middleware.print-test
+  "Tests for the print ns."
```
bbatsov, line 2: ns -> middleware

test/clojure/nrepl/middleware/print_test.clj
```diff
@@ -0,0 +1,221 @@
...
+  [{:keys [transport] :as msg}]
+  (t/send transport (select-keys msg (::print/keys msg))))
+
+(defn test-transport [queue]
+  (t/fn-transport
+   nil
+   #(swap! queue conj %)))
+
+(defmacro testing-print
+  "Marco for print tests. Exposes a function `handle`, which takes a msg.
```
bbatsov, line 20: marco -> macro

test/clojure/nrepl/middleware/print_test.clj
```diff
@@ -0,0 +1,221 @@
...
+   the `:value` in the message will be pritned, using whatever extra print
```
bbatsov, line 21: pritned -> printed

bbatsov:
Great work overall! I think it'd might be a good idea to incorporate some of the explanations from the PR in the test ns (e.g. as comments and docstrings).

PR #193, opened 2020-05-25
Make print middleware configurable via global config file

### LINKED ISSUE #93: Make the default printer configurable

Currently the default printer is hard-code in `pr-values` and you can also override it in eval messages with the `printer` key. It might be a good idea to be able to override it globally using nREPL's config files. 

We should add a `:printer` and `:printer-options` to the known config keys that can be passed via `.nrepl.edn`. Printer would be a reference to a var and printer-opts to an options map.

shen-tian:
`warp-print` actually has some fairly complex resolution order for `print-fn`, which cannot be directly overridden by a message directly. It looks something like:

1. request inferred fn
2. `:print-fn` in response
3. `*print-fn*` in session
4. `pr`-esque function

(I'm not 100% sure this is right).

This approach respects that by placing the weight of config file values quite high up, between 1) and 2), rather than  between 3 and 4, which I originally imagined. I think this is quite easy to reason about, as the config file is closer to what the client can do.

This can also act as a general approach for configuring middleware using the config file.

Does this seems like a good approach? If so, the code could be cleaned up a bit, and documentation updated.

Closes #93

### DIFF
--- a/src/clojure/nrepl/middleware/print.clj
+++ b/src/clojure/nrepl/middleware/print.clj
@@ -5,6 +5,7 @@
    :added  "0.6"}
   (:refer-clojure :exclude [print])
   (:require
+   [nrepl.config :as config]
    [nrepl.middleware :refer [set-descriptor!]]
    [nrepl.misc :as misc]
    [nrepl.transport :as transport])
@@ -59,6 +60,9 @@
 
 (def configuration-keys
   [::print-fn ::stream? ::buffer-size ::quota ::keys])
+
+(def providable-keys
+  [::print ::options ::stream? ::buffer-size ::quota ::keys])
 
 (defn- to-char-array
   ^chars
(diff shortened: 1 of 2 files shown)


shen-tian:
Separate question I picked up on:

Existing docs say:

> Note well that the printing function is expected to not interact with *out* or *err* at all, even rebinding them (e.g. via with-out-str). Output may be printed to either of those streams during its operation...

> Further, note that clojure.pprint/pprint rebinds *out* internally (even when using its explicit writer arity). It is not possible to prevent the interleaving of output when using clojure.pprint.

This seems to suggest `nrepl.util.print/pprint` is not a safe option for `:nrepl.middleware.print/print`? Not sure if @cichli is around to comment?

bbatsov:
> This approach respects that by placing the weight of config file values quite high up, between 1) and 2), rather than between 3 and 4, which I originally imagined. I think this is quite easy to reason about, as the config file is closer to what the client can do.

The problem with this is that I think most people would expect if they set some binding in the REPL explicitly it would take precedence over the server configuration. 

> This seems to suggest nrepl.util.print/pprint, is not a safe option for :nrepl.middleware.print/print, and this was designed for fipp or zprint. Not sure if @cichli is around to comment?

He hasn't been around much lately, but he wrote `nrepl.util.print/pprint`, so I guess it gets the job done. (I simply copied it from `cider-nrepl` recently, so there's some default function in nREPL that people can use without the need for external deps).

src/clojure/nrepl/middleware/print.clj
```diff
@@ -60,6 +61,9 @@
 (def configuration-keys
   [::print-fn ::stream? ::buffer-size ::quota ::keys])
 
+(def providable-keys
```
bbatsov, line 64: I guess the difference between config-keys and providable keys will confuse many people.

test/clojure/nrepl/middleware/print_test.clj
```diff
@@ -231,3 +232,33 @@
...
+  (testing-print "config file can be used to set streaming"
+    (with-redefs [config/config {::print/stream? 1
+                                 ::print/buffer-size 8}]
+      (is (= [{:value "(0 1 2 3"}
+              {:value " 4 5 6 7"}
+              {:value " 8 9)"}
+              {}]
+             (handle {:value (range 10)}))))))
+
+(deftest print-utils
```
bbatsov, line 260: Probably you can submit this test in a separate PR, as it's unrelated to this one.

shen-tian:
> > This approach respects that by placing the weight of config file values quite high up, between 1) and 2), rather than between 3 and 4, which I originally imagined. I think this is quite easy to reason about, as the config file is closer to what the client can do.
> 
> The problem with this is that I think most people would expect if they set some binding in the REPL explicitly it would take precedence over the server configuration.
> 

Fair expectation. The tricky bit is how `::print` + `::options` determine `::print-fn`, which can also be explicitly set in the session. Will need to have a think about how to do that in a sane fashion.

bbatsov:
After 4 years with no activity here, I guess we can close this PR.

PR #194, opened 2020-06-02
Fix load-file returns

eerohele:
The "Built-in Ops" doc says that the `load-file` op returns `*ns*` under the `:ns` key after successful evaluation. However, according to the change log, the `:ns` key for `load-file` was removed in v0.2.8. There's also [a comment in the code](https://github.com/nrepl/nrepl/blob/23361dd2792338310c048df22a5995c80bf4c37d/src/clojure/nrepl/middleware/load_file.clj#L92-L94) that says it was removed to avoid confusion.


bbatsov:
You'll also have to update the middleware descriptor, as the docs are auto-generated.

eerohele:
Ah, of course. Done.

bbatsov:
Thanks!

PR #199, opened 2020-06-21
Fix completion candidate types

eerohele:
Previously, `(ns-vars)` and `(ns-public-vars)` returned symbols instead of vars. As a result, `(var-type)` ended up calling `(meta)` on a symbol, which always returns `nil`.

Now, both `(ns-public-vars)` and `(ns-vars)` return vars instead of symbols.

### DIFF
--- a/src/clojure/nrepl/util/completion.clj
+++ b/src/clojure/nrepl/util/completion.clj
@@ -24,13 +24,12 @@
 (defn ns-public-vars
   "Returns a list of potential public var name completions for a given namespace"
   [ns]
-  (keys (ns-publics ns)))
+  (vals (ns-publics ns)))
 
 (defn ns-vars
   "Returns a list of all potential var name completions for a given namespace"
   [ns]
-  (for [[sym val] (ns-map ns) :when (var? val)]
-    sym))
+  (filter var? (vals (ns-map ns))))
 
 (defn ns-classes
   "Returns a list of potential class name completions for a given namespace"
@@ -102,15 +101,12 @@
 
 ;;; Candidates
 
-(defn var-type [var]
-  (let [m (meta var)]
-    (cond
-      (:macro m) :macro
-      (:arglists m) :function
-      :else :var)))
-
 (defn annotate-var [var]
-  {:candidate (name var) :type (var-type var)})
+  (let [{macro :macro arglists :arglists var-name :name} (meta var)
+        type (cond macro :macro
+                   arglists :function
+                   :else :var)]
+    {:candidate (name var-name) :type type}))
 
 (defn annotate-class
   [cname]
--- a/test/clojure/nrepl/util/completion_test.clj
+++ b/test/clojure/nrepl/util/completion_test.clj
@@ -1,5 +1,6 @@
 (ns nrepl.util.completion-test
-  (:require [clojure.test :refer :all]
+  (:require [clojure.set :as set]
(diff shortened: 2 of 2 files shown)


eerohele:
Looks like you can't call `(symbol)` on a var on older Clojure versions... TIL.

I replaced it with a call to `.sym`. Not sure if that's the best solution, though. Suggestions?

src/clojure/nrepl/util/completion.clj
```diff
@@ -24,13 +24,13 @@
...
   [ns]
-  (keys (ns-publics ns)))
+  (vals (ns-publics ns)))
 
 (defn ns-vars
   "Returns a list of all potential var name completions for a given namespace"
   [ns]
-  (for [[sym val] (ns-map ns) :when (var? val)]
-    sym))
+  (for [[_ val] (ns-map ns) :when (var? val)]
```
bbatsov, line 32: Seems to me this code will look better if it's just a `filter`.

src/clojure/nrepl/util/completion.clj
```diff
@@ -110,7 +110,7 @@
       :else :var)))
 
 (defn annotate-var [var]
-  {:candidate (name var) :type (var-type var)})
+  {:candidate (name (.sym var)) :type (var-type var)})
```
bbatsov, line 113: You can also get the name using `meta`.

eerohele:
Thanks for the comments, I force-pushed the changes you proposed. Although now I'm not sure whether `var-type` is a good name for that function now that it doesn't take a var as an arg...

eerohele:
Looks like one of the tests timeouted.

eerohele:
>Although now I'm not sure whether var-type is a good name for that function now that it doesn't take a var as an arg...

I guess one option would be to pull `(var-type)` into `(annotate-var)`. As in, something like:

```clojure
(defn annotate-var [var]
  (let [{macro :macro arglists :arglists var-name :name} (meta var)
        type (cond macro :macro
                   arglists :function
                   :else :var)]
    {:candidate (name var-name) :type type}))
```

bbatsov:
> I guess one option would be to pull (var-type) into (annotate-var). As in, something like:

Yeah, that's a good idea.

eerohele:
Done. I also changed the tests to use the same `(some #{:foo} bar)` idiom as some of the other tests.

eerohele:
`nrepl.core-test/sideloader-test` seems to fail sporadically.

eerohele:
Also, one additional thing that came to mind is that `(name var-name)` will throw an NPE of `var-name` is `nil`, but I'm not sure whether `:name` on meta can ever be `nil`. Probably not?

bbatsov:
> nrepl.core-test/sideloader-test seems to fail sporadically.

It'd be best to open a separate ticket to track this.

> Also, one additional thing that came to mind is that (name var-name) will throw an NPE if var-name is nil, but I'm not sure whether :name on meta can ever be nil. Probably not?

I think this can never be `nil`.

eerohele:
>>nrepl.core-test/sideloader-test seems to fail sporadically.
>
>It'd be best to open a separate ticket to track this.

I opened https://github.com/nrepl/nrepl/issues/200.

And thanks!

bbatsov:
I've cut a new alpha. The completions logic is definitely rough around the edges and has some room for improvement. Thanks for testing and improving it!

Issue #201, opened 2020-06-21
Completion candidate list can contain duplicate values


eerohele:
## Expected behavior

Completion candidate list doesn't contain duplicate values.

## Actual behavior

Completion candidate list contains duplicate values.

## Steps to reproduce the problem

```clojure
=> (completions "String/v")
({:candidate "String/valueOf", :type :static-method}
 {:candidate "String/valueOf", :type :static-method}
 {:candidate "String/valueOf", :type :static-method}
 {:candidate "String/valueOf", :type :static-method}
 {:candidate "String/valueOf", :type :static-method}
 {:candidate "String/valueOf", :type :static-method}
 {:candidate "String/valueOf", :type :static-method}
 {:candidate "String/valueOf", :type :static-method}
 {:candidate "String/valueOf", :type :static-method})
```

### Clojure version

1.10.0

### Java version

11

### Operating system

macOS

---

The easiest solution would be to `dedupe` the current return value of `nrepl.util.completion/completions`. Another possibility would be to use `sorted-set-by`, but I don't know whether the Bencode encoder will be able to handle that. I also don't know which is faster, or whether it even matters here.

I can create a PR if you want, but I'll be travelling for a while, so it might be a while before I get to it. Wanted to create an issue so that I don't forget.

bbatsov:
Hmm, I recall I had seen this a while ago and supposedly I had added some uniqueness check somewhere. The Java completion code is currently just a copy of the legacy clojure-complete Java completion (minus 2-3 bugs) and I didn't try much to understand/improve it. 

> Another possibility would be to use sorted-set-by, but I don't know whether the Bencode encoder will be able to handle that.

Sets become just regular bencode dictionaries in the end. Still, it seems to me that we should actually find out what's causing those duplicates in the first place.

eerohele:
>Hmm, I recall I had seen this a while ago and supposedly I had added some uniqueness check somewhere.

It looks like `ns-java-methods` calls `distinct`, so that might be it.

I think the issue is simply that `static-members` currently returns every overload of every static method (as with `String/valueOf`). Simply adding a `distinct` call (as with `ns-java-methods`) would probably be the simplest fix.

PR #202, opened 2020-06-22
Ensure distinct static member completion candidates #201

eerohele:
Static methods such as `String/valueOf` have many overloads. Prior to this commit, the `static-members` function returned one completion candidate for each overload.

### DIFF
--- a/src/clojure/nrepl/util/completion.clj
+++ b/src/clojure/nrepl/util/completion.clj
@@ -52,8 +52,9 @@
 (defn static-members
   "Returns a list of potential static members for a given class"
   [^Class class]
-  (for [member (concat (.getMethods class) (.getDeclaredFields class)) :when (static? member)]
-    (.getName ^Member member)))
+  (distinct
+   (for [member (concat (.getMethods class) (.getDeclaredFields class)) :when (static? member)]
+     (.getName ^Member member))))
 
 (defn path-files [^String path]
   (cond (.endsWith path "/*")
--- a/test/clojure/nrepl/util/completion_test.clj
+++ b/test/clojure/nrepl/util/completion_test.clj
@@ -45,6 +45,7 @@
            (candidates "java.lang.System/out")))
 
     (is (some #{"String/valueOf"} (candidates "String/")))
+    (is (every? #(= 1 %) (vals (frequencies (candidates "String/")))))
 
     (is (not (some #{"String/indexOf" ".indexOf"} (candidates "String/")))))
 


bbatsov:
I'm guessing there might be a similar problem for instance methods as well.

arichiardi:
Hey folks, I am checking exactly that part in orchard now and a lot of work went into polishing that.

I know this implementation is different but probably it is worth having a look there 😁

bbatsov:
@arichiardi The code completion is actually in `compliment` and it's way more sophisticated than anything I'd like to bundle with nREPL. I didn't want to duplicate a great library, so instead I've opted for the simplest thing that adds some value. I've started with `clojure-complete` and I've cleaned it up and improved a bit.

bbatsov:
> I'm guessing there might be a similar problem for instance methods as well.

Forget about this. I saw your comment on #201 just now.

src/clojure/nrepl/util/completion.clj
```diff
@@ -52,8 +52,9 @@
 (defn static-members
   "Returns a list of potential static members for a given class"
   [^Class class]
-  (for [member (concat (.getMethods class) (.getDeclaredFields class)) :when (static? member)]
-    (.getName ^Member member)))
+  (distinct
+   (for [member (concat (.getMethods class) (.getDeclaredFields class)) :when (static? member)]
```
bbatsov, line 56: Probably it's best to convert this to some `->>` pipeline. Seems the author of the original code was quite fond of `for`, but I'm definitely more fond of `map`, `filter` and friends. :-)

test/clojure/nrepl/util/completion_test.clj
```diff
@@ -45,6 +45,7 @@
            (candidates "java.lang.System/out")))
 
     (is (some #{"String/valueOf"} (candidates "String/")))
+    (is (every? #(= 1 %) (vals (frequencies (candidates "String/")))))
```
bbatsov, line 48: Might be a good idea to add some helper named `unique-candidates?` and use it in the other completion tests as well.

src/clojure/nrepl/util/completion.clj
eerohele: Yeah, I very rarely use `for`, too. I can certainly rewrite the function with `->>`. Would you prefer that I do it in a separate commit?

test/clojure/nrepl/util/completion_test.clj
eerohele: I can do that. I realized that I could just do `(apply distinct? (candidates "String/"))` instead, though.

eerohele:
I pushed the changes you proposed as separate commits. I can squash them prior to the possible merge if necessary.

bbatsov:
Thanks! I'll push a new alpha now.

PR #204, opened 2020-07-06
Support filesystem (UNIX) socket connections

rlbdv:
Add support for UNIX domain sockets (filesystem sockets) via
junixsocket along with a corresponding -s/--socket commandline option,
and depict these connections with an augmented URI scheme,
e.g. nrepl+unix:/some/where%20am%20i

    $ lein run -m nrepl.cmdline -s nreplsock
    nREPL server started on socket /home/rlb/test/nreplsock - nrepl+unix:/home/rlb/test/nreplsock

    $ echo -n 'd4:code7:(+ 2 2)2:op4:evale' | nc -U nreplsock
    d2:n...

When this transport is suitable, it avoids concerns over port
conflicts, and allows access control via the permissions of the parent
directories.

This is unfinished, but I wanted to see if it was likely to be of interest before proceeding further.  (After an obviously long hiatus, I'm finally back to working on nrepl access control.)  I don't see this proposal as an alternative to #46 -- I'm still interested in that too, but if/when they're feasible, filesystem sockets should allow solid (simple) access control without having to resolve all of the questions we were still sorting with respect to handling tokens, etc.

Of course if we do want to pursue this, among other things I'll need to add tests, and I imagine there are any number of questions to answer here too (e.g. the random choice I made about the URI scheme, dependencies, etc.).

The current socket provider is https://github.com/kohlschutter/junixsocket  I'd used it fairly recently for its support for forwarding file descriptors along with the connection: https://kohlschutter.github.io/junixsocket/filedescriptors.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ kondo:
 cloverage:
 	lein with-profile -user,+$(VERSION),+cloverage cloverage --codecov
 
+# Roughly match what runs in CI using the current JVM
+check: test eastwood kondo cljfmt cloverage
+
 verify_cljdoc:
 	curl -fsSL https://raw.githubusercontent.com/cljdoc/cljdoc/master/script/verify-cljdoc-edn | bash -s doc/cljdoc.edn
 
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ cljfmt:
 	lein with-profile -user,+$(VERSION),+cljfmt cljfmt check
 
 kondo:
-	clj-kondo --lint src
+	lein with-profile +clj-kondo run -m clj-kondo.main --lint src
 
 cloverage:
 	lein with-profile -user,+$(VERSION),+cloverage cloverage --codecov
--- a/doc/modules/ROOT/pages/usage/server.adoc
+++ b/doc/modules/ROOT/pages/usage/server.adoc
@@ -53,6 +53,30 @@ how you can easily start a ClojureScript capable nREPL:
 $ clj -R:nREPL -m nrepl.cmdline --middleware "[cider.piggieback/wrap-cljs-repl]"
 ----
 
+By default, nREPL listens for connections on a randomly chosen local
+port with no authentication, but you can specify the address or port,
+or, if you're using JDK 16 or newer, or you add a
+https://kohlschutter.github.io/junixsocket/[junixsocket] dependency,
+you can ask it to listen on a UNIX domain (filesystem) socket instead:
+
+[source,clojure]
+----
+{
+;; ...
+:aliases {:nREPL
+          {:extra-deps
+           ;; UNIX domain socket support was added in 0.9.0
+           {nrepl/nrepl {:mvn/version "0.9.0"}
+            com.kohlschutter.junixsocket/junixsocket-core {:mvn/version "2.3.2"}}}}
+}
+----
+
+[source,shell]
+----
+$ mkdir -m go-rwx nrepl-test
+$ clj -R:nREPL -m nrepl.cmdline --socket nrepl-test/socket
+----
+
 Here's a listing of all the options available via nREPL's command-line
 interface (this output was simply generated with `--help`):
 
@@ -63,6 +87,7 @@ interface (this output was simply generated with `--help`):
 -b/--bind ADDR              Bind address, by default "127.0.0.1".
 -h/--host ADDR              Host address to connect to when using --connect. Defaults to "127.0.0.1".
 -p/--port PORT              Start nREPL on PORT. Defaults to 0 (random port) if not specified.
+-s/--socket PATH            Start nREPL on filesystem socket at PATH.
 --ack ACK-PORT              Acknowledge the port of this server to another nREPL server running on ACK-PORT.
 -n/--handler HANDLER        The nREPL message handler to use for each incoming connection; defaults to the result of `(nrepl.server/default-handler)`.
 -m/--middleware MIDDLEWARE  A sequence of vars, representing middleware you wish to mix in to the nREPL handler.
@@ -186,6 +211,20 @@ WARNING: Keep in mind that running a nREPL server on a public address
 is an epic security hole! As the connections are insecure (no
 authentication, no authorization) by default anyone can connect to
 your app and modify its behaviour or run code on the remote host.
+
+You can also ask nREPL to listen on a UNIX domain (filesystem) socket
+with the `:socket` keyword (if you're using JDK 16 or newer add a
+https://kohlschutter.github.io/junixsocket/[junixsocket] dependency),
+which should be as secure as the access to the socket\'s parent
+directories:
+
+[source,clojure]
+----
+=> (require '[nrepl.server :refer [start-server stop-server]])
+nil
+=> (defonce server (start-server :socket "/some/where/safe/nrepl"))
+='user/server
+----
 
 Depending on what the lifecycle of your application is, whether you want to be
 able to easily restart the server, etc., you might want to put the value
--- a/project.clj
+++ b/project.clj
@@ -11,7 +11,7 @@
   :javac-options ["-target" "8" "-source" "8"]
 
   :aliases {"bump-version" ["change" "version" "leiningen.release/bump-version"]
-            "test-all" ["with-profile" "+1.7:+1.8:+1.9:+1.10:+fastlane" "test"]
+            "test-all" ["with-profile" "+1.7:+1.8:+1.9:+1.10:+fastlane:+junixsocket" "test"]
             "docs" ["with-profile" "+maint" "run" "-m" "nrepl.impl.docs" "--file"
                     ~(clojure.java.io/as-relative-path
                       (clojure.java.io/file "doc" "modules" "ROOT" "pages" "ops.adoc"))]
@@ -40,6 +40,8 @@
                     ;; TODO: replicate koacha's version filter logic here
                     :test-selectors {:default (complement :min-java-version)}
                     :aliases {"test" "test2junit"}}
+             :junixsocket {:jvm-opts ["-Dnrepl.test.junixsocket=true"]
+                           :dependencies [[com.kohlschutter.junixsocket/junixsocket-core "2.3.2"]]}
              ;; Clojure versions matrix
              :provided {:dependencies [[org.clojure/clojure "1.10.2"]]}
              :1.7 {:dependencies [[org.clojure/clojure "1.7.0"]]}
--- a/project.clj
+++ b/project.clj
@@ -41,11 +41,11 @@
                     :test-selectors {:default (complement :min-java-version)}
                     :aliases {"test" "test2junit"}}
              ;; Clojure versions matrix
-             :provided {:dependencies [[org.clojure/clojure "1.10.1"]]}
+             :provided {:dependencies [[org.clojure/clojure "1.10.2"]]}
              :1.7 {:dependencies [[org.clojure/clojure "1.7.0"]]}
              :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}
              :1.9 {:dependencies [[org.clojure/clojure "1.9.0"]]}
-             :1.10 {:dependencies [[org.clojure/clojure "1.10.1"]]
+             :1.10 {:dependencies [[org.clojure/clojure "1.10.2"]]
                     :source-paths ["src/spec"]}
              :master {:repositories [["snapshots"
                                       "https://oss.sonatype.org/content/repositories/snapshots"]]
--- a/project.clj
+++ b/project.clj
@@ -42,6 +42,7 @@
                     :aliases {"test" "test2junit"}}
              :junixsocket {:jvm-opts ["-Dnrepl.test.junixsocket=true"]
                            :dependencies [[com.kohlschutter.junixsocket/junixsocket-core "2.3.2"]]}
+             :clj-kondo {:dependencies [[clj-kondo "2021.06.18"]]}
              ;; Clojure versions matrix
              :provided {:dependencies [[org.clojure/clojure "1.10.2"]]}
              :1.7 {:dependencies [[org.clojure/clojure "1.7.0"]]}
--- a/src/clojure/nrepl/cmdline.clj
+++ b/src/clojure/nrepl/cmdline.clj
@@ -12,8 +12,11 @@
    [nrepl.ack :refer [send-ack]]
    [nrepl.misc :refer [noisy-future]]
    [nrepl.server :as nrepl-server]
+   [nrepl.socket :as socket]
    [nrepl.transport :as transport]
-   [nrepl.version :as version]))
+   [nrepl.version :as version])
+  (:import
+   [java.net URI]))
 
 (defn- clean-up-and-exit
   "Performs any necessary clean up and calls `(System/exit status)`."
@@ -127,6 +130,7 @@ Exit:      Control+D or (exit) or (quit)"
    "-b" "--bind"
    "-h" "--host"
    "-p" "--port"
+   "-s" "--socket"
    "-m" "--middleware"
    "-t" "--transport"
    "-n" "--handler"
@@ -176,6 +180,7 @@ Exit:      Control+D or (exit) or (quit)"
   -b/--bind ADDR              Bind address, by default \"127.0.0.1\".
   -h/--host ADDR              Host address to connect to when using --connect. Defaults to \"127.0.0.1\".
   -p/--port PORT              Start nREPL on PORT. Defaults to 0 (random port) if not specified.
+  -s/--socket PATH            Start nREPL on filesystem socket at PATH.
   --ack ACK-PORT              Acknowledge the port of this server to another nREPL server running on ACK-PORT.
   -n/--handler HANDLER        The nREPL message handler to use for each incoming connection; defaults to the result of `(nrepl.server/default-handler)`.
   -m/--middleware MIDDLEWARE  A sequence of vars, representing middleware you wish to mix in to the nREPL handler.
@@ -336,6 +341,7 @@ Exit:      Control+D or (exit) or (quit)"
   [options]
   {:port (->int (:port options))
    :host (:host options)
+   :socket (:socket options)
    :transport (options->transport options)
    :repl-fn (options->repl-fn options)})
 
@@ -344,10 +350,11 @@ Exit:      Control+D or (exit) or (quit)"
   Returns map of processed options to start an nREPL server."
   [options]
   (let [middleware (sanitize-middleware-option (:middleware options))
-        {:keys [host port transport]} (connection-opts options)]
+        {:keys [host port socket transport]} (connection-opts options)]
     (merge options
            {:host host
             :port port
+            :socket socket
             :transport transport
             :bind (:bind options)
             :middleware middleware
@@ -403,13 +410,14 @@ Exit:      Control+D or (exit) or (quit)"
   Returns connection header string."
   [server options]
   (let [transport (:transport options)
-        port (:port server)
         ^java.net.ServerSocket ssocket (:server-socket server)
-        host (.getHostName (.getInetAddress ssocket))]
+        ^URI uri (socket/as-nrepl-uri ssocket (transport/uri-scheme transport))]
     ;; The format here is important, as some tools (e.g. CIDER) parse the string
     ;; to extract from it the host and the port to connect to
-    (format "nREPL server started on port %d on host %s - %s://%s:%d"
-            port host (transport/uri-scheme transport) host port)))
+    (if-let [host (.getHost uri)]
+      (format "nREPL server started on port %d on host %s - %s"
+              (.getPort uri) host uri)
+      (str "nREPL server started on socket " (.toASCIIString uri)))))
 
 (defn save-port-file
   "Writes a file relative to project classpath with port number so other tools
@@ -427,10 +435,11 @@ Exit:      Control+D or (exit) or (quit)"
   "Creates an nREPL server instance.
   Takes map of CLI options.
   Returns nREPL server map."
-  [{:keys [port bind handler transport greeting]}]
+  [{:keys [port bind socket handler transport greeting]}]
   (nrepl-server/start-server
    :port port
    :bind bind
+   :socket socket
    :handler handler
    :transport-fn transport
    :greeting-fn greeting))
@@ -461,6 +470,12 @@ Exit:      Control+D or (exit) or (quit)"
       (dispatch-commands options))
     (catch clojure.lang.ExceptionInfo ex
       (let [{:keys [:nrepl/kind ::status]} (ex-data ex)]
-        (when (= kind ::exit)
-          (clean-up-and-exit status))
-        (throw ex)))))
+        (case kind
+          ::exit (clean-up-and-exit status)
+          (:nrepl.server/no-filesystem-sockets
+           :nrepl.server/invalid-start-request)
+          (do
+            (binding [*out* *err*]
+              (println (.getMessage ex)))
+            (clean-up-and-exit 2))
+          (throw ex))))))
--- a/src/clojure/nrepl/cmdline.clj
+++ b/src/clojure/nrepl/cmdline.clj
@@ -26,7 +26,8 @@
   "Requests that the process exit with the given `status`.  Does not
   return."
   [status]
-  (throw (ex-info nil {::kind ::exit ::status status})))
+  ;; :nrepl/kind is our shared (ns independent) ExceptionInfo discriminator
+  (throw (ex-info nil {:nrepl/kind ::exit ::status status})))
 
 (defn die
   "`Print`s items in `msg` to *err* and then exits with a status of 2."
@@ -458,7 +459,7 @@ Exit:      Control+D or (exit) or (quit)"
     (let [[options _args] (args->cli-options args)]
       (dispatch-commands options))
     (catch clojure.lang.ExceptionInfo ex
-      (let [{:keys [::kind ::status]} (ex-data ex)]
+      (let [{:keys [:nrepl/kind ::status]} (ex-data ex)]
         (when (= kind ::exit)
           (clean-up-and-exit status))
         (throw ex)))))
--- a/src/clojure/nrepl/cmdline.clj
+++ b/src/clojure/nrepl/cmdline.clj
@@ -10,6 +10,7 @@
    [nrepl.config :as config]
    [nrepl.core :as nrepl]
    [nrepl.ack :refer [send-ack]]
+   [nrepl.misc :refer [noisy-future]]
    [nrepl.server :as nrepl-server]
    [nrepl.transport :as transport]
    [nrepl.version :as version]))
@@ -97,9 +98,9 @@ Exit:      Control+D or (exit) or (quit)"
      (println (repl-intro))
      ;; We take 50ms to listen to any greeting messages, and display the value
      ;; in the `:out` slot.
-     (future (->> (client)
-                  (take-while #(nil? (:id %)))
-                  (run! #(when-let [msg (:out %)] (print msg)))))
+     (noisy-future (->> (client)
+                        (take-while #(nil? (:id %)))
+                        (run! #(when-let [msg (:out %)] (print msg)))))
      (Thread/sleep 50)
      (let [session (nrepl/client-session client)
            ns (atom "user")]
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -13,6 +13,17 @@
       (apply println "ERROR:" msgs)
       (when ex (.printStackTrace ^Throwable ex)))))
 
+(defmacro noisy-future
+  "Executes body in a future, logging any execptions that make it to the
+  top level."
+  [& body]
+  `(future
+     (try
+       ~@body
+       (catch Throwable ex#
+         (log ex#)
+         (throw ex#)))))
+
 (defmacro returning
   "Executes `body`, returning `x`."
   {:style/indent 1}
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -6,9 +6,9 @@
   (:require [clojure.java.io :as io]))
 
 (defn log
-  [ex & msgs]
-  (let [ex (when (instance? Throwable ex) ex)
-        msgs (if ex msgs (cons ex msgs))]
+  [ex-or-msg & msgs]
+  (let [ex (when (instance? Throwable ex-or-msg) ex-or-msg)
+        msgs (if ex msgs (filter identity (cons ex-or-msg msgs)))]
     (binding [*out* *err*]
       (apply println "ERROR:" msgs)
       (when ex (.printStackTrace ^Throwable ex)))))
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -12,9 +12,11 @@
    nrepl.middleware.session
    nrepl.middleware.sideloader
    [nrepl.misc :refer [log noisy-future response-for returning]]
+   [nrepl.socket :as socket :refer [inet-socket unix-server-socket]]
    [nrepl.transport :as t])
   (:import
-   [java.net InetSocketAddress ServerSocket SocketException]))
+   (java.net ServerSocket SocketException)
+   [java.nio.channels ClosedChannelException]))
 
 (defn handle*
   [msg handler transport]
@@ -47,28 +49,32 @@
       (log e "Failed to close " x))))
 
 (defn- accept-connection
-  [{:keys [^ServerSocket server-socket open-transports transport greeting handler]
+  [{:keys [server-socket open-transports transport greeting handler]
     :as server}]
-  (when-not (.isClosed server-socket)
-    (let [sock (.accept server-socket)]
-      (noisy-future (let [transport (transport sock)]
-                      (try
-                        (swap! open-transports conj transport)
-                        (when greeting (greeting transport))
-                        (handle handler transport)
-                        (catch SocketException ex
-                          nil)
-                        (finally
-                          (swap! open-transports disj transport)
-                          (safe-close transport)))))
-      (noisy-future (try
-                      (accept-connection server)
-                      (catch SocketException ex
-                        nil))))))
+  (when-let [sock (try
+                    (socket/accept server-socket)
+                    (catch ClosedChannelException ex
+                      nil))]
+    (noisy-future
+     (let [transport (transport sock)]
+       (try
+         (swap! open-transports conj transport)
+         (when greeting (greeting transport))
+         (handle handler transport)
+         (catch SocketException ex
+           nil)
+         (finally
+           (swap! open-transports disj transport)
+           (safe-close transport)))))
+    (noisy-future
+     (try
+       (accept-connection server)
+       (catch SocketException ex
+         nil)))))
 
 (defn stop-server
   "Stops a server started via `start-server`."
-  [{:keys [open-transports ^ServerSocket server-socket] :as server}]
+  [{:keys [open-transports ^java.io.Closeable server-socket] :as server}]
   (returning server
     (.close server-socket)
     (swap! open-transports
@@ -143,6 +149,7 @@
 
    * :port — defaults to 0, which autoselects an open port
    * :bind — bind address, by default \"127.0.0.1\"
+   * :socket - filesystem socket path (alternative to :port and :bind)
    * :handler — the nREPL message handler to use for each incoming connection;
        defaults to the result of `(default-handler)`
    * :transport-fn — a function that, given a java.net.Socket corresponding
@@ -161,20 +168,17 @@
    either via `stop-server`, (.close server), or automatically via `with-open`.
    The port that the server is open on is available in the :port slot of the
    server map (useful if the :port option is 0 or was left unspecified."
-  ^nrepl.server.Server
-  [& {:keys [port bind transport-fn handler ack-port greeting-fn]}]
-  (let [port (or port 0)
-        addr (fn [^String bind ^Integer port] (InetSocketAddress. bind port))
-        transport-fn (or transport-fn t/bencode)
-        ;; We fallback to 127.0.0.1 instead of to localhost to avoid
-        ;; a dependency on the order of ipv4 and ipv6 records for
-        ;; localhost in /etc/hosts
-        bind (or bind "127.0.0.1")
-        ss (doto (ServerSocket.)
-             (.setReuseAddress true)
-             (.bind (addr bind port)))
+  [& {:keys [port bind socket transport-fn handler ack-port greeting-fn]}]
+  (when (and socket (or port bind))
+    (let [msg "Cannot listen on both port and filesystem socket"]
+      (log msg)
+      (throw (ex-info msg {:nrepl/kind ::invalid-start-request}))))
+  (let [transport-fn (or transport-fn t/bencode)
+        ss (if socket
+             (unix-server-socket socket)
+             (inet-socket bind port))
         server (Server. ss
-                        (.getLocalPort ss)
+                        (when-not socket (.getLocalPort ^ServerSocket ss))
                         (atom #{})
                         transport-fn
                         greeting-fn
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -11,10 +11,10 @@
    nrepl.middleware.lookup
    nrepl.middleware.session
    nrepl.middleware.sideloader
-   [nrepl.misc :refer [log response-for returning]]
+   [nrepl.misc :refer [log noisy-future response-for returning]]
    [nrepl.transport :as t])
   (:import
-   [java.net InetSocketAddress ServerSocket]))
+   [java.net InetSocketAddress ServerSocket SocketException]))
 
 (defn handle*
   [msg handler transport]
@@ -36,7 +36,7 @@
    Returns nil when [recv] returns nil for the given transport."
   [handler transport]
   (when-let [msg (normalize-msg (t/recv transport))]
-    (future (handle* msg handler transport))
+    (noisy-future (handle* msg handler transport))
     (recur handler transport)))
 
 (defn- safe-close
@@ -51,15 +51,20 @@
     :as server}]
   (when-not (.isClosed server-socket)
     (let [sock (.accept server-socket)]
-      (future (let [transport (transport sock)]
-                (try
-                  (swap! open-transports conj transport)
-                  (when greeting (greeting transport))
-                  (handle handler transport)
-                  (finally
-                    (swap! open-transports disj transport)
-                    (safe-close transport)))))
-      (future (accept-connection server)))))
+      (noisy-future (let [transport (transport sock)]
+                      (try
+                        (swap! open-transports conj transport)
+                        (when greeting (greeting transport))
+                        (handle handler transport)
+                        (catch SocketException ex
+                          nil)
+                        (finally
+                          (swap! open-transports disj transport)
+                          (safe-close transport)))))
+      (noisy-future (try
+                      (accept-connection server)
+                      (catch SocketException ex
+                        nil))))))
 
 (defn stop-server
   "Stops a server started via `start-server`."
@@ -174,7 +179,11 @@
                         transport-fn
                         greeting-fn
                         (or handler (default-handler)))]
-    (future (accept-connection server))
+    (noisy-future
+     (try
+       (accept-connection server)
+       (catch SocketException ex
+         nil)))
     (when ack-port
       (ack/send-ack (:port server) ack-port transport-fn))
     server))
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -53,7 +53,7 @@
     :as server}]
   (when-let [sock (try
                     (socket/accept server-socket)
-                    (catch ClosedChannelException ex
+                    (catch ClosedChannelException _
                       nil))]
     (noisy-future
      (let [transport (transport sock)]
@@ -61,7 +61,7 @@
          (swap! open-transports conj transport)
          (when greeting (greeting transport))
          (handle handler transport)
-         (catch SocketException ex
+         (catch SocketException _
            nil)
          (finally
            (swap! open-transports disj transport)
@@ -69,7 +69,7 @@
     (noisy-future
      (try
        (accept-connection server)
-       (catch SocketException ex
+       (catch SocketException _
          nil)))))
 
 (defn stop-server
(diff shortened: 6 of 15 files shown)


bbatsov:
I'm okay with the suggestion in general, but keep in mind that nREPL can't have any runtime deps, as any of them is a potential conflict with the user deps. If it's hard to provide this transport without that library we'll have to move this to an extension, similar to Drawbridge and Fastlane.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> I'm okay with the suggestion in general, but keep in mind that nREPL
> can't have any runtime deps, as any of them is a potential conflict
> with the user deps. If it's hard to provide this transport without
> that library we'll have to move this to an extension, similar to
> Drawbridge and Fastlane.

Ahh, OK.  I was also quite surprised to just notice:

  https://bugs.openjdk.java.net/browse/JDK-8238588

In any case, from your perspective, would you prefer I pursue this
approach, or resume work on the token based authorization?  I suppose we
might end up with both, but perhaps best to try to finish the one we
favor first.

Ideally, of course, I'd like to get to a point where we could consider
guarding access by default, and perhaps that's more likely to be
feasible sooner with the token based auth?

rlbdv:
Hmm, though I suppose another way we could look at it is that if the JVM really is going to add unix sockets soon, then adding them to nrepl as an extension first (as long is it's sufficiently easy to enable) might provide a reasonable path, i.e. begin with an extension, include the support directly in nrepl once the JVM allows it, and eventually, default to filesystem sockets.

bbatsov:
Yeah, exactly. I'd be happy to host such an extension under the umbrella of the official nREPL GitHub org.

rlbdv:
Bozhidar Batsov <notifications@github.com> writes:

> Yeah, exactly. I'd be happy to host such an extension under the umbrella of the official nREPL GitHub org. 

OK, I'll need to see what that would/should look like.  Haven't yet
used drawbridge, etc. -- Thanks

rlbdv:
Do you have some obvious way you think that might work best?  i.e. let's say we have a transitional extension, is there a reasonably easy way we could allow people to enable it for themselves (or at least make it available) more globally?

bbatsov:
> i.e. let's say we have a transitional extension, is there a reasonably easy way we could allow people to enable it for themselves (or at least make it available) more globally?

It's going to be a regular dep for their projects, that they can easily add in their built tool profiles if they want to enable it globally. I think that's pretty straightforward.

rlbdv:
I realized that we might be able to make the dependency completely optional (while we wait to see what happens with the JEP).  I don't know that we'd want to pursue that approach, but I went ahead and pushed an update to demonstrate it.  I suppose It'd avoid having to maintain another project, but the implementation (at least the way I have it now) is a bit awkward.  As the commit mentions, you can try it by enabling and disabling the dep in project.clj.  I also added a couple of other minor commits you might be interested in.

bbatsov:
I'm open to this approach, we just have to come up with the most straightforward way to handle this. I'll try to find some time soon to take a more careful look at the implementation. 

Btw, even if the JEP gets approved, we'd still have to support JDKs without it for a very long time, so that's something we have to keep in mind.

rlbdv:
Oh, certainly, and I'm fine either way, i.e. were you to eventually
decide you'd rather pursue the separate project.

(I also suspect there may be a better way to do what I did in this
 iteration.)

bbatsov:
Sorry for the long radio silence. I completely forgot about this PR for a while. After thinking about it a bit more I think the approach with an optional dependency is probably the best one in terms of usability/maintainability (and an eventual transition to native Java support for UNIX sockets).

project.clj
```diff
@@ -29,6 +29,9 @@
                                     :password :env/clojars_password
                                     :sign-releases false}]]
 
+  ;; Add/remove this line to test behavior
+  ;;:dependencies [[com.kohlschutter.junixsocket/junixsocket-core "2.3.2"]]
```
bbatsov, line 33: For test purposes you can add this to some of the profiles - e..g the `test` one or a dedicated `unix-sockets` profile.

bbatsov:
Btw, I've noticed there are several UNIX socket libraries for Java. Any reason why you picked this one in particular?

src/clojure/nrepl/cmdline.clj
```diff
@@ -26,7 +28,7 @@
   "Requests that the process exit with the given `status`.  Does not
   return."
   [status]
-  (throw (ex-info nil {::kind ::exit ::status status})))
+  (throw (ex-info nil {:nrepl/kind ::exit ::status status})))
```
bbatsov, line 31: I think you should add some comment here explaining why `:kind` is different.

src/clojure/nrepl/cmdline.clj
```diff
@@ -401,13 +407,21 @@ Exit:      Control+D or (exit) or (quit)"
...
-        ^java.net.ServerSocket ssocket (:server-socket server)
-        host (.getHostName (.getInetAddress ssocket))]
+        ^java.net.ServerSocket ssocket (:server-socket server)]
     ;; The format here is important, as some tools (e.g. CIDER) parse the string
     ;; to extract from it the host and the port to connect to
-    (format "nREPL server started on port %d on host %s - %s://%s:%d"
-            port host (transport/uri-scheme transport) host port)))
+    (let [unix-socket (nrepl.server/unix-server-socket-class)]
+      (if (some-> unix-socket (instance? ssocket))
+        ;; Q: str or .toASCIIString?
```
bbatsov, line 415: First time I hear of `.toASCIIString`. What would be the benefit of using it?

src/clojure/nrepl/server.clj
```diff
@@ -133,11 +135,58 @@
   java.io.Closeable
   (close [this] (stop-server this)))
 
+(defn inet-socket [bind port]
```
bbatsov, line 138: I'd deliniate with some section comments the boundaries of the UNIX socket part of the implementation.

bbatsov:
I've left a bit of small feedback here and there. Overall the PR seems to be in a good shape to me and with a some polish and documentation it can make the next nREPL release.

rlbdv:
Radio silence for a bit on my part this time, but I'm back, and will see about incorporating your comments soon.  Thanks.

rlbdv:
> Btw, I've noticed there are several UNIX socket libraries for Java. Any reason why you picked this one in particular?

If I recall correctly, it was in part because I'd used it elsewhere, and so was familiar with it.  In that case I'd picked it because it had the ability to forward file descriptors.  But happy to switch to something else if you have a preference.

rlbdv:
Hah, and forgot -- if it's useful there's a [comparison matrix](https://kohlschutter.github.io/junixsocket/).

src/clojure/nrepl/cmdline.clj
rlbdv: I'm not sure it should be, but I think I was imagining that we might not want a `kind` per-namespace.  If the clojure world had a solid expectation that `:kind` was always a type marker in `ex-data`, then I might just use that (and we could decide to anyway), but `:nrepl/kind` avoids the question while still providing a general (not ns-specific) marker.

Put another way, while the value might well deserve to be fully namespaced, having a shared/common `kind` allows unified dispatch handling in `ExceptionInfo` catch clauses, perhaps via `case`, etc.

rlbdv: Oh, hah, maybe you just meant that I should include the commit message comment (similar in spirit to https://github.com/nrepl/nrepl/pull/204#discussion_r559236620) in the code...

rlbdv: Perhaps something like this, unless there's a better place for it:
```
(defn exit
  "Requests that the process exit with the given `status`.  Does not
  return."
  [status]
  ;; :nrepl/kind is our shared (ns independent) ExceptionInfo discriminator
  (throw (ex-info nil {:nrepl/kind ::exit ::status status})))
```

src/clojure/nrepl/cmdline.clj
rlbdv: I don't recall for sure yet, but I think I was wondering what junixsocket expects/allows here because I believe if you have a URI with non-ASCII characters, then as long as they're not in the `other` category, `str` (i.e. `toString`) will just pass them through (not RFC compliant), while `.toASCIIString` will octet escape them (compliant).  So if junixsocket can handle the non-compliant chars, then great, but if not, we might need the encoding?

rlbdv: See also https://docs.oracle.com/javase/8/docs/api/java/net/URI.html (particularly "These operations are exposed in the constructors and methods of this class as follows: ...").

rlbdv:
Oh, and if I read it right, also looks like the JEP has been completed and should be in JDK 16.

bbatsov:
> Oh, and if I read it right, also looks like the JEP has been completed and should be in JDK 16.

Hmm, that's a convenient development. Perhaps it'd be best to just rely on the JDK 16 implementation and make JDK 16 a requirement for using UNIX sockets?

rlbdv:
If it's plausible, I think I'd prefer to support both 16 and some other fallback like junixsocket (assuming we can do that without too much trouble), since I suspect it may be a good while before everyone can count on 16, and I'd like them to have a secure option in the interim.

bbatsov:
Okay, that's fair. Let's work in this direction then.

rlbdv:
OK, I think I've accounted for your previous suggestions, and proposed some documentation, but if I've overlooked anything, please let me know.  (Though I haven't tested the `clj` example in the docs.)

bbatsov:
@rlbdv I'll take a closer look tomorrow, but overall the PR seems to be in a good shape. You'll have to replace the usages of `ex-message` as we need to support Clojure 1.7+.

src/clojure/nrepl/server.clj
```diff
@@ -133,11 +135,65 @@
...
+      (.bind (addr bind port)))))
+
+
+;;; Unix domain (file namespace) sockets
+
+(def unix-socket-address-class
+  (memoize
+   (fn []
+     (try
+       (clojure.core/import* "org.newsclub.net.unix.AFUNIXSocketAddress")
```
bbatsov, line 159: Doesn't the regular `import` function work here?

src/clojure/nrepl/server.clj
```diff
@@ -133,11 +135,65 @@
...
+       (catch ClassNotFoundException ex
+         nil)))))
+
+(def unix-server-socket-class
```
bbatsov, line 163: I seem that's almost the same as the previous function - I think creating a simple helper wouldn't hurt here.

doc/modules/ROOT/pages/usage/server.adoc
```diff
@@ -63,6 +91,7 @@ interface (this output was simply generated with `--help`):
 -b/--bind ADDR              Bind address, by default "127.0.0.1".
 -h/--host ADDR              Host address to connect to when using --connect. Defaults to "127.0.0.1".
 -p/--port PORT              Start nREPL on PORT. Defaults to 0 (random port) if not specified.
+-s/--socket PATH            Start nREPL on filesystem socket at PATH.
```
bbatsov, line 94: I wonder if `-u/--unix-socket` won't be better (clearer) names here.

doc/modules/ROOT/pages/usage/server.adoc
```diff
@@ -187,6 +216,18 @@ is an epic security hole! As the connections are insecure (no
 authentication, no authorization) by default anyone can connect to
 your app and modify its behaviour or run code on the remote host.
 
+You can also ask nREPL to listen on a UNIX domain (filesystem) socket
```
bbatsov, line 219: Probably we should also add a note that this won't work on Windows. :-)

src/clojure/nrepl/cmdline.clj
bbatsov: Yeah, that's what I had in mind.

src/clojure/nrepl/cmdline.clj
bbatsov: Fair enough. I'd add some comment expanding on this then, so at least your thoughts on the subject don't get lost.

bbatsov:
I've also noticed a small error discovered by the linter:

```
src/clojure/nrepl/server.clj:171:7: misplaced-docstrings: Possibly misplaced docstring, fs-socket
src/clojure/nrepl/server.clj:171:1: unused-ret-vals: Constant value is discarded: "Returns a filesystem socket bound to the path if the dependencies (which are\n  loaded dynamically) are available.  If\n  com.kohlschutter.junixsocket/junixsocket-core is not available, throws the\n  ex-info map {:nrepl/kind ::no-filesystem-sockets}."
```

src/clojure/nrepl/server.clj
rlbdv: Yep, not sure why I thought we needed that, but `import` appears to work fine.

src/clojure/nrepl/server.clj
rlbdv: Added as `class-finder`.

doc/modules/ROOT/pages/usage/server.adoc
rlbdv: Just fine with me if you prefer.  Would you also like me to change `:socket` to `:unix-socket`?

Oh, and haven't thought about it carefully yet, but remembered that as another approach, some tools like `psql` (postgresql cli) just treat any `--host` that starts with a slash as a filesystem socket path.

doc/modules/ROOT/pages/usage/server.adoc
rlbdv: Hmm, no idea - though junixsocket claims to provide some support.  Suppose someone with a system handy could try it there: https://github.com/kohlschutter/junixsocket#junixsocket

rlbdv:
...I do that all the time.  Thanks.

rlbdv:
OK, I think I've addressed everything you mentioned except the Windows-related documentation changes, and the option names.  Happy to fix those up too, once we decide what we want there.

src/clojure/nrepl/cmdline.clj
rlbdv: I imagine what we really want here is test coverage, i.e. do we return a URI that can make a round trip for "other" characters?  If I get a chance, I might see if I can come up with some relevant tests, and please let me know if you think that's critical for this first pass.  I suppose in the end, what we'll *really* care about is what works correctly with the new JDK functions.  Offhand, I'd guess toASCIIString, but don't know yet.

rlbdv:
Oh, and regarding the other eastwood complaint, don't know if you'd want to, but you could suppress the `unused-ret-vals` complaint like this:
```
(defn- ignore-value [x] x)

(defn- check-response-format
  "checks response against spec, if available it to do a spec check later"
  [resp]
  (ignore-value
   (when-require 'nrepl.spec
     (when-not (#'clojure.spec.alpha/valid? :nrepl.spec/message resp)
       (throw (Exception. (#'clojure.spec.alpha/explain-str :nrepl.spec/message resp))))))
  resp)
```

bbatsov:
Seems now the CI is failing because of an extra blank line.

src/clojure/nrepl/cmdline.clj
bbatsov: I'm with the current state of the code.

bbatsov:
Overall I'm happy with the current state of the code and I'm okay to merge it when the CI gets green.

rlbdv:
Sounds good -- I'll see about fixing that soon, and did you decide you wanted to keep the existing option names, i.e. `-s/--socket`, or should I change them to `-u/--unix-socket` as you had been pondering?

rlbdv:
> Seems now the CI is failing because of an extra blank line.

Oh, and which line?

bbatsov:
image

doc/modules/ROOT/pages/usage/server.adoc
bbatsov: @rlbdv Hmm, the idea to recycle `--host` is not bad either. Without giving it much thought - it definitely makes sense to me and saves us from having to come up with a name for this CLI option. :D

src/clojure/nrepl/server.clj
```diff
@@ -133,11 +135,64 @@
...
+      (doto (new-instance)
+        (.bind addr)))))
+
+
 (defn start-server
   "Starts a socket-based nREPL server.  Configuration options include:
 
    * :port — defaults to 0, which autoselects an open port
    * :bind — bind address, by default \"127.0.0.1\"
+   * :socket - filesystem socket path (alternative to :port and :bind)
```
bbatsov, line 195: I'm guessing that here we can recycle `:bind` as well and just check if it's a path. Too many decisions. 😆

rlbdv: We could, if you prefer -- be happy to rework all this to just rely on some kind of "/foo" detection if you like, and regarding the extra lines, I suspect one of those is a mistake, but otherwise, I'd used two blank lines to separate the unix domain stuff.  Happy to remove that, though.

rlbdv: Oh, and I may nearly have JDK 16 sockets working too -- or rather, I think I do have them working (though we may have to go back to `import*` (not sure yet)), but I'm having a little trouble getting `lein` to handle `lein test` for (opportunistic) JDK socket testing) and `lein with-profiles test-junixsocket test` for junixsocket.  The profile merge is causing some unexpected errors.  I'll get back to you once I investigate further, and of course, we can always add that later, once I sort it out.

rlbdv: Got it, I think.  Should have an update in the next day or two.

bbatsov:
Btw, JDK 16 was released today! Very convenient timing. 😆

src/clojure/nrepl/server.clj
bbatsov: > We could, if you prefer -- be happy to rework all this to just rely on some kind of "/foo" detection if you like, and regarding the extra lines, I suspect one of those is a mistake, but otherwise, I'd used two blank lines to separate the unix domain stuff. Happy to remove that, though.

Btw, should we always pass the socket explicitly? I've been thinking that for the server it also makes some sense to just create a socket file automatically if one is not specified. I seem to recall some services operated in this manner. Perhaps I'm obsessing over the details too much, but given how hard it is to change something in nREPL once it's released it's better to be safe than sorry. :D

rlbdv:
Yep, that's what prompted me to check, and I noticed it was already in Debian, so I thought I'd give it a try.

  By the way, I'm not sure we'll need it, but if it were to simplify the code, might it be acceptable to make the `import` probing unconditional, rather than memoized (deferred)?

rlbdv:
(And of course, I mean with respect to the two `-class` definitions.)

rlbdv:
I may have spoken too soon (it wasn't actually working as well as I'd thought).  I'll take a day or two to see if I can get the JDK sockets working too, but if not, I'll just finish up the current work and we can handle the JDK sockets separately.

If that ends up being the case, I think the only remaining questions might be whether you'd like me to change `--socket` etc., and whether you'd prefer to just detect a filesystem socket via a leading slash character.

bbatsov:
> I may have spoken too soon (it wasn't actually working as well as I'd thought). I'll take a day or two to see if I can get the JDK sockets working too, but if not, I'll just finish up the current work and we can handle the JDK sockets separately.

No rush. nREPL releases don't happen very often, so it'd be nice to cover this in 0.9.

> If that ends up being the case, I think the only remaining questions might be whether you'd like me to change --socket etc., and whether you'd prefer to just detect a filesystem socket via a leading slash character.

I'm still on the fence about this, but it seems that if we go down this route we won't have to do additional changes in Lein and Boot, which is good. The only downside that I see is that `/` inference would mean people always have to specify absolute paths, which may be a bit annoying.

rlbdv:
I got closer, but it looks like we might have to rework server.clj  a bit more to get it working, i.e. I believe JDK fs sockets may not support the "legacy" Socket and ServerSocket APIs (and calling `(.socket sock)` after the bind does throw `UnsupportedOperationException`.  Under [Limitations](https://inside.java/2021/02/03/jep380-unix-domain-sockets-channels):
> Support for java.net.Socket and java.net.ServerSocket. The legacy socket networking classes are tied to TCP/IP mainly due to their fixed use of java.net.InetAddress for addressing.

I might pursue that a bit more, but I suspect I'll end up just trying to finish up this pr for now and we can defer JDK socket support until later.

rlbdv:
OK, so I believe I got JDK 16 support working too, which resulted in some additional  complexity (see the commit messages).  Happy to adjust if you like, and here's what I think might be remaining:
  - Deciding if we'd prefer to drop `-s` in preference to `-b ABSOLUTE_PATH`.  After realizing that would preclude relative paths, I think I might prefer the separate option.
  - Deciding if we want to keep the `nrepl+unix:PATH` URI syntax, or just always render an absolute PATH, so you can detect a unix domain path via the first character in `nrepl:PATH`.
  - Maybe adding some tests that use unix domain paths.  If nothing else, I'd thought about adding a `:slow` `-s` test, though it'd be better to test unix domain sockets during a normal `lein test` -- haven't looked to see how easy that might be yet.

More generally, I wanted to see how you felt about the changes I had to make to support JDK 16+ -- and more than happy to rework the commits, and/or remove or raise some of them separately if you'd prefer that (e.g. not even sure if CI supports openjdk16 yet).

bbatsov:
> Deciding if we'd prefer to drop -s in preference to -b ABSOLUTE_PATH. After realizing that would preclude relative paths, I think I might prefer the separate option.

Fine by me.

> Deciding if we want to keep the nrepl+unix:PATH URI syntax, or just always render an absolute PATH, so you can detect a unix domain path via the first character in nrepl:PATH.

I guess it's fine to omit `+unix` as the protocol remains mostly the same (beconde), regardless of the change of the transport. I don't remember very well the client inference logic (the client code was deciding which transport to use based on this protocol prefix or something along those lines), but that's probably not a big deal. If you can get it to work with the built-in client then we're in business. 

> Maybe adding some tests that use unix domain paths. If nothing else, I'd thought about adding a :slow -s test, though it'd be better to test unix domain sockets during a normal lein test -- haven't looked to see how easy that might be yet.

Yeah, that'd be nice.

> More generally, I wanted to see how you felt about the changes I had to make to support JDK 16+ -- and more than happy to rework the commits, and/or remove or raise some of them separately if you'd prefer that (e.g. not even sure if CI supports openjdk16 yet).

It seems that the CI supports JDK 16, so I added it to the matrix.

src/clojure/nrepl/misc.clj
```diff
@@ -6,13 +6,24 @@
...
-  (let [ex (when (instance? Throwable ex) ex)
-        msgs (if ex msgs (cons ex msgs))]
+  [ex-or-msg & msgs]
+  (let [ex (when (instance? Throwable ex-or-msg) ex-or-msg)
+        msgs (if ex msgs (filter identity (cons ex-or-msg msgs)))]
     (binding [*out* *err*]
       (apply println "ERROR:" msgs)
       (when ex (.printStackTrace ^Throwable ex)))))
 
+(defmacro noisy-future
```
bbatsov, line 16: Not sure that's the best name for this function. :D

bbatsov:
As for the changes for JDK 16 - superficially they seemed okay, but I'll have to review them more carefully.

src/clojure/nrepl/server.clj
rlbdv: Could you elaborate a bit - not sure I understand what you're considering here yet.

rlbdv: And no worries at all about the details, I completely understand for something like this, and am happy to try to help settle on something we think might be durable.

rlbdv:
> > If that ends up being the case, I think the only remaining questions might be whether you'd like me to change --socket etc., and whether you'd prefer to just detect a filesystem socket via a leading slash character.
> 
> I'm still on the fence about this, but it seems that if we go down this route we won't have to do additional changes in Lein and Boot, which is good. The only downside that I see is that `/` inference would mean people always have to specify absolute paths, which may be a bit annoying.

That's a good point, and I'm not sure I understand the broader contexts well yet (including the clients).  What's generally the source of the connection information?  If nrepl itself is generally the source of the information (a URI) and the consumer, then `nrepl+unix:PATH` might work fine (and path can be relative or absolute) if there were a way to provide it on the command line and to the server.  (And then options like `-s` or `-b` would just be alternate conveniences.)

src/clojure/nrepl/misc.clj
rlbdv: Heh, happy to change it if you like - any preferred name.  And not positive you'd want to use it as pervasively as the commit does.  I just added it after being thwarted by some hidden exceptions when I was working on the sockets.

rlbdv:
OK, I've added an initial cmdline test.

rlbdv:
Rebased.

rlbdv:
Rebased, and while I fixed up some of the eastwood reflection warnings, I wondered how you might want to handle the rest (since we're dynamically looking up the socket classes, and they don't share a common parent class).  One option might be some suitable easwood suppressions, and/or isolating that code even further, namespace-wise, but I don't yet know how feasible that'll be and I wanted to see what your inclinations were before spending more time on it.

bbatsov:
I think it's fine to simply suppress the things we don't really need to fix.

rlbdv:
I see now that it's not eastwood, it's just a `*warn-on-reflection* true` in `:global-vars`, so I don't know of any great "selective" approaches.  Although I suppose something a bit crude like this would work:
```clj
....
(set! *warn-on-reflection* false)

(defn get-local-port [sock]
  (.getLocalPort sock))

(set! *warn-on-reflection* true)
...
```

rlbdv:
OK, not sure if it's exactly how you'd like to handle it, but I added a new commit that either fixes or suppresses all of the reflection warnings that had been introduced by these changes.

bbatsov:
Thanks. I'm with a computer for a few days (vacation) and I'll check this next week. @vemv Might also have ideas about the optimal way to handle those reflection warnings.

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,222 @@
...
+   (java.nio ByteBuffer)
+   (java.nio.file Path)
+   (java.nio.channels Channels ClosedChannelException
+                      ServerSocketChannel SocketChannel)))
+
+(def orig-warn-on-reflection *warn-on-reflection*)
+
+(defmacro find-class [full-path]
+  `(try
+     (import ~full-path)
```
vemv, line 23: I think `Class/forName` would be a better / more standard choice here.

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,222 @@
...
+     (catch ClassNotFoundException ex#
+       nil)))
+
+;;; InetSockets (TCP)
+
+(defn inet-socket [bind port]
+  (let [port (or port 0)
+        addr (fn [^String bind ^Integer port] (InetSocketAddress. bind port))
```
vemv, line 31: The arg is int not Integer https://docs.oracle.com/javase/7/docs/api/java/net/InetSocketAddress.html

In any case, I wouldn't use hints here but casting, it's what's recommended for primitives:

image

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,222 @@
...
+  (find-class 'java.net.UnixDomainSocketAddress))
+
+(def ^Class jdk-unix-server-class
+  (find-class 'java.nio.channels.ServerSocketChannel))
+
+(def ^:private test-junixsocket?
+  ;; Make it possible to test junixsocket even when JDK >= 16
+  (= "true" (System/getProperty "nrepl.test.junixsocket")))
+
+(def ^:const unix-domain-flavor
```
vemv, line 58: :const is not a feature that is officially described in https://clojure.org/ so I'd just like it a vanilla def.

Savvy users can use direct linking which is cleaner and more generalized

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,222 @@
...
+    (and jdk-unix-address-class jdk-unix-server-class) :jdk
+    (and junixsocket-address-class junixsocket-server-class) :junixsocket
+    :else nil))
+
+(defn unix-socket-address
+  "Returns a filesystem socket address for the given path string."
+  [^String path]
+  (case unix-domain-flavor
+    :jdk
+    (Reflector/invokeStaticMethod jdk-unix-address-class "of"
```
vemv, line 75: Not sure if lang.Reflector is a public API. Did you consider using java's vanilla reflection API?

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,222 @@
...
+    (Reflector/invokeConstructor junixsocket-address-class
+                                 (into-array File [(File. path)]))
+    (let [msg "Support for filesystem sockets requires JDK 16+ or a junixsocket dependency"]
+      (log msg)
+      (throw (ex-info msg {:nrepl/kind ::no-filesystem-sockets})))))
+
+(set! *warn-on-reflection* false)
+
+(defn- bind [sock addr] (.bind sock addr))
+(defn- get-path [addr] (.getPath addr))
```
vemv, line 87: Could you summarize why reflection is used here? What are the classes that can go down this path? Do they implement a common interface or not?

src/clojure/nrepl/transport.clj
```diff
@@ -73,8 +81,16 @@
        (map (fn [[k v]] [k (<bytes v)]))
        (into {})))
 
+(set! *warn-on-reflection* false)
+
+(defn- connected? [s] (.isConnected s))
+(defn- flush-output [s] (.flush s))
+(defn- close-stream [s] (.close s))
```
vemv, line 88: Why the reflection?

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,222 @@
...
+                nil  ;; userInfo
+                (-> sock .getInetAddress .getHostName)
+                (.getLocalPort sock)
+                nil  ;; path
+                nil  ;; query
+                nil))))))  ;; fragment
+
+(set! *warn-on-reflection* orig-warn-on-reflection)
+
+(defprotocol Acceptable
```
vemv, line 148: Couldn't this very same technique be used to avoid reflection? defprotocol + extend-protocol for all relevant classes

vemv (commented):
Hey there!

Some quick feedback and ideas.

Great work btw!

src/clojure/nrepl/socket.clj
rlbdv: Thanks, fix pending.

src/clojure/nrepl/socket.clj
rlbdv: Thanks, fix pending.

src/clojure/nrepl/socket.clj
rlbdv: If it matters, I didn't add this for performance, I added it as a signal that other code may (and I think right now does) assume that the value will never change, but happy to remove it if it's undesirable, or ineffective.

src/clojure/nrepl/socket.clj
rlbdv: I have now :)
(Fix pending.)

src/clojure/nrepl/socket.clj
rlbdv: Looks like we can drop `bind`, but I think we'll want to keep `get-path` (or something equivalent) since that method doesn't exist for `ServerSocket` before JDK 16.

rlbdv: (...and of course AFUnixDomainSocket doesn't exist unless someone's added junixsocket it to their dependencies, and it's getPath returns String rather than Path.)

src/clojure/nrepl/transport.clj
rlbdv: Good question, and it turns out to be unnecessary.

src/clojure/nrepl/socket.clj
rlbdv: In some cases, but of course not where we don't know a suitable class will be available, either because earlier versions of the JVM didn't have it (e.g. (.getPath SocketAddress) before JVM 15), or because the relevant classes are optional and there's no useful parent (across all the supported JVM versions).

Prompted by your review, I was able to eliminate all of the `*warn-on-reflection*` manipulations except this one.  I'll push an update within the next few days.

Thanks much for the help.

src/clojure/nrepl/socket.clj
vemv: I'd say that Clojure programmers have the general assumption that vars never change - otherwise it's signaled by the use of a reference type (atoms, etc)

So I'd vote on removing it, FWIW :)

vemv (commented):
Kudos for the iteration, looking forward to see it!

rlbdv:
Rebased and updated to hopefully reflect all the outstanding review comments.

Thanks

vemv:
Hi again,

please make sure the build is green, then I'll gladly have a look

While you're there, please upgrade Eastwood (it's simply a version bump - nothing else needed). Now it bundles the https://github.com/jonase/eastwood#reflection linter so we'd know in advance whether something is off.

Thanks!

rlbdv:
Hmm, the tests pass when I run them here (assuming I'm running them correctly).  I can't see what's failing on CircleCI since I haven't given them access to my account, and it doesn't look like they make the results visible otherwise, even for public projects?

Locally I've run lein check, lein test with +eastwood, and lein test with all the +X.Y versions listed in the project file in combinations with and without +junixsocket and with both jdk 1.8 and jdk 16.

rlbdv:
Ahh, I see I probably wasn't running everything, i.e. I had overlooked `make check` as the top level.  I'll look in to it.

vemv:
>  and it doesn't look like they make the results visible otherwise, even for public projects?

pretty sure they do, although I'd understand if you don't want to grant them access (GH's API is a mess! For this reason I have my repos under a dedicated org - this way I can opt not to grant permissions for the given org)

rlbdv:
For what it's worth, when I click on those "details" links on the failures, it takes me to a "log in, and we'll need read/write access to your repos" page.  But I think I know what's going wrong.  I think it's probably some cljfmt complaints where cljfmt doesn't agree with clojure-mode.

rlbdv:
i.e. what I have vs what cljfmt appears to want:
```clj
--- a/src/clojure/nrepl/socket.clj
+++ b/src/clojure/nrepl/socket.clj
@@ -187,12 +187,12 @@

 (definterface Writable
   (write [byte-array]
-    "Writes the given bytes to the output as per OutputStream write.")
+         "Writes the given bytes to the output as per OutputStream write.")
   (write [byte-array offset length]
-    "Writes the given bytes to the output as per OutputStream write."))
+         "Writes the given bytes to the output as per OutputStream write."))

 (defrecord BufferedOutputChannel
-    [^SocketChannel channel ^ByteBuffer buffer]
+           [^SocketChannel channel ^ByteBuffer buffer]

   java.io.Flushable
   (flush [this]
```

rlbdv:
Well, looks like that got a bit further, but no idea what's wrong with jdk 8, 1.10; that works fine here.

vemv:
Here's one example, hope it helps

image

vemv:
clj-kondo is failing as well

rlbdv:
I have fixes for all the kondo complaints except these.  Is there some other way it wants me to define the `Writable` interface?
```clj
$ lein with-profiles +kondo run -m clj-kondo.main --lint src
OpenJDK 64-Bit Server VM warning: Options -Xverify:none and -noverify were deprecated in JDK 13 and will likely be
removed in a future release.
src/clojure/nrepl/socket.clj:191:11: warning: unused binding byte-array
src/clojure/nrepl/socket.clj:191:22: warning: unused binding offset
src/clojure/nrepl/socket.clj:191:29: warning: unused binding length
src/clojure/nrepl/socket.clj:198:11: warning: unused binding this
linting took 781ms, errors: 0, warnings: 4
```

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,238 @@
...
+    (.accept s)))
+
+;; We have to handle this ourselves for NIO because unfortunately read and write
+;; hang if we use both Channels/newInputStream and Channels/newOutputStream.
+;; Read and write deadlock on a shared channel input/output stream lock
+;; (cf. https://bugs.openjdk.java.net/browse/JDK-4509080).  Verified that this
+;; still happens (via thread dump when hung) with jdk 16.
+
+(definterface Writable
+  (write [byte-array]
```
vemv, line 189:
```suggestion
  (write [_byte-array]
```

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,238 @@
...
+         "Writes the given bytes to the output as per OutputStream write.")
+  (write [byte-array offset length]
```
vemv, line 191:
```suggestion
  (write [_byte-array _offset _length]
```

vemv (commented):
Suggestions for addressing clj-kondo

(I don't think these should be linter faults btw... will ask around)

src/clojure/nrepl/socket.clj
rlbdv: Ahh, thanks - I didn't recall that you could use an underscore prefix like that.

rlbdv:
Looks like I may have finally found/fixed everything CI didn't like.  I also added a few additional commits (e.g. a convenience `make check` -- see the commit message) that may or may not be interesting.  If not, happy to remove them, or make other adjustments.

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,239 @@
...
+      (log msg)
+      (throw (ex-info msg {:nrepl/kind ::no-filesystem-sockets})))))
+
+(set! *warn-on-reflection* false)
+
+;; SocketAddress doesn't have .getPath until JDK 16, and we can't refer to
+;; AFUnixSocketAddress unconditionally in the junixsocket cases.  Also note that
+;; the former returns a Path, and the latter returns a string.
+
+(defn- get-path [addr] (.getPath addr))
```
vemv, line 98: I think something like this should work and address the nuances you mention, WDYT?

```clj
(defmacro getPath [addr]
  (if (Class/forName ...)
    `(let [^Path v# (.getPath (with-meta ~addr {:tag `com.Foo}))]
       ...)
    `(let [^String v# (.getPath (with-meta ~addr {:tag `com.Bar}))]
       ...)))
```

An important point is that all usages of the macro should perceive the same return type regardless of the `if` branch. e.g. always return string, or file, whatever, by using logic _within_ the macro.

This way, consumers of `get-path` can be abstracted away. Wouldn't look good to have to repeat assumptions/inspections in every usage of `get-path`; that would seem error-prone.

Makefile
```diff
@@ -20,11 +20,18 @@ cljfmt:
 	lein with-profile -user,+$(VERSION),+cljfmt cljfmt check
 
 kondo:
-	clj-kondo --lint src
+	if command -v clj-kondo > /dev/null; then \
+	  clj-kondo --lint src; \
+	else \
+	  lein with-profile +kondo run -m clj-kondo.main --lint src; \
```
vemv, line 26: Nice initiative 👍

You can go as far as eliminating the `clj-kondo` binary, see https://github.com/clojure-emacs/cider-nrepl/pull/693

vemv (commented):
Great work!

src/clojure/nrepl/socket.clj
```diff
@@ -0,0 +1,239 @@
...
+  (:import
+   (java.io BufferedInputStream BufferedOutputStream File)
+   (java.net InetSocketAddress ProtocolFamily ServerSocket Socket SocketAddress
+             StandardProtocolFamily URI)
+   (java.nio ByteBuffer)
+   (java.nio.file Path)
+   (java.nio.channels Channels ClosedChannelException NetworkChannel
+                      ServerSocketChannel SocketChannel)))
+
+(def orig-warn-on-reflection *warn-on-reflection*)
```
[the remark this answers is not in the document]
rlbdv: Unless I misunderstand, the problem is that it'll break the tests (CI) right now, since it looks like reflection warnings were recently made errors.  That's why I've been reworking the pr to remove all reflections.

Makefile
rlbdv: Oh, sure, if that's acceptable, I'd be happy to.  I'd just assumed that the binary had been intentionally preferred, and I'd have to check, but might even be able to have a `lein kondo` alias that does all that if it were desirable.

src/clojure/nrepl/socket.clj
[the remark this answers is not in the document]
vemv: They're not really part of a hard CI check right now. Still please do try getting rid of the reflection warnings, I think it's possible.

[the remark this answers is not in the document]
rlbdv: Hmm, I thought they were now.  I must have conflated the cause of some earlier failures.

src/clojure/nrepl/socket.clj
rlbdv: Interesting, but I wonder if this will work with current or future AOT, i.e. for this to be correct, the macro would have to always re-expand at run time, not compile time, since the available classes (and decision about the socket flavor) aren't known until then.

vemv: Very good observation :)

However shipping AOT-compiled libs tends to be very much frowned upon. I guess there can be exceptions, is nrepl one? @bbatsov

rlbdv: I suppose, unless we think of something better (or unless the above concern isn't relevant), I'd be tempted to just leave the reflection override for now, or to remove it and live with the warning, whichever's preferable.

vemv: Personally I'd be inclined to follow (or at least try) my suggestion, it should work. If you don't think it would work I'd appreciate a specific reason / your attempt

As for AOT:

* in principle I guess nrepl is not AOT-distributed, and/as most clojure libs never will or should
* there's the edge case of using nrepl in an AOT-compiled production app (for having a production repl)
  * however, for an issue to arise, the uberjar-er environment and the production environment should be executing different JDKs and/or computed classpaths
    *  this drift between "source" and "target" environments seems a notably bad practice, so having an error would nudge people into the right direction.

Makefile
rlbdv: Changed.

src/clojure/nrepl/socket.clj
rlbdv: I'm happy to handle it however's preferred, though personally, I'd lean against the extra complexity, since it's a rarely called function.  In the end, I may also need to defer to you with respect to whether or not it's safe enough in all relevant situations.

My assumption was that if you AOT compile the file, the macro will expand based on the compile time values, hard-coding the type coercions, which might or might not be correct at runtime.  Note that the "right" value can also be changed later, on any jvm invocation via the nrepl.test.junixsocket property.  Granted that shouldn't be used outside of testing, but it contributes to the additional complexity we have to think about, that we otherwise wouldn't if we didn't worry about the one reflection.

But, as mentioned, I'm happy to switch to the macro if you all feel confident it's sufficiently safe, and would prefer that.

vemv: Yes I understood what you're saying, thanks! What I said about AOT remains applying

The grander context is that we've been eliminating reflection from a variety of projects, from nrepl, to the most prominent projects in https://github.com/clojure-emacs/, to the clojurescript compiler (WIP), to having a dedicated linter in Eastwood, etc.

It takes one warning to foil the goal of a reflection-free tooling environment that won't pollute end-users' repls when they opt for a global `*warn-on-reflection* true` setting.

In that way a "no broken windows" approach might be considered pragmatic: it would take one broken window to set the precedent that introducing warnings is OK after all the fixing effort.

As always, this is just my input (only partly qualified as I'm a new clojure-emacs member) so we can always hear out @bbatsov when in doubt 😁

rlbdv: Another option I just thought of might be to mark the return as `^String`, which I think it should be, and handle any internal reflection manually, i.e. just rewrite get-path to look up the right method, etc.  Not beautiful, but it might work, and would avoid the need for the `set!` bits, etc.  I can try that if it seems interesting/desirable.

vemv: Not entirely sure of what you mean. Feel free to try a few things though, it's good to have options 👍

If you mean doing reflection by hand (using Java's reflection API instead of Clojure's), while nominally it would "prevent" warnings, in practice it would be the same, if not slightly worse because these can't be automatedly detected.

I've personally seen usages of Java's reflection API creating actual issues, especially with newer JDKs which are sensitive around `reflective access`. Typically they print a warning at JDK level, but it is foreseen that they will just fail in newer JDKs.

rlbdv: Nevermind - that won't work, I think, because we still don't know the actual class at compile time, and so even a reflection based `(.invoke method addr)` is going to cause a complaint because we don't (and can't) know the class of addr at compile time.

bbatsov:
It tooks us a while, but this PR is finally merged! 🥳 🎈 🚀

rlbdv:
Thanks much.

bbatsov:
I also cut 0.9.0-beta1, so people can play with it before we ship the final version.

src/clojure/nrepl/socket.clj
bbatsov: Sorry for not joining the party here earlier - during my vacation I try to pay less attention to my projects (and it seems I succeed :D). 

> However shipping AOT-compiled libs tends to be very much frowned upon. I guess there can be exceptions, is nrepl one?

Currently it's not and I don't have any plans to ship an AOT version in the foreseeable future. I guess we just have document the impact of this solution to a potential AOT version of nREPL. At any rate - I'm not very concerned about this. Still, it would have been nice if this could have been solved in some simpler way, but I'm short on ideas.

project.clj
```diff
@@ -40,12 +40,15 @@
                     ;; TODO: replicate koacha's version filter logic here
                     :test-selectors {:default (complement :min-java-version)}
                     :aliases {"test" "test2junit"}}
+             :junixsocket {:jvm-opts ["-Dnrepl.test.junixsocket=true"]
```
vemv, line 43: ⚠️ I belive that the new profile, which is a very nice addition, is not exercised in CircleCI currently.

Note that we use `make test` which doesn't use the `:test-all` alias.

While we're there we might want to re-review the PR and make sure that all code paths (depending on the JDK and available classes) are exercised.

@rlbdv , what do you think?

src/clojure/nrepl/socket.clj
vemv: Adding a note in the documentation would seem a nice middle ground - definitely better than an hypothetical "you're using AOT, go figure it out" 😄

project.clj
rlbdv: I think we should test it there -- I'd certainly intended to, and I have been here.  I'll see about it.

PR #205, opened 2020-07-07
Fix lookup of non-special symbol special forms

eerohele:
`let` is a special form, but not a special symbol. That is, the `special-symbol?` fn returns `false` on `let`.

(`let*` is the special symbol that backs `let`.)

A lookup on `let` returns the `:special-form` key. Its value is a boolean, which cannot be Bencoded. We therefore convert the boolean into a string first, the same way as with the `:macro` key.

### DIFF
--- a/test/clojure/nrepl/middleware/lookup_test.clj
+++ b/test/clojure/nrepl/middleware/lookup_test.clj
@@ -15,6 +15,7 @@
 
 (def-repl-test lookup-op
   (doseq [op [{:op "lookup" :sym "map" :ns "clojure.core"}
+              {:op "lookup" :sym "let" :ns "clojure.core"}
               {:op "lookup" :sym "map" :ns "nrepl.core"}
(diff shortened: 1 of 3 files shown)


test/clojure/nrepl/middleware/lookup_test.clj
```diff
@@ -15,6 +15,7 @@
 
 (def-repl-test lookup-op
   (doseq [op [{:op "lookup" :sym "map" :ns "clojure.core"}
+              {:op "lookup" :sym "let" :ns "clojure.core"}
```
bbatsov, line 18: I think it's going to be nice to update the unit test as well, not just the integration test.

eerohele: True, done. I initially figured it was only really the bencoding part that was the problem, but it's definitely good to have the unit test as well.

bbatsov:
Thanks!

PR #210, opened 2020-09-02
Fix for latest version of cider-nrepl/Java8

### LINKED ISSUE #206: Syntax error occured with nREPL 0.8.0 and cider-nrepl on Java8
Moved from https://github.com/clojure-emacs/orchard/issues/98

These changes may be the cause.
https://github.com/nrepl/nrepl/commit/c3ec4c6#diff-1b9ca3ef5a3a453d96b5f6c1b8c778a5L73-L138

@shen-tian Do you have a time to confirm?

## Expected behavior

`lein repl` will launch.

## Actual behavior

```
ERROR: Unhandled REPL handler exception processing message {:code (clojure.core/binding [clojure.core/*ns* (clojure.core/or (clojure.core/find-ns (clojure.core/symbol "reply.eval-mode
s.nrepl")) clojure.core/*ns*)] (set! *print-length* nil)), :id d9e84203-70ce-4144-8d6c-8c8aa8994446, :op eval, :session 71e6dc37-2c17-48a9-b70e-b42454c30d3d}
Syntax error compiling at (cider/nrepl/inlined_deps/orchard/v0v6v0/orchard/java/legacy_parser.clj:1:1).
```

Full error messages are here
https://gist.github.com/liquidz/6c2da4394b3a238947ad493c604e0d7a

- create project
```sh
docker run --rm -it clojure:openjdk-8-lein bash
lein new foo
cd foo
```
- edit `project.clj`
```clojure
(defproject foo "0.1.0-SNAPSHOT"
  :description "FIXME: write description"
  :url "http://example.com/FIXME"
  :license {:name "EPL-2.0 OR GPL-2.0-or-later WITH Classpath-exception-2.0"
            :url "https://www.eclipse.org/legal/epl-2.0/"}
  :dependencies [[org.clojure/clojure "1.10.1"]
                 [nrepl "0.8.0"]]
  :plugins [[cider/cider-nrepl "0.25.3"]]
  :repl-options {:init-ns foo.core})
```
- run `lein repl`

### Clojure version

- 1.10.1

### Java version

- 1.8

### Operating system

- Debian (Docker image: clojure:openjdk-8-lein)
- macOS Catalina

shen-tian:
Fixes #206, but breaks #113 (Edit: only breaks it for Java 8)

Given that Java 8 + Latest Cider support is kinda important, this feels like a reasonable tradeoff.

In terms of a more robust fix for #113, I think more research would be in order.

Tested by follow repro steps in #206 + lein install of this branch.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -32,10 +32,13 @@
   :profiles {:fastlane {:dependencies [[nrepl/fastlane "0.1.0"]]}
              :test {:dependencies [[com.hypirion/io "0.3.1"]
                                    [commons-net/commons-net "3.6"]
-                                   [lambdaisland/kaocha "1.0-612"]
-                                   [lambdaisland/kaocha-junit-xml "0.0-70"]]
+                                   [lambdaisland/kaocha "1.0.672"]
+                                   [lambdaisland/kaocha-junit-xml "0.0.76"]]
                     :plugins      [[test2junit "1.4.2"]]
                     :test2junit-output-dir "test-results"
+                    ;; This skips any tests that doesn't work on all java versions
+                    ;; TODO: replicate koacha's version filter logic here
+                    :test-selectors {:default (complement :min-java-version)}
                     :aliases {"test" "test2junit"}}
              ;; Clojure versions matrix
              :provided {:dependencies [[org.clojure/clojure "1.10.1"]]}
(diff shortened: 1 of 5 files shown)


bbatsov:
How about making this conditional based on the JDK? Might be a nice middle ground.

shen-tian:
Good idea... though I suspect there are cases other than orchard/java8 that doesn't work right now. Just not sure any of them matter.

I think the thing to do is to explicitly support a range of classpath modification mechanism/libraries.

shen-tian:
Sho. Forgot I had to port some sort of test selector logic back to `lein test` for Clojure 1.7 and 1.8

bbatsov:
Thanks for tackling this!

Issue #207, opened 2020-08-18
:arglists in lookup op response map cannot always be serialized into Bencode


eerohele:
## Steps to reproduce the problem

```clojure
(def output (java.io.ByteArrayOutputStream.))
(write-bencode output {{:foo 1} :bar})
;; or
(write-bencode output {[:foo] :bar})
```

Throws NPE:

```clojure
Execution error (NullPointerException) at nrepl.bencode/string>payload (bencode.clj:175).
null
{:via [{:type java.lang.NullPointerException,
        :at [nrepl.bencode$string_GT_payload invokeStatic "bencode.clj" 175]}],
 :trace [[nrepl.bencode$string_GT_payload invokeStatic "bencode.clj" 175]
         [nrepl.bencode$string_GT_payload invoke "bencode.clj" 173]
         [clojure.core$comp$fn__5807 invoke "core.clj" 2569]
         [clojure.core$juxt$fn__5822 invoke "core.clj" 2594]
         [clojure.core$map$fn__5866 invoke "core.clj" 2755]
         [clojure.lang.LazySeq sval "LazySeq.java" 42]
         [clojure.lang.LazySeq seq "LazySeq.java" 51]
         [clojure.lang.RT seq "RT.java" 535]
         [clojure.core$seq__5402 invokeStatic "core.clj" 137]
         [clojure.core.protocols$seq_reduce invokeStatic "protocols.clj" 24]
         [clojure.core.protocols$fn__8146 invokeStatic "protocols.clj" 75]
         [clojure.core.protocols$fn__8146 invoke "protocols.clj" 75]
         [clojure.core.protocols$fn__8088$G__8083__8101 invoke "protocols.clj" 13]
         [clojure.core$reduce invokeStatic "core.clj" 6828]
         [clojure.core$into invokeStatic "core.clj" 6895]
         [clojure.core$into invoke "core.clj" 6887]
         [nrepl.bencode$eval3461$fn__3462 invoke "bencode.clj" 405]
         [clojure.lang.MultiFn invoke "MultiFn.java" 234]
         [nrepl.bencode$eval3474 invokeStatic "form-init10775299742697767996.clj" 1]
         [nrepl.bencode$eval3474 invoke "form-init10775299742697767996.clj" 1]
         [clojure.lang.Compiler eval "Compiler.java" 7177]
         [clojure.lang.Compiler eval "Compiler.java" 7132]
         [clojure.core$eval invokeStatic "core.clj" 3214]
[...]

PR #211, opened 2020-09-14
Fix lookup op errors

eerohele:
The lookup op currently fails in these scenarios:

- The result includes the `:protocol` key (`clojure.core.protocols/datafy`)
- The result does not contain the `:file` key or the value of `:file` is `nil`
- The result has `:arglists` that is not serializable as Bencode

This PR should fix all of these issues. It also adds a test that looks up every public var in and ensures they can be serialized into a Bencode string. There's probably a smarter or a more exhaustive way to do that, though. I'm certainly open to suggestions on how to improve it.

With `:arglists`, I've chosen to retain `:arglists-str`, because that's what the only consumers of the lookup API that I know of use at the moment. Another option might be to convert `:arglists` into a string and drop `:arglists-str`.

Regarding this point from the comments in #207:

>Yeah, that's a good point. I've been using :arglists-str only as well in CIDER. Probably it's best to make both keys the same and add some notes explaining how this happened.

I'm not sure what's the best place to have a note like that. Ideas?

### DIFF
--- a/doc/modules/ROOT/pages/usage/misc.adoc
+++ b/doc/modules/ROOT/pages/usage/misc.adoc
@@ -49,6 +49,9 @@ user=> (set! nrepl.middleware.completion/*complete-fn* my.namespace/my-completio
 NOTE: Symbol look support was added in nREPL 0.8 and the API is still
 considered experimental.
 
+NOTE: In nREPL 0.8.1 and older, xref:ops.adoc#lookup[the `lookup` op] fails for some symbols
+because the result contains data that the xref:design/transports.adoc#bencode-transport[Bencode transport] does not support (see https://github.com/nrepl/nrepl/issues/207[nrepl/nrepl#207]). In versions newer than 0.8.1, `:arglists` and `:arglists-str` have identical string values. `:arglists-str` is subject to removal.
+
 nREPL bundles a lookup middleware with a flexible interface. By default it will use nREPL's own
 lookup logic in `nrepl.util.lookup`, but you can instruct nREPL to use
 another lookup function (e.g. `info` from `orchard`).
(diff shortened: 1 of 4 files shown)


eerohele:
>I think we should keep `:arglists` as well, mostly because it has a more intuitive name.

All right, so `:arglists` and `:arglists-str` would be equivalent? Will do.

>https://nrepl.org/nrepl/0.8/building_clients.html or https://nrepl.org/nrepl/0.8/usage/misc.html#symbol-lookup are great options.

Thanks! I'll look into updating the docs as well when I get the chance.

bbatsov:
Thanks!

eerohele:
All right, I restored `:arglists` (https://github.com/nrepl/nrepl/pull/211/commits/d2131c49edcf0209f96be621aa3ee42a6f8c843e) and gave updating the docs a shot (https://github.com/nrepl/nrepl/pull/211/commits/3eb6b56ed8cd7851f99b45c2fc1b6e78edd8d8cd). I'm not sure whether I messed up the AsciiDoc link syntax, though, since I couldn't figure out how to use Antora to preview my changes locally.

bbatsov:
> AsciiDoc link syntax, though, since I couldn't figure out how to use Antora to preview my changes locally.

Did you see the instructions here https://nrepl.org/nrepl/about/contributing.html#working-on-the-manual ?

eerohele:
I did, but the way I understood them was that following those instructions clones the nREPL Git repo and generates the documentation from the `docs` folder there. I'm not sure how to run Antora to preview my changes without having to commit anything.

Although I guess I could just point Antora to my PR branch now...

bbatsov:
> I'm not sure how to run Antora to preview my changes without having to commit anything.

You can just point it to a local branch. :-) The repo config is in the `antora-playbook.yml`. There's no concept of "preview", as it's a static site generator - you run it and you just open the generated HTML.

eerohele:
>You can just point it to a local branch. :-) The repo config is in the antora-playbook.yml. There's no concept of "preview", as it's a static site generator - you run it and you just open the generated HTML.

Yeah, I actually just figured it out. :)

As a note to the future me, I made these changes to the nrepl.org repository to make it work:

```diff
diff --git a/antora-playbook.yml b/antora-playbook.yml
index 9e7241f..b814570 100644
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -1,14 +1,13 @@
 site:
   title: nREPL
   url: https://nrepl.org/
   start_page: 0.8@nrepl::index.adoc
   keys:
     google_analytics: 'UA-128069102-1'
 content:
   sources:
-  - url: https://github.com/nrepl/nrepl.git
-    branches: ['master']
-    tags: ['0.5.3', '0.6.0', '0.7.0', '0.8.1-antora']
+  - url: /path/to/local/nrepl/clone
+    branches: ['issues/207']
     start_path: doc
 asciidoc:
   attributes:
```

Anyway, I did mess it up a bit initially, so I just pushed a fix. Should be OK now, unless you want to change something about the note.

bbatsov:
That's fine. Thanks for tackling this issue! :bowing_man:

bbatsov:
I released 0.8.2 with your fixes. I'll have to update `cider-nrepl` to match the arglists change eventually.

PR #214, opened 2020-10-20
Make thread factory configurable

fp7:
Google App Engine does not allow threads to call the app engine apis unless the
whole thread chain were created by there own ThreadFactory. This behaviour
disallows using threads via `future`.

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*


src/clojure/nrepl/threads.clj
```diff
@@ -0,0 +1,4 @@
+(ns nrepl.threads)
+
+(def ^{:dynamic true
```
bbatsov, line 3: I don't think that a separate namespace is warranted for just one configuration value, so I'd probably move this to the session namespace.

bbatsov:
It would definitely be great to be able to run nREPL in the Google App Engine, so thanks for tackling this. My main concern for the suggested implementation is that it adds complexity to the server namespace, and the ideally this should be mostly constained to the sessions which are actually leveraging the thread pool. I'm wondering if it won't be better to avoid the dynamic bindings and aim for just passing the thread factory directly via some optional param. In general I'm not very fond of abusing dynamic bindings and try to use them infrequently. 

Also, a change like this should definitely be documented in the manual, otherwise no one will ever figure out that something like this is supported.

bbatsov:
@shen-tian @SevereOverfl0w I'd love to hear your takes on this as well.

shen-tian:
I'm not familiar with the problem being solved, and whether/how that is solved here. So would defer to people more knowledgable on that.

In terms of nREPL codebase. At its heart, we seem to want a slightly different behaviour for `(future ...)`. I'd capture that intent with making this a macro:

```clojure
(.start
 (.newThread threads/*platform-thread-factory*
             (bound-fn [] ...)))
```
Maybe this will justify the `nrepl.threads` ns? Second @bbatsov on this should have some clear "why" documentation. 

The net effect would be "in nREPL, we use `platform-future` instead of `future` (or whatever we call it), for _complicated reasons_, read this to see why."

I think I'm ok with use of dynamic bindings here, as long as everything is well documented.

Thanks for tackling this.

fp7:
Thanks for your feedback and sorry for the delay. I was busy the last weeks.

I incorporated your feedback and removed the dynamic binding and also introduced a macro that's basically a copy of clojure.core/future. It used `CompletableFuture` which is only available in java8. Is java7 still supported by nrepl? 
I didn't know where to put it because it is only used in the server ns. So it is at the top for now. Feel free to suggest a better place.

And I will provide some documentation next.

shen-tian:
nREPL [supports Java 8, 11, and latest](https://nrepl.org/nrepl/0.8/about/compatibility.html).

src/clojure/nrepl/middleware/session.clj
```diff
@@ -61,24 +75,22 @@
...
       :or {keep-alive 30000
            queue (SynchronousQueue.)}}]
-  (let [^ThreadFactory thread-factory (or thread-factory (configure-thread-factory))]
+  (let [^ThreadFactory thread-factory (or thread-factory (configure-thread-factory thread-factory))]
     (ThreadPoolExecutor. 1 Integer/MAX_VALUE
                          (long 30000) TimeUnit/MILLISECONDS
                          ^BlockingQueue queue
                          thread-factory)))
 
-(def default-executor "Delay containing the default Executor." (delay (configure-executor)))
```
bbatsov, line 70: Deleting this seems like a breaking API change to me.

bbatsov:
I like the updated version of the PR, but it I'll have to review it more carefully to make sure it doesn't introduce any breaking API changes. I see some public function signatures that were changed, which we should be careful with it. Keep in mind that backwards API compatibility in an absolute must for any change to nREPL's internals.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -61,24 +75,22 @@
...
-
 (defn default-exec
   "Submits a task for execution using #'default-executor.
    The submitted task is made of:
    * an id (typically the message id),
    * thunk, a Runnable, the task itself,
    * ack, another Runnable, ran to notify of successful execution of thunk.
    The thunk/ack split is meaningful for interruptible eval: only the thunk can be interrupted."
-  [id ^Runnable thunk ^Runnable ack]
+  [^ExecutorService es id ^Runnable thunk ^Runnable ack]
```
bbatsov, line 91: This seems like a breaking change as well.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -179,7 +191,7 @@
                  (.getState t))
       (.stop t))))
 
-(defn session-exec
+(defn- session-exec
```
bbatsov, line 194: Making a public method private is a breaking change.

src/clojure/nrepl/server.clj
```diff
@@ -16,10 +17,61 @@
...
+  return it on all subsequent calls to deref/@. If the computation has
+  not yet finished, calls to deref/@ will block, unless the variant of
+  deref with timeout is used. See also - realized?.
+
+  Copied from clojure.core and modified to accept a thread-factory"
+  [thread-factory & body] `(platform-future-call ~thread-factory (^{:once true} fn* [] ~@body)))
+
 (defn handle*
-  [msg handler transport]
+  [msg handler transport thread-factory]
```
bbatsov, line 70: This extra positional argument seems like a breaking change.

src/clojure/nrepl/server.clj
```diff
@@ -34,10 +86,11 @@
 (defn handle
   "Handles requests received via [transport] using [handler].
    Returns nil when [recv] returns nil for the given transport."
-  [handler transport]
+  [handler transport thread-factory]
```
bbatsov, line 89: Same here.

bbatsov:
I left a few more remarks. Generally it seems to me it'd be a good idea for you to test the impact of your changes against middleware like `cider-nrepl`, `refactor-nrepl`, `iced-nrepl`, `piggieback`, etc, as you can't know which of the public API methods they are currently using and we definitely don't want to break any of them. I guess is most cases the `tread-factory` should be an optional parameter (e.g. coming from a different arity of the function).

bbatsov:
Closing due to lack of activity. @fp7 feel free to reopen if you ever come back to this.

PR #217, opened 2020-11-22
Keyword completion

eerohele:

- [x] All tests are passing

See also the review comments I left on my own changes.

### DIFF
--- a/eastwood.clj
+++ b/eastwood.clj
@@ -8,3 +8,7 @@
 (disable-warning
  {:linter :deprecations
   :symbol-matches #{#"public final void java.lang.Thread.stop\(\)"}})
+
+(disable-warning
+ {:linter :unused-ret-vals
+  :if-inside-macroexpansion-of #{'clojure.test/deftest}})
--- a/src/clojure/nrepl/util/completion.clj
+++ b/src/clojure/nrepl/util/completion.clj
@@ -144,7 +144,7 @@
     {:candidate name :type :static-method}))
 
 (defn scoped-candidates
-  [prefix ns]
+  [^String prefix ns]
   (when-let [prefix-scope (first (.split prefix "/"))]
     (let [scope (symbol prefix-scope)]
       (map #(update % :candidate (fn [c] (str scope "/" c)))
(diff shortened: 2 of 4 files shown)


eastwood.clj
```diff
@@ -8,3 +8,7 @@
 (disable-warning
  {:linter :deprecations
   :symbol-matches #{#"public final void java.lang.Thread.stop\(\)"}})
+
+(disable-warning
```
eerohele, line 12: I'm not very familiar with Eastwood, so I'm not sure whether this is a sensible way to disable warnings for the new tests.

src/clojure/nrepl/util/completion.clj
```diff
@@ -8,8 +8,93 @@
...
+
+;; Code adapted from Compliment (https://github.com/alexander-yakushev/compliment)
+
+(defn annotate-keyword
+  [kw]
+  {:candidate kw :type :keyword})
+
+(defn all-keywords
+  []
+  (let [^Field field (.getDeclaredField clojure.lang.Keyword "table")]
```
eerohele, line 23: Compliment [memoizes](https://github.com/alexander-yakushev/compliment/blob/a112bb85b748d4a34da89ee783e7c64d8183dab7/src/compliment/sources/keywords.clj#L7) this function, but I'm not sure how that works exactly. If I create a new keyword, the memoized function seems to return it as well. Not sure why.

I've omitted memoization for now, but can add it if necessary.

src/clojure/nrepl/util/completion.clj
```diff
@@ -8,8 +8,93 @@
...
+
+(defn keyword-candidates
+  [^String prefix ns]
+  (assert (string? prefix))
+  (let [double-colon? (.startsWith prefix "::")
+        single-colon? (.startsWith prefix ":")
+        slash-pos (.indexOf prefix "/")]
+    (cond
+      (and double-colon? (pos? slash-pos))
+      (let [ns-alias (subs prefix 2 slash-pos)
```
eerohele, line 87: Compliment [uses a regex](https://github.com/alexander-yakushev/compliment/blob/a112bb85b748d4a34da89ee783e7c64d8183dab7/src/compliment/sources/keywords.clj#L42) for this. I figured that since we need to find the position of the slash anyway, why not just use `subs`.

bbatsov:
Your changes seem solid! Thanks!

PR #226, opened 2021-01-08
Add doc and arglists to completion responses

eerohele:
- [x] All tests are passing

What it says on the tin. Allows clients to implement a richer auto-completion experience. Like this:

Screenshot 2021-01-08 at 23 10 01

One thing to consider is whether we should return `:doc` and `:arglists` by default (as this PR does), or only if the client asks for them. I can't imagine sending them by default causing much harm, unless there's a var with an absolutely massive docstring that causes the client to choke or something.

The `nrepl.util.completion/completions` function does take an options map, which we could leverage to do something like `{:keys #{:doc :arglists}}`. I've left that out for now, but I can certainly implement it if necessary.

Another option is to use the same options map to give the clients the option of opting out of those keys. That would be fairly straightforward to implement later and retain backwards compatibility if necessary, I think.

### DIFF
--- a/doc/modules/ROOT/pages/usage/misc.adoc
+++ b/doc/modules/ROOT/pages/usage/misc.adoc
@@ -44,6 +44,28 @@ setting `nrepl.middleware.completion/*complete-fn*`.
 user=> (set! nrepl.middleware.completion/*complete-fn* my.namespace/my-completion)
 ----
 
+=== Options
+
+The default completion function accepts the `extra-metadata` option. It allows clients to request additional metadata about completion items. For example, to request docstrings and argument lists for applicable completion items:
+
+[source,clojure]
+----
+;; -> client sends completion op
+{:op "completions"
+ :prefix "ma"
+ :ns "clojure.core"
+ :options {:extra-metadata ["arglists" "doc"]}}
+
+;; <- server returns completion items
+{:completions
+ [,,,
+  {:candidate "map"
+   :type "function"
+   :arglists "([f] [f coll] [f c1 c2] [f c1 c2 c3] [f c1 c2 c3 & colls])"
+   :doc "Returns a lazy sequence..."}
+  ,,,]}
+----
+
 == Symbol Lookup
 
 NOTE: Symbol look support was added in nREPL 0.8 and the API is still
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -103,7 +103,8 @@
   [path]
   (or (some-> path io/resource str) path))
 
-(defn normalize-meta
+(defn sanitize-meta
+  "Sanitize a Clojure metadata map such that it can be bencoded."
   [m]
   (-> m
       (select-keys safe-var-metadata)
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -99,10 +99,6 @@
   [:ns :name :doc :file :arglists :forms :macro :special-form
    :protocol :line :column :added :deprecated :resource])
 
-(defn resolve-file
-  [path]
-  (or (some-> path io/resource str) path))
-
 (defn sanitize-meta
   "Sanitize a Clojure metadata map such that it can be bencoded."
   [m]
@@ -111,7 +107,7 @@
       (update :ns str)
       (update :name str)
       (update :protocol str)
-      (update :file resolve-file)
+      (update :file #(or (some-> % io/resource str) %))
       (cond-> (:macro m) (update :macro str))
       (cond-> (:special-form m) (update :special-form str))
       (assoc :arglists-str (str (:arglists m)))
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -3,7 +3,8 @@
   useful for anyone extending it)."
   {:author "Chas Emerick"}
   (:refer-clojure :exclude [requiring-resolve])
-  (:require [clojure.string :as str]))
+  (:require [clojure.java.io :as io]
+            [clojure.string :as str]))
 
 (defn log
   [ex & msgs]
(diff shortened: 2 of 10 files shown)


src/clojure/nrepl/misc.clj
```diff
@@ -89,3 +90,28 @@
   []
   (.startsWith (System/getProperty "java.runtime.version")
                "1.8"))
+
+(def var-meta-whitelist
```
bbatsov, line 94: Let's avoid the `whitelist` termnology where. I guess we can call this `safe-var-metadata` or something along those lines.

bbatsov: (and I guess the original oversight was mine)

bbatsov:
> One thing to consider is whether we should return :doc and :arglists by default (as this PR does), or only if the client asks for them. I can't imagine sending them by default causing much harm, unless there's a var with an absolutely massive docstring that causes the client to choke or something.

I think that's fine. 

> The nrepl.util.completion/completions function does take an options map, which we could leverage to do something like {:keys #{:doc :arglists}}. I've left that out for now, but I can certainly implement it if necessary.

Yeah, I think doing something like this down the road would nice for clients who want to optimize their payloads. 

> Another option is to use the same options map to give the clients the option of opting out of those keys. That would be fairly straightforward to implement later and retain backwards compatibility if necessary, I think.

I've always preferred to say what you need instead of what you don't need. Seems more straightforward to me as it doesn't assume you know the actual return format, which might change, as it's about to change now.

src/clojure/nrepl/misc.clj
```diff
@@ -89,3 +90,28 @@
...
+  with bencode. We also optimize the response payloads by not returning
+  redundant metadata."
+  [:ns :name :doc :file :arglists :forms :macro :special-form
+   :protocol :line :column :added :deprecated :resource])
+
+(defn resolve-file
+  [path]
+  (or (some-> path io/resource str) path))
+
+(defn normalize-meta
```
bbatsov, line 106: I wonder if this function shouldn't get a better name, now that it's getting promoted to a (more) public api.

bbatsov: Probably it deserves a docstring now as well.

src/clojure/nrepl/misc.clj
```diff
@@ -89,3 +90,28 @@
...
+
+(def var-meta-whitelist
+  "A list of var metadata attributes are safe to return to the clients.
+  We need to guard ourselves against EDN data that's not encodeable/decodable
+  with bencode. We also optimize the response payloads by not returning
+  redundant metadata."
+  [:ns :name :doc :file :arglists :forms :macro :special-form
+   :protocol :line :column :added :deprecated :resource])
+
+(defn resolve-file
```
bbatsov, line 102: We should probably add a dostring and a test for this one as well.

bbatsov:
Apart from my small remarks the PRs seems in a pretty good shape.

eerohele:
Good points.

I think it's better if I change this PR so that clients have to request the `:arglists` and `:docs` keys explicitly to receive them. That way we can avoid making a breaking change in the future. I'll try to make that change and the other changes you proposed soon.

In the meantime, let the bike shedding on the option key name commence! Compliment uses `*extra-metadata*`, so how about something like this?

```clojure
{:op "completions" :ns "clojure.core" :prefix "map" :options {"extra-meta" ["arglists" "doc"]}}
```

src/clojure/nrepl/misc.clj
eerohele: `sanitize-meta`?

src/clojure/nrepl/misc.clj
eerohele: Good point, will do.

eerohele: In fact, since `normalize-meta`/`sanitise-meta` is the only call site, I think we can just inline this function instead. As in:

```clojure
(update :file #(or (some-> % io/resource str) %))
```

After all, the only thing the test for `normalize-meta` currently tests for is actually the behavior of the `resolve-file` function.

eerohele:
I pushed a first cut of the proposed changes. I'll take another look at them tomorrow with fresh eyes, but I just wanted to get them out there already.

Relocating vars from `nrepl.util.completion` to `nrepl.misc` is, of course, a breaking change, as is modifying the arities of some of the functions in `nrepl.util.completion`. However, I kinda think of `nrepl.util.completion` as a "private" namespace, even though it isn't explicitly marked as such, and `nrepl.middleware.completion` as its public interface. However, if you think we should avoid breaking changes, let me know and I'll fix them. I think it's unlikely that anyone's calling the functions in `nrepl.util.completion` directly, though, and a quick search on GitHub supports that.

bbatsov:
I think the breaking changes are fine here, as I agree it's unlikely someone's using this API directly, plus there's a note that `completions` and `lookup` are currently considered experimental and subject to changes.

eerohele:
Foiled by cljfmt once again. Curses!

doc/modules/ROOT/pages/usage/misc.adoc
```diff
@@ -44,6 +44,26 @@ setting `nrepl.middleware.completion/*complete-fn*`.
...
+
+The default completion function accepts the `extra-meta` option. It allows clients to request additional metadata about completion items. For example, to request docstrings and argument lists for applicable completion items:
+
+[source,clojure]
+----
+;; -> client sends completion op
+{:op "completions"
+ :prefix "ma"
+ :ns "clojure.core"
+ :extra-meta ["arglists" "doc"]}
```
bbatsov, line 57: The example seems wrong, as `:extra-meta` should in a map. As this is public API we might also use a more descriptive name like `extra-metadata` or `additional-metadata`. `Meta` is a bit too `meta`. 😆

bbatsov:
The changes seem good to me overall. You might want to extend the middleware descriptor to mention the `:extra-metadata` option and regen the ops docs.

eerohele:
>The example seems wrong, as :extra-meta should in a map.

Agh, that's right, it should be under `:options`. Will fix.

>Meta is a bit too meta.

I figured since `meta` is so common in Clojure... but fair enough, I'll sort it out. 🙂

Thanks for the notes. I'll leave the fixes for tomorrow.

eerohele:
Done.

bbatsov:
Looks good. I'll cut some alpha release soon.

bbatsov:
0.9.0-snapshot is out!

PR #227, opened 2021-01-14
Fix completion for static class members

eerohele:
`(resolve sym)` is the same as `(ns-resolve *ns* &env sym)`. That is, prior to this fix, the `resolve-class` fn attempted to resolve the symbol against the ns stored in the `*ns*` var (often the `user` ns), not the ns that was passed in via the completion op.

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * Add keyword completion support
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Nice catch!

PR #233, opened 2021-03-05
Remove duplicate ids.

th994:

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/src/clojure/nrepl/core.clj
+++ b/src/clojure/nrepl/core.clj
@@ -84,7 +84,7 @@
    \"done\" :status."
   [client {:keys [id] :as msg :or {id (uuid)}}]
   (let [f (delimited-transport-seq client #{"done" :done} {:id id})]
-    (f (assoc msg :id id))))
+    (f msg)))
 
 (defn new-session
   "Provokes the creation and retention of a new session, optionally as a clone


th994:
Since the following code in delimited-transport-seq merges the id, there is no need to create a new message with the id.

```
#(merge % delimited-slots)
```

I'm sorry if this is a misunderstanding.

bbatsov:
I see that tests are passing, so most likely you're right. How did you notice this?

th994:
Thanks for the confirmation.

I'm curious about the behavior of nrepl and noticed it while reading the code from scratch.

There was no problem with the operation.

bbatsov:
Thanks!

bbatsov:
Hmm, after merging it I realized that now this doesn't quite work like the description of the function, so probably this has to be updated as well. (the description explicitly mentions adding the id to the message)

th994:
Thanks.

I'll try to fix the description too.
I'm not good at English, so I'd appreciate your suggestions.

bbatsov:
Sure! I guess the most important thing is to make it clear how the a `message` obtains its id. Please, also add a docstring for `delimited-transport-seq`, so it's clearer for casual readers of the code what is its purpose.

PR #234, opened 2021-03-07
Fix and Add docstring to nrepl.core

th994:
…mited-transport-seq

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/src/clojure/nrepl/core.clj
+++ b/src/clojure/nrepl/core.clj
@@ -64,14 +64,14 @@
     (concat head (take 1 tail))))
 
 (defn- delimited-transport-seq
-  "Sends a message via client and returns the head of client's response seq
-   by following the steps described below.
+  "Returns a function of one arument that performs described below.
+   The following \"message\" is the argument of the function returned by this function.
 
-    - Add an :id to the message(argument)
-    - Sends a message(argument) via client
-    - Filter only messages(response) related to the :id of client's response seq
-    - Returns head of the seq that will terminate upon receipt
-      of a \"done\" :status"
+    - Merge delimited-slots to the message
+    - Sends a message via client
+    - Filter only items related to the delimited-slots of client's response seq
+    - Returns head of the seq that will terminate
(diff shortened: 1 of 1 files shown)


th994:
Added and fixed docstring to changes in Pull Request #233.

The docstring in nrepl.core/message was so detailed that it was almost identical to the docstring in nrepl.core/delimited-transport-seq.

Is there any good solution for this?

src/clojure/nrepl/core.clj
```diff
@@ -64,6 +64,14 @@
     (concat head (take 1 tail))))
 
 (defn- delimited-transport-seq
+  "Sends a message via client and returns the head of client's response seq
```
bbatsov, line 67: Seems to me this returns a function that will do all of the above. The actual `delimited-transport-seq` doesn't do any sending. It's also not returning a seq, which makes its name slightly confusing, but that's not a big deal I guess (and I don't have an idea for a better name).

bbatsov:
Good points. I think the best thing to do here is to differentiate between building the function that does the message sending over the client from the actual process of invoking this function, which results in a message being sent.

src/clojure/nrepl/core.clj
```diff
@@ -64,6 +64,14 @@
...
+   by following the steps described below.
+
+    - Add an :id to the message(argument)
```
bbatsov, line 70: I guess here it'd be good to clarify that the argument in question is the argument of the function that this function returns. A bit meta, right? :-)

th994:
Thanks for the confirmation and the very clear advice.

I modified the docstring with the following in mind.

- delimited-transport-seq builds a function, and that function will be invoked by nrepl.core/message.
- The 'message' that appears in the docstring of delimited-transport-seq is the argument of the function returned by delimited-transport-seq.

Since delimited-transport-seq is also used in nrepl.core/client-session, excluded specific keys from the docstring such as :id.

Please Let me know if there are any problems

bbatsov:
Thanks!

Issue #231, opened 2021-02-17
nrepl.misc/sanitize-meta throws when file meta is a java.net.URL


eerohele:
HugSQL, for example, creates symbols whose `:file` meta is a `java.net.URL` instance, which `nrepl.misc/sanitize-meta` can't handle.

bbatsov:
I guess we should convert URLs to strings.

PR #236, opened 2021-03-15
Sanitize url error fix

solar05:

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

#231

### DIFF
--- a/src/clojure/nrepl/misc.clj
+++ b/src/clojure/nrepl/misc.clj
@@ -3,7 +3,8 @@
   useful for anyone extending it)."
   {:author "Chas Emerick"}
   (:refer-clojure :exclude [requiring-resolve])
-  (:require [clojure.java.io :as io]))
+  (:require [clojure.java.io :as io])
+  (:import [java.net URL]))
 
 (defn log
   [ex & msgs]
@@ -98,6 +99,12 @@
   [:ns :name :doc :file :arglists :forms :macro :special-form
    :protocol :line :column :added :deprecated :resource])
 
+(defn- check-filetype [file]
+  (cond
+    (instance? URL file) file
+    (instance? java.io.File file) file
+    :else (io/resource file)))
+
 (defn sanitize-meta
   "Sanitize a Clojure metadata map such that it can be bencoded."
   [m]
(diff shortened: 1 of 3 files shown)


bbatsov:
Please, also add a unit test for the URL case.

bbatsov:
You should also mention the fix in the changelog.

src/clojure/nrepl/misc.clj
```diff
@@ -106,7 +113,7 @@
       (update :ns str)
       (update :name str)
       (update :protocol str)
-      (update :file #(or (some-> (str %) io/resource str) %))
+      (update :file #(or (some-> % check-filetype str) %))
```
bbatsov, line 116: I'd replace this entire lambda with a function, so you can extend the function you introduced and name it something like `handle-file-meta` or something like this.

solar05: Did it, but I think I got it worse :\

src/clojure/nrepl/misc.clj
```diff
@@ -98,6 +99,14 @@
   [:ns :name :doc :file :arglists :forms :macro :special-form
    :protocol :line :column :added :deprecated :resource])
 
+(defn handle-file-meta [file]
+  (cond
```
bbatsov, line 103: You can use `case` instead of `cond` and a `let` to remove most of the duplication.

src/clojure/nrepl/misc.clj
```diff
@@ -99,14 +99,24 @@
...
+(defn handle-file-meta-1 [file]
   (cond
     (nil? file) file
     (instance? URL file) (str file)
     (instance? java.io.File file) (str file)
     (not (nil? (io/resource file))) (str (io/resource file))
     :else file))
 
+(defn handle-file-meta [file]
+  (let [type-map {java.net.URL #(str %)
```
bbatsov, line 111: My point was that you don't really need to apply `str` explicitly for each case, but rather you can just apply `str` once to whatever the `result` from `cond`. You can leave the `some` outside of the function if you want or put the nil handling outside the `cond` logic, so it wouldn't be converted to an emptry string.

solar05: Hmm, last time I made a function with same logic, but for some reason the local tests did not pass, it seems I forgot to save the file and tests ran on the wrong version of code, sorry for constantly distracting you with my PR :(

bbatsov:
Thanks!

Issue #230, opened 2021-02-17
Completion op should return namespace-not-found when appropriate


eerohele:
Given a client that sends an op like this:

```clojure
{"id" "1" "op" "completions" "prefix" "a" "ns" "my.unknown.ns"}
```

If the nREPL server doesn't know `my.unknown.ns`, it returns:

```clojure
{"id" "1" "status" ["completion-error" "done"]}
```

It would be useful if the server returned something like this instead:

```clojure
{"id" "1" "status" ["namespace-not-found" "done"]}
```

That way, clients could e.g. evaluate the `ns` form in the current view and then retry the completion op.

Currently, clients have no way of distinguishing missing namespace errors from other types of completion errors.

bbatsov:
Yeah, that's a good point and would be easy to fix this. I'm guessing something like:

```
{"id" "1" "status" ["completion-error" "namespace-not-found" "done"]}
```

would be even better. The status is a set and we can have as many statuses as we want there.

eerohele:
Same thing for the lookup op.

bbatsov:
Fixed in #238.

PR #238, opened 2021-07-03
Expand completion and lookup error message when ns not found

solar05:

Issue #230 

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,22 +2,16 @@
 
 ## master (unreleased)
 
-### New feature
-
-* [#238](https://github.com/nrepl/nrepl/pull/238): Expand completion and lookup error message when ns not found.
-
-### Bugs fixed
-
-* [#231](https://github.com/nrepl/nrepl/issues/231): Fix sanitize error when file is java.net.URL.
-
 ### New features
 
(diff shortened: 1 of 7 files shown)


src/clojure/nrepl/middleware/lookup.clj
```diff
@@ -33,7 +33,7 @@
           lookup-fn (or (and lookup-fn (misc/requiring-resolve (symbol lookup-fn))) *lookup-fn*)]
       (response-for msg {:status :done :info (lookup-fn ns sym)}))
     (catch Exception _e
-      (response-for msg {:status #{:done :lookup-error}}))))
+      (response-for msg {:status #{:done :lookup-error :namespace-not-found}}))))
```
bbatsov, line 36: As this is a catch all, this can be some other exception as well (e.g. something might blow up in the `lookup-fn`). For proper granularity this should be contingent on the ns not being present, but I guess this would mean adding some additional check.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New feature
```
bbatsov, line 5: It seems we already have this section a bit down.

bbatsov:
Thanks!

PR #240, opened 2021-07-21
Upgrade Eastwood

vemv:
Introduces a couple new linters:

https://github.com/jonase/eastwood#non-dynamic-earmuffs
https://github.com/jonase/eastwood#reflection

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -77,7 +77,8 @@
                                          testing-print [[:inner 0]]}}}
 
              :eastwood [:test
-                        {:plugins [[jonase/eastwood "0.7.1"]]
+                        {:plugins [[jonase/eastwood "0.9.4"]]
                          :global-vars {*warn-on-reflection* true}
                          :eastwood {:config-files ["eastwood.clj"]
-                                    :ignored-faults {:unused-ret-vals {nrepl.util.completion-test true}}}}]})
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -77,7 +77,8 @@
                                          testing-print [[:inner 0]]}}}
 
              :eastwood [:test
-                        {:plugins [[jonase/eastwood "0.7.1"]]
+                        {:plugins [[jonase/eastwood "0.9.4"]]
                          :global-vars {*warn-on-reflection* true}
                          :eastwood {:config-files ["eastwood.clj"]
-                                    :ignored-faults {:unused-ret-vals {nrepl.util.completion-test true}}}}]})
+                                    :ignored-faults {:non-dynamic-earmuffs {nrepl.middleware.load-file true}
```
vemv, line 83: Silences:

```
src/clojure/nrepl/middleware/load_file.clj:53:1: non-dynamic-earmuffs: #'nrepl.middleware.load-file/load-file-code should use the earmuff naming convention: please use #'nrepl.middleware.load-file/*load-file-code* instead.
```

The suggested rename seems on point, however I'd be afraid of breaking APIs.

bbatsov: I don't think any middleware/transport would have a reason to use this, so renaming it is probably fine. At the very least we can fix the name and keep the old one as a depreciated alias.

bbatsov:
Thanks!

PR #239, opened 2021-07-11
Fix namespace resolution in the cmdline repl (#208)

mikemar10:
The namespace resolution in `nrepl.cmdline/run-repl` was broken due to
usage of `clojure.core/read` which was performing all namespace
resolution in the user namespace. This has been fixed by rebinding the
`*ns*` var in the main loop any time that the nrepl server changes
namespaces.

Before submitting a PR make sure the following things have been done:

- [x] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [x] You've added tests to cover your change(s)
- [x] All tests are passing
- [x] The new code is not generating reflection warnings
- [x] You've updated the [changelog](../CHANGELOG.md)(that only applies to user-visible changes)

Thanks!

*If you're just starting out to hack on nREPL you might find this [section of its
manual][1] extremely useful.*

[1]: https:/nrepl.org/nrepl/hacking_on_nrepl.html

PR #241, opened 2021-08-01
Elaborate on filesystem socket permission issues

rlbdv:
I wondered if we might want to elaborate a bit on the filesystem socket permission constraints, also discussed in Linux's `unix(7)`.  I also didn't know if having non-ascii "dashes" in the `start-server` docstring was intentional, but I hadn't noticed before, and ao changed `:socket` to match the others.

From the manpage:
```
Pathname socket ownership and permissions
    In  the  Linux  implementation,  pathname sockets honor the permissions of the directory
    they are in.  Creation of a new socket fails if the process  does  not  have  write  and
    search (execute) permission on the directory in which the socket is created.

    On Linux, connecting to a stream socket object requires write permission on that socket;
    sending a datagram to a datagram socket  likewise  requires  write  permission  on  that
    socket.   POSIX  does  not  make  any statement about the effect of the permissions on a
    socket file, and on some systems (e.g., older BSDs), the socket permissions are ignored.
    Portable programs should not rely on this feature for security.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,10 +67,6 @@ commands:
           command: |
             sudo apt-get install make
       - run:
-          name: Install clj-kondo
-          command: |
-           sudo curl -sLO https://raw.githubusercontent.com/clj-kondo/clj-kondo/master/script/install-clj-kondo && sudo chmod +x install-clj-kondo && sudo ./install-clj-kondo
-      - run:
           name: Generate Cache Checksum
           command: |
             for file in << parameters.files >>
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,10 @@ test:
 
 ifeq ($(VERSION),$(filter $(VERSION),1.9 1.10 master))
 	lein with-profile -user,+$(VERSION),+test run -m kaocha.runner
+	lein with-profile -user,+$(VERSION),+test,+junixsocket run -m kaocha.runner
 else
 	lein with-profile -user,+$(VERSION),+test test
+	lein with-profile -user,+$(VERSION),+test,+junixsocket test
 endif
 
 eastwood:
--- a/doc/modules/ROOT/pages/usage/server.adoc
+++ b/doc/modules/ROOT/pages/usage/server.adoc
@@ -216,7 +216,8 @@ You can also ask nREPL to listen on a UNIX domain (filesystem) socket
 with the `:socket` keyword (if you're using JDK 16 or newer add a
 https://kohlschutter.github.io/junixsocket/[junixsocket] dependency),
 which should be as secure as the access to the socket\'s parent
-directories:
+directories (POSIX doesn't specify the effect of the socket file's
+permissions (if any), and some systems have ignored them):
 
 [source,clojure]
 ----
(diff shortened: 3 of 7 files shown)


rlbdv:
Test failures appear to be the reflection warnings again.  I thought we'd resolved all those, but while `lein check` is fine, it looks like eastwood is not.  I don't know if that's new, or if I was only working from `lein check` before, but I did notice a new version of easwood being fetched when I ran the tests locally.  Please let me know how you'd prefer I try to handle that.  Thanks.

bbatsov:
I'm fine with whatever solution for the reflection warning you prefer. I think @vemv mentioned something on the subject in a different PR. (#239)

rlbdv:
Well, the `(set! *warn-on-reflection* false)` "works" for `lein check` but doesn't work for (current?) eastwood, and I don't know what will, but I can try (again) to figure something else out.  I believe the difficulty is that we can't mention classes that might or might not be available in type annotations as far as I can tell right now, and eastwood doesn't respect that `set!` override.

And really, we *can't* legitimately get rid of reflection warnings in these cases?  That is, the classes really are unknowable at compile time.

I can imagine some other hacks, but I don't know if they'll work, or if you'd want them.  i.e. if eastwood will let us opt-out of the checks for a particular ns or function, the we could segregate the few bits that are affected.  (Note that I suspect the complaints may vary depending on the JDK version and/or use of `+junixsocket` since that changes the available classes.)

Also, fwiw, this particular PR is just an adjustment to the docs, though I'm happy to help fix the broader issue if we can figure out how we want to.

vemv:
The relevant discussion is https://github.com/nrepl/nrepl/pull/204#discussion_r675346773 , ideally @bbatsov would chime in

In any case reflection warnings, like any other warning, can be silenced.

See https://github.com/nrepl/nrepl/blob/f7d489f4027671b78713a033b97aae84b57d5858/project.clj#L86-L87 

and https://github.com/jonase/eastwood#ignored-faults

rlbdv:
Ahh, ok, thanks.  I'll take a look.

rlbdv:
@bbatsov I've pushed an update with one potential solution (see the two additional commits).  I'm also happy to break that out into a separate pr if that'd be preferable.  We could also use eastwood's "line number" based suppression instead, but I wasn't sure you'd want to have to update the suppression every time the line number changes (FWIW I've also suggested adding function name suppressions to eastwood https://github.com/jonase/eastwood/issues/415 which would avoid the need for the separate namespace.)

And if y'all would prefer the approach using dynamic annotations via macroexpansion, that's fine too.  Personally, I still lean toward avoiding the complexity/risk there on balance, but happy to change this to work that way if you like.

vemv (approved):
LGTM

I'd humbly suggest, please remember about https://github.com/nrepl/nrepl/pull/204#discussion_r682477225 - it's not good to have code that isn't exercised in CI in any way.

Sibling projects like Orchard have had problems in the past due to insufficient test coverage. For dense JDK stuff it's better to have the author in question provide enough assurance in advance

rlbdv:
I removed the CI clj-kondo install (which should be obsolete now -- see commit message), and added junixsocket testing (cf. https://github.com/nrepl/nrepl/pull/204#discussion_r684800064).  For discussion, I just added it to `make test`, which means that it'll run for every clj/jdk combination, which will approximately double the CI run times, and which might not be deemed necessary.

Not sure, but happy to adjust that as you like.  Also happy to break some of all of this out into separate prs if that'd be preferred.

bbatsov:
Looks good to me as well. Thanks!

PR #243, opened 2021-08-31
Sideloader state in session

plexus:
Pending resources in the sideloader are kept in an atom, which currently is
defined in a closure which closes over the classloader and sideloader ops. When
the middleware stack is rebuilt (e.g. via an `add-middleware` op), this atom
gets re-initialized, and the state is lost, causing the side loader to respond
to previously requested resources with `:unexpected-provide`.

The only state that does persist is the session. We already use the session to
store the classloader itself, it makes sense to also keep the `pending` atom in
the session, so that the pending state that we see always corresponds with the
current session classloader.

This fixes an issue which was pointed out in the PR that adds dynamic middleware
loading: https://github.com/nrepl/nrepl/pull/185#issuecomment-624317912

This also makes the base64-decode implementation more conservative (see also https://github.com/clojure-emacs/cider/pull/3038)

Part of https://github.com/clojure-emacs/cider/issues/3037

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   is version 16 or newer or
   [junixsocket](https://kohlschutter.github.io/junixsocket/) is
   available as a dependency.
+* [#243](https://github.com/nrepl/nrepl/pull/243): Keep the sideloader state in the session so it persists across middleware changes. Sanitize the input in base64-decode.
 
 ### Bugs fixed
 
--- a/src/clojure/nrepl/middleware/sideloader.clj
+++ b/src/clojure/nrepl/middleware/sideloader.clj
@@ -81,15 +81,17 @@
 (defn wrap-sideloader
   "Middleware that enables the client to serve resources and classes to the server."
   [h]
-  (let [pending (atom {})]
-    (fn [{:keys [op type name content transport session] :as msg}]
-      (case op
-        "sideloader-start"
-        (alter-meta! session assoc :classloader
-                     (sideloader msg pending))
+  (fn [{:keys [op type name content transport session] :as msg}]
+    (case op
+      "sideloader-start"
+      (let [pending (atom {})]
+        (alter-meta! session assoc
+                     :classloader (sideloader msg pending)
+                     ::pending pending))
 
-        "sideloader-provide"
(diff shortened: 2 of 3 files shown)


bbatsov:
You're on fire, Arne! 

The changes look great to me and I'll merge them later today, unless @shen-tian has some feedback on them. I'll also cut you a new beta of nREPL 0.9.

bbatsov (approved):

plexus:
Just turning my poking around into PRs :)

The failure seems unrelated: nrepl.cmdline-test/basic-fs-socket-behavior - nrepl.cmdline-test

A question that I had here, why use the session atom directly in some places, and store stuff in its middleware in other places? In this case when I try to use the atom directly I actually get errors so I'm guessing there's a good reason, but I don't fully understand it.

Another consideration: should the sideloader cache resources? Since it seems a `require` triggers two classloader resource lookups this immediately doubles the bandwith use... on the other hand you may want to actually repeatedly load a namespace that you are changing locally so there's an argument not to cache.

bbatsov:
> The failure seems unrelated: nrepl.cmdline-test/basic-fs-socket-behavior - nrepl.cmdline-test

Yeah, it's from the socket connection support that we recently added. Guess this test is somewhat flaky. //cc @rlbdv 

> A question that I had here, why use the session atom directly in some places, and store stuff in its middleware in other places? In this case when I try to use the atom directly I actually get errors so I'm guessing there's a good reason, but I don't fully understand it.

I don't remember this off the top of my head, so I'll have to dig through the code to remember. Likely the different usages are just remnants of different stages of the evolution of the code. The session middleware was massively modified when we decided to store the eval thread in the session a few releases ago. 

> Another consideration: should the sideloader cache resources? Since it seems a require triggers two classloader resource lookups this immediately doubles the bandwith use... on the other hand you may want to actually repeatedly load a namespace that you are changing locally so there's an argument not to cache.

I guess we never considered optimization here. I think caching makes sense in most use-cases, as the expectation is that usually you'd be loading 3rd party code via the sideloader. If you want to reload something you can always restart the sideloader.

bbatsov:
0.9.0-beta2 is on Clojars now.

PR #244, opened 2021-09-01
Log errors that happen when loading dynamic middleware

plexus:
It seems prudent to have these errors at least go *somewhere*, instead of just swallowing them.

Part of https://github.com/clojure-emacs/cider/issues/3037

- [x] All tests are passing


bbatsov:
Totally agree!

PR #246, opened 2021-09-09
Sideloader+dynload dogfooding

plexus:
With the work that's been done it's becoming possible to use the sideloader+dynamic loading of middleware to upgrade an nREPL connection to use the cider-nrepl middleware. Now I'm starting to "dogfood" this, i.e. use it to develop on real world projects. This PR is an assortment of changes that I'm doing to try to get to a point where it's actually usable.

- Make sure any remaining dynamic requires (e.g. in the print middleware) use the session classloader
- cache classloading results, so we don't endlessly (re-)fetch the same resources
- add prefix-based filtering, to prevent lots of no-op network round trips.

Draft and open for discussion, I'll probably split this into smaller PRs that will then be accompanied with tests and changelog updates etc.

--------

- Part of: https://github.com/clojure-emacs/cider/issues/3037
- Related: https://github.com/clojure-emacs/cider/issues/3049

-------

- [ ] All tests are passing

### DIFF
--- a/src/clojure/nrepl/middleware/print.clj
+++ b/src/clojure/nrepl/middleware/print.clj
@@ -177,8 +177,7 @@
 (defn- resolve-print
   [{:keys [::print transport session] :as msg}]
   (when-let [var-sym (some-> print (symbol))]
-    (let [print-var (misc/with-session-classloader session
-                      (misc/requiring-resolve var-sym :log))]
+    (let [print-var (misc/session-resolve session var-sym)]
       (when-not print-var
         (let [resp {:status ::error
                     ::error (str "Couldn't resolve var " var-sym)}]
@@ -234,7 +233,7 @@
                  (update ::stream? #(if (= [] %) false (boolean %))))]
       (handler (assoc msg :transport (printing-transport msg opts))))))
 
-(set-descriptor! #'wrap-print {:requires #{}
+(set-descriptor! #'wrap-print {:requires #{"clone"}
                                :expects #{}
                                :handles {}})
 
--- a/src/clojure/nrepl/middleware/print.clj
+++ b/src/clojure/nrepl/middleware/print.clj
@@ -175,9 +175,10 @@
       this)))
 
 (defn- resolve-print
-  [{:keys [::print transport] :as msg}]
+  [{:keys [::print transport session] :as msg}]
   (when-let [var-sym (some-> print (symbol))]
-    (let [print-var (misc/requiring-resolve var-sym)]
+    (let [print-var (misc/with-session-classloader session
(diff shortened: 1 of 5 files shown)


bbatsov:
Another bunch of great suggestions! It's hard for me to write detailed feedback from my phone, so I'll just mention a related idea - an explicit "sideload" op to limit the impact of sideloading in regular evaluation. This may simplify things for clients that would have to start and stop the sideloader manually to get a similar result. In general I always assumed that the clients would do only "targeted" sideloading, so keeping the sideloader on is probably an overkill in most instances. Anyways, that's just an idea for the future. Smaller incremental improvements are fine, and the combination of caching and filtering solves the bulk of the problem.

src/clojure/nrepl/middleware/print.clj
```diff
@@ -175,9 +175,10 @@
       this)))
 
 (defn- resolve-print
-  [{:keys [::print transport] :as msg}]
+  [{:keys [::print transport session] :as msg}]
   (when-let [var-sym (some-> print (symbol))]
-    (let [print-var (misc/requiring-resolve var-sym)]
+    (let [print-var (misc/with-session-classloader session
```
bbatsov: I'd add here your commit message as an inline comment.

bbatsov:
Now that I'm a computer and I could actually read the diff more carefully I like the PR even more. :-) 

I've noticed a small typo in the commit message - `missesi`. The change of the op params should also be reflected in the middleware descriptor, but I guess you planned to do this at the end.

bbatsov:
> Another bunch of great suggestions! It's hard for me to write detailed feedback from my phone, so I'll just mention a related idea - an explicit "sideload" op to limit the impact of sideloading in regular evaluation.

After thinking about this a bit more, it can't really be named `sideload`, as the sideloading is supposed to be transparent, not explicit. Perhaps `inject` or something along those lines. I'm assuming in spirit it'd be pretty close to what you mentioned on another ticket - you load a library explicitly by using `eval` + `lambaisland.classpath`. In the end that's just a fancy `load` + classpath modification.

bbatsov:
@shen-tian @PEZ @liquidz @vemv  Input from you would be most welcome as well!

vemv:
It LGTM, not very familiar with this stuff but I checked for things not to be obstrusive, for people not opting in to use classloading stuff.

Looks like that's the case 👍

shen-tian:
👋🏽 

Just catching up a bit on what's happening, and this is very exciting. I'll need to get myself up to speed a bit first, but should then be able to at least provide some input/dig up reasoning for "why does this do that".

bbatsov:
We've decided to drop the sideloader, so this PR can be closed.

PR #248, opened 2021-09-21
Fewer new loaders

plexus:
Add a few checks to be more conservative about setting the context classloader. This also prevents us from overwriting/shadowing classloaders set up by other tools.

In middleware.session this only creates and sets up a new DynamicClassLoader when one isn't there yet. In misc/with-session-classloader it only set/unsets the current loader if a session classloader is present and it differs from the context classloader.

----

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * [#227](https://github.com/nrepl/nrepl/pull/227): Fix completion for static class members.
 * [#231](https://github.com/nrepl/nrepl/issues/231): Fix sanitize error when file is `java.net.URL`.
 * [#208](https://github.com/nrepl/nrepl/issues/208): Fix namespace resolution in the cmdline REPL.
+* [#248](https://github.com/nrepl/nrepl/pull/248): Create fewer new classloaders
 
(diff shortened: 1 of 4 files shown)


bbatsov (approved):
The proposed changes look solid to me. Great work!

plexus:
Looks like some of the CI builds are timing out... is that a known issue?

bbatsov:
Yeah, this started to happen after adding support for Unix sockets.

plexus:
Ten minutes isn't that much though... maybe it helps to split the tests across multiple tasks.

plexus:
That slowest test is spending 20 seconds because it waits for 10 seconds for an ack, twice.

```
(ack/wait-for-ack 10000)

[(:timeout
.:timeout
.)]
```

bbatsov:
I think this test was brittle with a smaller timeout, but my memory is fuzzy on that matter. ack is an obscure functionality that has been used only by Leiningen (it notifies one nREPL server that another nREPL server has started). Perhaps we can exclude some slower tests via a selector from the CI build to reduce the flakiness there?

PR #254, opened 2021-10-29
Improved reliability of TTY cmdline tests

shen-tian:
- TTY is not a fully fledged transport, and significantly, does not handle ack.
- `cmdline-test` uses a test fixture that loops between all the transports (typically bencode + edn). This doesn't make sense with TTY tests 

I think these have been a contributing factor to `tty-server` being flaky. Moving the tty tests into its own ns seems to have helped, with no error over about ~5 runs on CI in my branch.

I've also added an extra delay into the socket test, which also seems to have helped with consistency.

Most remaining flaky tests have to do with sideloader tests stalling and eventually timing out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ target
 pom.xml
 build.xml
 /test-results
+/.clj-kondo/
+/.lsp/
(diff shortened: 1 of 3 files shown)


bbatsov:
Okay, I'll merge this and we can update the TTY docs and the sideloader tests separately, as they are orthogonal to the goals of the current PR. Thanks for tackling this!

PR #259, opened 2021-11-22
BufferedOutputChannel/write: fix SocketChannel/write call

rlbdv:
The offset and length (.write channel ...) variant is a gathering
operation, and so expects a vector of ByteBuffers.  Switch to the
non-gathering variant, and handle the offset and length in the
ByteBuffer wrapper.

Without the fix, cider connections would cause exceptions like this:
```
  Exception updating the ns-cache #error {
   :cause class java.nio.HeapByteBuffer cannot be cast to class [Ljava.nio.ByteBuffer; (java.nio.HeapByteBuffer and [Ljava.nio.ByteBuffer; are in module java.base of loader 'bootstrap')
   :via
   [{:type java.lang.ClassCastException
     :message class java.nio.HeapByteBuffer cannot be cast to class [Ljava.nio.ByteBuffer; (java.nio.HeapByteBuffer and [Ljava.nio.ByteBuffer; are in module java.base of loader 'bootstrap')
     :at [nrepl.socket.BufferedOutputChannel write socket.clj 207]}]
   :trace
   [[nrepl.socket.BufferedOutputChannel write socket.clj 207]
    [nrepl.socket.BufferedOutputChannel write socket.clj 202]
    [nrepl.transport$safe_write_bencode invokeStatic transport.clj 112]
  ...
```

Thanks to Herwig Hochleitner for reporting the problem.

### DIFF
--- a/src/clojure/nrepl/socket.clj
+++ b/src/clojure/nrepl/socket.clj
@@ -204,7 +204,7 @@
     (if (> length (.capacity buffer))
       (do
         (.flush this)
-        (.write channel (ByteBuffer/wrap byte-array) offset length))
+        (.write channel (ByteBuffer/wrap byte-array offset length)))
       (do
         (when (> length (.remaining buffer))
           (.flush this))


rlbdv:
https://github.com/clojure-emacs/cider-nrepl/issues/727

bbatsov:
Is there some regression test we can add for this?

rlbdv:
One option might be to add a test alongside `basic-socket-fs-behavior` that tests a round trip of a message larger than [8192 bytes](https://github.com/nrepl/nrepl/blob/dfecc450ca936dd5716ba144f4d262d53b33be04/src/clojure/nrepl/socket.clj#L233).  I suppose that would be fairly expensive/heavyweight, for the very specific case it's testing, but I think it would work, and it's fairly insensitive to code changes as long as we move the 8192 to a var they can share.

I think the reason cider hit this is because it appears to send "all" the docstrings (or similar) at startup.

Another option would be a narrower `nrepl.socket` level test, though that might be a bit more complex to set up, i.e. we'd need to manually create the fs socket, etc., or we might be able to use a localhost `SocketChannel` via `InetSocketAddress`, and then we'll need to find an open port, which will have no security (presumably already the case for other tests).

It might not take very long to create a reasonable quality heavyweight test -- not sure about the narrower test -- likely not bad, but unless there's an existing similar test we can easily adjust, if creating it takes much time, it may be a bit before I come up with one, assuming no one else pursues it.

bbatsov:
Got it. Well, I guess I'll just merge this is as then and I'll cut another (hopefully final) 0.9 beta release.

rlbdv:
OK, I suppose this might also suggest we might want to eventually add some additional top-level tests across all the available transports, i.e. at least one "large" message, or some randomized messages, including large ones (not sure if that's worthwhile here), and so on.

bbatsov:
Yeah, probably we should create a separate issues about better transport tests.

PR #264, opened 2022-01-17
Upgrade linters

vemv:


project.clj
```diff
@@ -80,8 +80,7 @@
                                          testing-print [[:inner 0]]}}}
 
              :eastwood [:test
-                        {:plugins [[jonase/eastwood "0.9.9"]]
-                         :global-vars {*warn-on-reflection* true}
```
vemv, line 84: As of lately Eastwood doesn't care about the global `*warn-on-reflection*` value, it sets it to true anyway

Issue #263, opened 2022-01-04
Avoid illegal reflective access


vemv:
alexmiller wrote on Clojurians Slack:

> yeah, this kind of code is not valid in 17 without setting additional jvm flags 

https://github.com/nrepl/nrepl/blob/8223894f6c46a2afd71398517d9b8fe91cdf715d/src/clojure/nrepl/middleware/interruptible_eval.clj#L32-L40 

Which relates to actual issues that users have experienced. I too think I have experienced an issue directly related to LineNumberingPushbackReader hacking, which is undone somewhere in my personal setup.

bbatsov:
How can you undo this? Without it there would be no location metadata associated with `eval`, which is massive regression.

frenchy64:
Perhaps use tools.reader's [SourceLoggingPushbackReader](https://github.com/clojure/tools.reader/blob/6918abc8b3c009228790cf091097eb3037858ad6/src/main/clojure/clojure/tools/reader/reader_types.clj#L263) and [read](https://github.com/clojure/tools.reader/blob/6918abc8b3c009228790cf091097eb3037858ad6/src/main/clojure/clojure/tools/reader.clj#L961) instead?

vemv:
Sounds nice to me, thanks!

Can nrepl include 3rd party dependencies? @bbatsov

vemv:
With a bit of luck we can inline only the relevant code (by hand, copy-paste, without an automated ns prefix) which might mean only one extra ns or such?

bbatsov:
That'd be fine by me.

alexander-yakushev:
I'm a bit confused because this particular code doesn't transgress JDK reflection limitations.

```clj
$ clj -Srepro                                                                                                                                            ~
Clojure 1.11.3
user=> (System/getProperty "java.version")
"22"
user=> (let [reader (clojure.lang.LineNumberingPushbackReader. (java.io.StringReader. "(+ 1 2)"))
      field (doto (.getDeclaredField clojure.lang.LineNumberingPushbackReader "_columnNumber")
              (.setAccessible true))]
  (.set field reader (int 42))
  (.getColumnNumber reader))
42
```

This is because `clojure.lang.LineNumberingPushbackReader` lives in the "unnamed" module, same as nREPL.

bbatsov:
@alexander-yakushev Is there any action you'd recommend? I'm concerned that once this reflection breaks the impact will be quite severe, as code navigation will stop working.

[...]

PR #265, opened 2022-01-17
Guard against `InaccessibleObjectException`s

vemv:
Fixes https://github.com/nrepl/nrepl/issues/263

### DIFF
--- a/src/clojure/nrepl/middleware/interruptible_eval.clj
+++ b/src/clojure/nrepl/middleware/interruptible_eval.clj
@@ -39,11 +39,15 @@
         (doto (.setAccessible true))
         (.set reader column))))
 
-(defn- source-logging-pushback-reader
+(defn source-logging-pushback-reader
   [code line column]
   (let [reader (LineNumberingPushbackReader. (StringReader. code))]
-    (when line (set-line! reader (int line)))
-    (when column (set-column! reader (int column)))
+    (try
+      (when line (set-line! reader (int line)))
+      (when column (set-column! reader (int column)))
+      (catch Exception e
+        (when-not (misc/inaccessible-object-exception? e)
+          (throw e))))
     reader))
 
 (defn- interrupted?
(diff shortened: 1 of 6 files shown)


vemv:
mmm InaccessibleObjectExceptions doesn't exist under JDK8, will have to tweak.

vemv:
PR ready. Unfortunately I never got to reproduce the issue in question.

It's weird: our CI has JDK17 and I know from various sources that nrepl's reflective code throws a `InaccessibleObjectException`.

Maybe it depends on how the JDK itself was built, idk.

bbatsov:
I'll have to think more about this, as I'm quite puzzled by it. We're guarding towards something that doesn't happen in our own CI? I also don't quite get what the fallback achieves other than catching the exception. I mean you still the need the reflection to achieve what the code needs to do.

src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -39,11 +39,15 @@
         (doto (.setAccessible true))
         (.set reader column))))
 
-(defn- source-logging-pushback-reader
+(defn source-logging-pushback-reader
```
bbatsov, line 42: My make this public? It is just to test it?

src/clojure/nrepl/middleware/interruptible_eval.clj
```diff
@@ -39,11 +39,15 @@
...
   [code line column]
   (let [reader (LineNumberingPushbackReader. (StringReader. code))]
-    (when line (set-line! reader (int line)))
-    (when column (set-column! reader (int column)))
+    (try
```
bbatsov, line 45: This and the other similar instance should have some comments explaining the need for the try/catch, as it's definitely not obvious.

alexander-yakushev:
This can be shelved until we have to access something from `java.base` or any other module through reflection. #263 is already closed as Fixed.

PR #278, opened 2022-05-26
Fix reflection warning

### LINKED ISSUE #268: [EASY] Reflection Warning (with solution)
## Expected behavior
When you set `*warn-on-reflection*` to `true` you would expect to don't have any warnings because they decrease performance

## Actual behavior
I'm getting a warning for this project on this line https://github.com/nrepl/nrepl/blob/011f5db319830f105f943ef44e69968c82fc531c/src/clojure/nrepl/cmdline.clj#L431

Just add this at the beginning of the file `(set! *warn-on-reflection* true)` and load the namespace on repl

### Clojure version
1.10.3

### Java version
11.0.13

### Operating system
any

# SOLUTION
the function of [line](https://github.com/nrepl/nrepl/blob/011f5db319830f105f943ef44e69968c82fc531c/src/clojure/nrepl/cmdline.clj#L422) should be like this:
```clojure
(defn save-port-file
  "Writes a file relative to project classpath with port number so other tools
  can infer the nREPL server port.
  Takes nREPL server map and processed CLI options map.
  Returns nil."
  [server options]
  ;; Many clients look for this file to infer the port to connect to
  (let [port (:port server)
        port-file (io/file ".nrepl-port")]
    (.deleteOnExit ^File port-file)
    (spit port-file port)))
```

solar05:
Fixes #268

### DIFF
--- a/src/clojure/nrepl/cmdline.clj
+++ b/src/clojure/nrepl/cmdline.clj
@@ -454,7 +454,7 @@ Exit:      Control+D or (exit) or (quit)"
   ;; Many clients look for this file to infer the port to connect to
   (let [port (:port server)
         port-file (io/file ".nrepl-port")]
-    (.deleteOnExit port-file)
+    (.deleteOnExit ^java.io.File port-file)
     (spit port-file port)))
 
 (defn start-server


bbatsov:
Thanks!

Issue #272, opened 2022-05-02
nrepl transports should `read` with `{:read-cond :allow}` options


dpsutton:
Had an issue come up in inf-clojure. We ask for arglists with the following form:

```clojure
(try
  (:arglists
   (clojure.core/meta
     (clojure.core/resolve
       (clojure.core/read-string \"%s\"))))
  (catch #?(:clj Throwable :cljr Exception) e nil))
```

Someone is using nrepl's transport/tty on a Clojure repl and it is blowing up on the reader conditional.

The reader conditional exists because cljr is on .NET and has no notion of `Throwable`.  There are a few ways to fix this in inf-clojure:
- just collapse back to catching `Exception` for both branches and thus remove the need for conditional reading
- fully disentangle the new branches and make a new interaction layer for cljr vs "regular" clj on the jvm
- have nrepl allow reader conditionals.

The last choice would be the most work, but I think it is the most correct. Prior art for why this is correct:
- clojure's socket repl allows for reader conditionals by default: https://github.com/clojure/clojure/blob/master/src/clj/clojure/core/server.clj#L174
- clojure's main repl allows reader conditionals as well: https://github.com/clojure/clojure/blob/master/src/clj/clojure/main.clj#L168

That is a strong argument that arbitrary repls aught to default to allowing reader conditionals in general.

The changes required should be localized to all of the reads in transport.clj:
- the tty's read: https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/transport.clj#L179

And it appears nrepl's default bencode transport allows for reader conditionals:

```clojure
❯ lein repl
nREPL server started on port 61122 on host 127.0.0.1 - nrepl://127.0.0.1:61122
REPL-y 0.5.1, nREPL 0.8.3
Clojure 1.10.3
OpenJDK 64-Bit Server VM 17.0.1+12-LTS
    Docs: (doc function-name-here)
          (find-doc "part-of-name-here")
  Source: (source function-name-here)
 Javadoc: (javadoc java-object-or-class-here)
    Exit: Control+D or (exit) or (quit)
 Results: Stored in vars *1, *2, *3, an exception in *e

user=> (+ 1 #?(:clj 2))
3
user=>
```

bbatsov:
@dpsutton I've got no objections to moving in the direction you've suggested, so a PR would be welcome.

JasonKDarby:
@bbatsov @dpsutton, see ^

PR #279, opened 2022-05-28
:read-cond :allow for reading during tty transport

JasonKDarby:
Fixes https://github.com/nrepl/nrepl/issues/272

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/src/clojure/nrepl/transport.clj
+++ b/src/clojure/nrepl/transport.clj
@@ -164,14 +164,6 @@
             (.close in)
             (.close out))))))))
 
-(defn tty-read-msg
-  "Provides a fn that invokes read on r, the provided java.io.PushbackReader.  The current
-  namespace, cns, and session id (provided as atoms) are used with the read result to return an eval op."
-  [r cns session-id]
-  #(let [code (read {:read-cond :allow} r)]
-     (merge {:op "eval" :code [code] :ns @cns :id (str "eval" (uuid))}
-            (when @session-id {:session @session-id}))))
-
 (defn tty
   "Returns a Transport implementation suitable for serving an nREPL backend
    via simple in/out readers, as with a tty or telnet connection."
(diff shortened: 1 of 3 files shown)


src/clojure/nrepl/transport.clj
```diff
@@ -164,6 +164,14 @@
             (.close in)
             (.close out))))))))
 
+(defn tty-read-msg
```
JasonKDarby, line 167: I really don't like returning a function from this, if you have suggestions I'd appreciate it.

bbatsov: You can always keep it as a lambda and do some integration test of the transport instead. I guess the only reason you made this a public function is to be able to write that unit test for it.

JasonKDarby: Yeah, I did it so I could demonstrate the change was very focused.  I'll put it back.

test/clojure/nrepl/transport_test.clj
```diff
@@ -9,3 +9,15 @@
...
+
+(deftest tty-read-conditional-test
+  (testing "tty-read-msg is configured to read conditionals"
+    (let [in (-> "(try nil (catch #?(:clj Throwable :cljr Exception) e nil))"
+                 (java.io.StringReader.)
+                 (java.io.PushbackReader.))
+          out (ByteArrayOutputStream.)]
+      (is (= ['(try nil (catch Throwable e nil))]
+             (let [FnTransport (sut/tty in out nil)]
+               (.recv FnTransport)     ;; :op "clone"
```
JasonKDarby, line 21: @bbatsov this wasn't as bad as I originally thought.
I saw eastwood fails on `.recv` saying "reference to field recv can't be resolved".  Any tip on how to resolve this?

bbatsov: Probably you need to add a type signature there. Btw, I think the variable should be named `fn-transport`, not `FnTransport` as this looks like some type name. //cc @vemv

JasonKDarby: There's a type `FnTransport` and also a defn `fn-transport` in nrepl.transport.  There must be a better name but I just named it `fn-transport` as suggested.

bbatsov:
Thanks for tackling this!

PR #281, opened 2022-07-20
Make junixdomain integration forward compatible with 2.5.x

DerGuteMoritz:
Starting from version 2.5.0, the constructor of
`org.newsclub.net.unix.AFUNIXSocketAddress` which was used by
`junix-address-of` is private. However, since 2.4.0 an equivalent `of`
factory method is available. This patch makes the code compatible with
both generations of the library.

- [X] You've added tests to cover your change(s): I bumped the `junixsocket` dependency so that the tests now run against that latest version. Alternatively, we could have one test run with the old version and one with the new instead. Let me know if you'd prefer that.
- [X] All tests are passing
- [X] You've updated the [changelog](../CHANGELOG.md)(that only applies to user-visible changes)

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#279](https://github.com/nrepl/nrepl/pull/279): Allow reader conditionals for tty transport.
 * [#217](https://github.com/nrepl/nrepl/pull/270): Add nREPL client support for unix domain sockets.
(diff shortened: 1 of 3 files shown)


bbatsov:
Eastwood is not happy about the use of reflection. Otherwise your changes look good to me.

DerGuteMoritz:
> Eastwood is not happy about the use of reflection. Otherwise your changes look good to me.

Ah, sorry, didn't run that one. Looks like it has some trouble with following the conditional in my original patch. I restructured the code to appease it :smile:

bbatsov:
Thanks!

PR #283, opened 2022-09-01
Support TLS sockets

ivarref:
Please see issue: https://github.com/nrepl/nrepl/issues/266

`make check` passes.

Thanks!

Before submitting a PR make sure the following things have been done:

- [X] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [X] You've added tests to cover your change(s)
- [X] All tests are passing
- [X] The new code is not generating reflection warnings
- [X] You've updated the [changelog](../CHANGELOG.md)(that only applies to user-visible changes)

Thanks!

*If you're just starting out to hack on nREPL you might find this [section of its
manual][1] extremely useful.*

[1]: https:/nrepl.org/nrepl/hacking_on_nrepl.html

PR #287, opened 2023-01-10
Fix dropped accept-connection promise recursion

desa:
When TLS was introduced in https://github.com/nrepl/nrepl/pull/283 the ability to consume an exception was introduced so that we could catch SSL related exceptions. This introduced an issue where the first failed SSL connection would be handled, but further connections would no longer be accepted.

This PR introduces a recursive function accept-connections in start-server so that the first caught SSL connection does not prevent future good connections from being handled.

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -197,17 +197,23 @@
                         (atom #{})
                         transport-fn
                         greeting-fn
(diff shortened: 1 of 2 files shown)


desa:
Thought I had this working, but seems like the tests are failing now. Closing until I have something that's working.

PR #288, opened 2023-01-10
Fix dropped accept-connection promise recursion

desa:
When TLS was introduced in https://github.com/nrepl/nrepl/pull/283 the ability to consume an exception was introduced so that we could catch SSL related exceptions. This introduced an issue where the first failed SSL connection would be handled, but further connections would no longer be accepted.

This PR introduces a recursive function accept-connections in start-server so that the first caught SSL connection does not prevent future good connections from being handled.

- [ ] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -197,17 +197,23 @@
                         (atom #{})
                         transport-fn
                         greeting-fn
-                        (or handler (default-handler)))]
-    (noisy-future
-     (try
-       (accept-connection server)
-       (catch Throwable t
-         (cond consume-exception
-               (consume-exception t)
-               (instance? SocketException t)
-               nil
-               :else
-               (throw t)))))
+                        (or handler (default-handler)))
+        accept-connections (fn accept-connections []
+                             (noisy-future
+                              (try
+                                (accept-connection server)
+                                (catch Throwable t
+                                  (cond consume-exception
+                                        (do (consume-exception t)
+                                            (accept-connections))
+                                        (instance? SocketException t)
+                                        nil
+                                        :else
+                                        (throw t))))))]
+    (try
+      (accept-connections)
+      (catch SocketException _
+        nil))
     (when ack-port
       (ack/send-ack (:port server) ack-port transport-fn))
     server))
--- a/test/clojure/nrepl/tls_test.clj
+++ b/test/clojure/nrepl/tls_test.clj
@@ -70,6 +70,37 @@
                            :value)))
           (is (instance? SSLHandshakeException @exception)))))))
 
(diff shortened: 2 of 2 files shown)


src/clojure/nrepl/server.clj
```diff
@@ -197,17 +197,23 @@
...
-       (accept-connection server)
-       (catch Throwable t
-         (cond consume-exception
-               (consume-exception t)
-               (instance? SocketException t)
-               nil
-               :else
-               (throw t)))))
+                        (or handler (default-handler)))
+        accept-connections (fn accept-connections []
```
bbatsov, line 201: The indentation here seems off.

bbatsov:
@ivarref Please take a look at this PR.

ivarref:
Hi @desa and cc @bbatsov 

Thanks for spotting this bug and for providing a PR!

I have two comments for your PR and one comment for nrepl as it is before this PR:

`PR #1`. This fix will only fix the server if the `consume-exception` parameter is set. The default value of `consume-exception` is nil. Thus by default the server will have this bug. Can you fix this?

`PR #2` The stack will grow when `accept-connections` calls itself. Can you fix this so that the stack won't grow?

`Nrepl #1` I find it odd that `accept-connection` creates two new threads, one for handling the connection and one for accepting more connections, upon receiving a new connection. Would it be better to just create one, and "re-use" the current thread for accepting more connections?

Thanks and kind regards.

bbatsov:
The PR feedback makes sense to me. @desa Can you please address them both?

> Nrepl #1 I find it odd that accept-connection creates two new threads, one for handling the connection and one for accepting more connections, upon receiving a new connection. Would it be better to just create one, and "re-use" the current thread for accepting more connections?

I agree. Probably we should file a separate ticket for this, given that it's somewhat orthogonal to the PR and we'll need to carefully check the reasoning before the current design with two threads - might be just an oversight, but it might have been done like this for a reason. Sadly, I don't remember the details myself.

bbatsov:
@desa ping :-)

bbatsov:
@desa ping 2 :-)

bbatsov:
@ivarref Any chance you take over this PR and drive it to completion? I'd like to finally cut a new nREPL release and it'd be nice to have a fix for the bug there.

ivarref:
Hi @bbatsov Yes, I will look into this.

ivarref:
Sorry I've been sick lately, but I will look into it, hopefully during next week or so.

bbatsov:
@ivarref Thanks for the update! Get well soon!

bbatsov:
Superseded by #303.

PR #289, opened 2023-01-16
Extend timeouts in cmdline tests to accommodate slower hosts

rlbdv:
The tests were failing on the Debian arm64 buildd's, and it turns out that's because those machines take longer to start the subprocesses that the tests need.  To allow the tests to succeed, increase the wait times in cmdline-test and cmdline-tty-test, and rework the tty-server test to be able to detect when the server is available instead of just sleeping unconditionally.  That should also allow the test to run faster on faster machines.

Thanks to Louis-Philippe Véronneau for reporting the problem.

### DIFF
--- a/test/clojure/nrepl/cmdline_test.clj
+++ b/test/clojure/nrepl/cmdline_test.clj
@@ -152,7 +152,7 @@
                                   "nrepl.main"
                                   "--ack" (str ack-port)
                                   "--transport" (var->str *transport-fn*)])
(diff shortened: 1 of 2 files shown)


rlbdv:
Here's an [example of the original failures](https://buildd.debian.org/status/fetch.php?pkg=nrepl-clojure&arch=arm64&ver=1.0.0-3&stamp=1673843158&raw=0).

bbatsov:
The proposed solution looks good to me. Thanks!

Issue #291, opened 2023-02-12
test suite fails on MS-Windows 


ikappaki:
## Expected behavior

All tests should pass when running on MS-Windows

## Actual behavior

Some tests fail, e.g. first failure

```
lein test :only nrepl.cmdline-test/ack-server

FAIL in (ack-server) (cmdline_test.clj:130)
bencode transport
expected: (= "ack'ing my port 6000 to other server running on port 8000\n" output)
  actual: (not (= "ack'ing my port 6000 to other server running on port 8000\n" "ack'ing my port 6000 to other server running on port 8000\r\n"))

lein test :only nrepl.cmdline-test/ack-server

FAIL in (ack-server) (cmdline_test.clj:130)
edn transport
expected: (= "ack'ing my port 6000 to other server running on port 8000\n" output)
  actual: (not (= "ack'ing my port 6000 to other server running on port 8000\n" "ack'ing my port 6000 to other server running on port 8000\r\n"))
```

and then the test suit hangs and needs to be killed:

```
...
nREPL server started on port 53495 on host 127.0.0.1 - nrepl://127.0.0.1:53495
```
## Steps to reproduce the problem

Run `lein test`.

There are a few failures then the test suit hangs.

### Clojure version

1.11.1

### Java version

17

### Operating system

MS-Windows 

---

I can have a look and also add a corresponding job to circleci.

Thanks

bbatsov:
I think this can be closed now.

PR #292, opened 2023-02-26
circleci win and clj 1.11 support, fix completion issues

ikappaki:
Hi,

could you please review patch to add circleci test support for MS-Windows. It addresses #291.

I've also took the opportunity to move Clojure version to 1.11 in `project.clj` and include that version in ci tests.

Changes are
1. Introduce functions in `test/nrepl/test_helpers.clj` to handle difference in newlines (`\n` vs `\r\n` on windows) and file separators (`/` vs `\` on windows) in test results and use them in failing tests as such.
2. Skipped tests that had to do with unix sockets on windows.
3. Placed `sh` std in/out/err streams in a future because one (all?) are blocking calls on windows (I haven't really looked into the details, but I took it, in my experience, as a legit thing to do).
4. Wrapped the `:transport-fn` in `nrepl.tls-test/bad-keys` in a try catch block because there appears to be a (race?) condition in the nrepl mechanism that the transport layer could try to send data to the a client even though the initial connection has thrown an exception. I haven't looked into the details why that is, but I thought this is a straightforward workaround to catch the exception. @ivarref , could you please assess/review this change in `tls_test.clj` please? There is also another intermittent unix exception seen from bad-keys test as per https://gist.github.com/ikappaki/6958e8d7689eb916a887d78be93b0b87 in this PR that I need your input please. It seems that `nrepl/message` can also throw a `javax.net.ssl.SSLHandshakeExceptio` in addition to `SocketException`?
5. Fixed issues with completion middleware 
  a. local class file paths were returned starting with a dot, i.e.  `.<package>.<classname>`, and added a test for it (subtracting the base directory name from the filename left a `/` at the front of the subtraction result).
  b. .jar file completion did not work on windows (files inside a jar were always returned using `/` but the old code was using the system file separator -- `\` on windows -- to convert them into dots).
7. Added circle ci windows test, but for convenience only using the java version found on the system and the latest Clojure version 1.11.
8. While at it, I've used Clojure 1.11 as the default version in `project.clj` and added it to circle ci tests.
9. Prefixed circle ci test names with `U` (for Unix) and `W` (for Windows) so that we know which one it in the web page, and also shorted the name of `Clojure` to `Clj` because it doesn't fit any more in the output.

I'm also thinking of converting the `Makefile` to babashka `bb.edn` so that the commands can be used seamlessly both on Linux and windows (`make` is not a standard command on windows) . What do you think?

Thanks,

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * [#266](https://github.com/nrepl/nrepl/pull/266): Add TLS support.
 
+### Bugs fixed
+
+* [#291](https://github.com/nrepl/nrepl/pull/291): Fix issue with completion middleware not returning values for local class files, or .jar files on windows.
+
 ## 1.0.0 (2022-08-24)
 
 ### New Features
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,9 @@ version: 2.1
 
 # Default settings for executors
 
+orbs:
+  win: circleci/windows@5.0.0
+
 defaults: &defaults
   working_directory: ~/repo
   environment:
@@ -34,8 +37,6 @@ executors:
 # Runs a given set of steps, with some standard pre- and post-
 # steps, including restoring of cache, saving of cache.
 #
-# we also install `make` here.
-#
 # Adapted from https://github.com/lambdaisland/meta/blob/master/circleci/clojure_orb.yml
 
 commands:
@@ -60,9 +61,11 @@ commands:
         default: "1"
     steps:
       - run:
-          name: Install make
+          name: Install babashka latest
           command: |
-            sudo apt-get install make
+            bash <(curl -s https://raw.githubusercontent.com/borkdude/babashka/master/install) --dir ~
+            sudo mv ~/bb /usr/local/bin/bb
+
       - run:
           name: Generate Cache Checksum
           command: |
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 * [#266](https://github.com/nrepl/nrepl/pull/266): Add TLS support.
 
+### Bugs fixed
+
+* [#291](https://github.com/nrepl/nrepl/pull/291): Introduce CI support for MS-Windows. Fixes issue with completion middleware not returning values for local class files, or .jar files on windows.
```
bbatsov, line 11: I think the first sentence is redundant here, as the CI support is not really a bug-fix.

bbatsov:
It will also be great to split this into several commits - e.g. one for each point you've made in the description of the PR.

CHANGELOG.md
ikappaki: Removed.

ikappaki:
> > Prefixed circle ci test names with U (for Unix) and W (for Windows) so that we know which one it in the web page, and also shorted the name of Clojure to Clj because it doesn't fit any more in the output.
> 
> How much characters can we fit there in total? I'm a bit worried the output might be somewhat cryptic to newcomers.

In my web browser, it shows around 25 chars; I've expanded U/W to Lnx/Win, let me know what you think how it looks like in your browser in case is related to the display resolution.
 
> > Skipped tests that had to do with unix sockets on windows.
> 
> Btw, supposedly Windows 10+ supports Unix sockets, but I haven't programmed under Windows in ages and I never tried this. (see https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/)

I will have a look, I get the following error on windows:

`actual: java.lang.UnsupportedOperationException: 'posix:permissions' not supported as initial attribute`

> 
> > I'm also thinking of converting the Makefile to babashka bb.edn so that the commands can be used seamlessly both on Linux and windows (make is not a standard command on windows) . What do you think?
> 
> That would be fine with me. We'll just have to update the hacking documentation and make it clear people will need to install bb.

I've included a `bb.edn` file to replace and added a reference to `babashka` website in the docs (shall I add an entry in the `README.md` to that `babashka` is now a requirement and how to install it?)

ikappaki:
> It will also be great to split this into several commits - e.g. one for each point you've made in the description of the PR.

Split PR into several commits roughly matching the original description points made (please note that `test/java/nrepl/test/Dummy.java` should have been part of the completion fix, but missed it and added it with the last commit).

bbatsov:
> In my web browser, it shows around 25 chars; I've expanded U/W to Lnx/Win, let me know what you think how it looks like in your browser in case is related to the display resolution.

Seems those are always truncated in the CircleCI's UI, but not in the listing of the builds in GitHub. Perhaps we should only prefix the Windows build, but that's something we can easily change down the road.

bbatsov:
Thanks for working on this!

PR #293, opened 2023-03-05
win test support: remove unecessary posix permissioning on test dir

ikappaki:
Hi,

could you please consider patch to enable the remaining unixsocket couple of tests on MS-Windows. This is a follow up to #292 

They were initially disabled because it was thought that there was no unixsocket support on win, but it turned out it was a test fn that was creating a temp directory that was incompatible on win; it was setting the dir permission unix posix attributes. It is not necessary to set temp dir permissions for the tests to run, so I removed them (they are already set to the right rwx permissions).

- [x] All tests are passing

Thanks!


ikappaki:
I've also removed the `Makefile` as it is now replaced with `bb.edn`

bbatsov:
I see there's some test failure on Windows. A couple of more things:

1. Might be good to update the Unix socket support docs to mention that this actually works on Windows.
2. Check there are no references to this Makefile in the docs (e.g. in the hacking on nREPL sections)

ikappaki:
Hi @bbatsov , sorry for the delay in replying

> I see there's some test failure on Windows. A couple of more things:

The win test failure was due to an an assumption that the exception thrown on tls failure was of `Socket` type, which might be not. I've used a `try/catch` block instead.

>     1. Might be good to update the Unix socket support docs to mention that this actually works on Windows.

IMHO this is not necessary since there was nothing in the docs suggesting this was not supported on windows to begin with? 

>     2. Check there are no references to this Makefile in the docs (e.g. in the hacking on nREPL sections)

This was already done, but I've added some more references to `babashka` to make it clearer it's need to run the tasks.

Other than that
1. The `cloverage` lein task did not seem to ever work, it was throwing an `Wrong number of args (3) passed to: clojure.core/complement` exception. I've updated to remove the `test` reference from the lein project. It also uncovered an inconsistency in `transport.clj` that the `throwable` test had to check if the exception is socket related (which is not the case when using unix sockets). Also corrected some `:user/xxx` keywords that appeared under a different namespace in `cmdline_test.clj` when ran with cloverage.
2. Converted the unix specific `bin/kaocha` script to a cross platform `bb kaocha` task and updated the docs.
3. Fixed a couple of profile misspelling in `project.clj` introduced with the previous commit.
4. Fixed a couple of word misspellings already in the codebase reported by `codespell` in the pull request.

The `misspell` CI task seemed to have failed without a reason, and there appears to be an intermittent failure with `Win Java sys, Clj 1.11`. 

I suspect there will be some intermittent failures with win due to the slower windows runtime that will need investigated. I plan to deal with them as they come through, after I introduce nREPL requests logging.

Actually this whole exercise is a prelude to introduce some nREPL logging to deal with the intermittent failures we've seen in CIDER with the win runtime.

Would it be possible to give the CI another kick to at least rerun the `misspell` task?

Thanks

ikappaki:
Further to the previous commit, I've noticed that the `nrepl.cmdline-test/unix-domain-flavor` was indefinitely hanging on CI because it was using the `junixsocket` library with jdk 11. The file `.exists` method is always returning false for a socket file in JDK 11, thus I've introduced a new test fn to check for socket connection to determine when the socket file is created.

bbatsov:
> (notice the there is a random intermittent test failure on win that is not reproducible; I hope to be in a better position to look into these after I introduce server request logging next as I mentioned earlier):

Sounds good.

PR #297, opened 2023-07-03
Fix an oversight

vemv:


bbatsov:
@vemv It seems that something's still broken on Clojure 1.7.

PR #303, opened 2023-10-27
Address issues in PR https://github.com/nrepl/nrepl/pull/288

ivarref:
Hi @bbatsov  and @desa

I believe this change covers the issues I noted in https://github.com/nrepl/nrepl/pull/288#issuecomment-1381654873

I've tried to keep accept connection as close as possible to the original, and no stack growth.

Edit: All tests pass now.

Thanks and kind regards

### DIFF
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -16,7 +16,8 @@
    [nrepl.tls :as tls]
    [nrepl.transport :as t])
   (:import
-   (java.net SocketException)
+   (java.net ServerSocket SocketException)
+   (javax.net.ssl SSLException)
    [java.nio.channels ClosedChannelException]))
 
 (defn handle*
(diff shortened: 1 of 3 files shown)


ivarref:
Let me know if there is need for additional tweaking @bbatsov

bbatsov:
@ivarref Your changes look good to me. I'd suggest squashing the commits together and using a more descriptive message for the final commit (I think it's better to explain what we're fixing directly than to refer to #288), before we can merge this.

src/clojure/nrepl/server.clj
```diff
@@ -51,27 +52,40 @@
...
+                      :recur)
+                    (catch ClosedChannelException cce
+                      (consume-exception cce)
+                      nil)
+                    (catch Throwable t
+                      (consume-exception t)
+                      (if (.isClosed ^ServerSocket server-socket)
+                        nil
+                        (throw t))))]
+    (if (= sock :recur)
```
bbatsov, line 70: Might be a good idea to document with a comment the usage of `recur` here.

ivarref:
Added comment about :tls-exception (renamed from :recur) and squashed now.

Any other feedback @bbatsov ?

Thanks and kind regards.

bbatsov:
Seems one of the new tests has some degree of flakiness on Windows:

```
nrepl.tls-test
  bad-config
  bad-keys
  bad-keys-then-good
Too long with no output (exceeded 10m0s): context deadline exceeded
```

But if it's not easy to fix this we can live with it.

ivarref:
I'll have a go at the flakyness.

Regards.

ivarref:
I've found one problem, though I'm not sure if that was causing the flakiness or not:


Original flaky code:
```clojure
(defn accept
  "Accepts a new TLS connection. Waits 10 000 milliseconds for the TLS handshake
  to complete. Requires that the client certificate is different from the server certificate."
  [^javax.net.ssl.SSLServerSocket server]
  (let [p (promise)
        ^SSLSocket sock (.accept server)]
    (.addHandshakeCompletedListener sock
                                    (reify HandshakeCompletedListener
                                      (handshakeCompleted [_ e]
                                        (if (= (into [] (.getLocalCertificates e))
                                               (into [] (.getPeerCertificates e)))
                                          (deliver p :handshake-bad!)
                                          (deliver p :handshake-ok!)))))
    (future
      (when (= :timeout (deref p 10000 :timeout))
        (deliver p :handshake-bad!)
        (close-silently sock)))
    (try
      (.startHandshake sock)
     ; accept thread is done and blocked here, just after the handshake completes:
     ; now the above future fires and delivers :handshake-bad!
     ; which closes the socket and then returns nil.
     ; That in turn stops the server from accepting new connections
     ; and never delivers any exception. Thus @exception in the tls-test would hang forever.
      (let [v @p]
        (if (= v :handshake-bad!)
          (close-silently sock)
          sock))
      (catch SSLException e
        (close-silently sock)
        (throw e)))))

```

bbatsov:
Thanks for tackling this!

bbatsov:
@ivarref I finally cut nREPL 1.1, so more people can enjoy the TLS support. Thanks for working on this!

ivarref:
Thanks for friendly and timely feedback @bbatsov You do a great job as a maintainer IMO.

Kind regards.

PR #304, opened 2023-11-06
[completion] Use compliment-lite for code completion

alexander-yakushev:
First attempt.

I replaced the existing code with [Compliment-lite](https://github.com/alexander-yakushev/compliment/tree/master/lite) but kept the compatibility with the current call signature of `completions` in nREPL.

### DIFF
--- a/src/clojure/nrepl/util/completion.clj
+++ b/src/clojure/nrepl/util/completion.clj
@@ -1,319 +1,688 @@
 (ns nrepl.util.completion
   "Code completion functionality.
 
-  The functionality here is experimental and
-  the API is subject to changes."
-  {:author "Bozhidar Batsov"
+  This namespace is based on compliment-lite.
+  (https://github.com/alexander-yakushev/compliment/tree/master/lite)
+
+  The functionality here is experimental and the API is subject to changes."
+  {:author "Oleksandr Yakushev, Bozhidar Batsov"
    :added  "0.8"}
-  (:require [clojure.main]
-            [nrepl.misc :as misc])
-  (:import [java.util.jar JarFile]
-           [java.io File]
-           [java.lang.reflect Field Member]
-           [java.util.jar JarEntry]
-           [java.util.concurrent ConcurrentHashMap]))
+  (:import java.io.File
+           java.nio.file.Files
+           (java.util.function Function Predicate)
+           java.util.concurrent.locks.ReentrantLock
+           (java.util.jar JarEntry JarFile)
+           java.util.stream.Collectors
+           (java.lang.reflect Field Member Method Modifier)
+           (java.util Comparator HashSet)))
 
-;; Code adapted from Compliment (https://github.com/alexander-yakushev/compliment)
+;; Compliment lite was generated at Tue Nov 14 01:00:10 EET 2023
+;; SPDX-License-Identifier: EPL-1.0
 
-(defn annotate-keyword
-  [kw]
-  {:candidate kw :type :keyword})
+;; compliment/utils.clj
 
-(defn all-keywords
-  []
-  (let [^Field field (.getDeclaredField clojure.lang.Keyword "table")]
-    (.setAccessible field true)
-    (.keySet ^ConcurrentHashMap (.get field nil))))
+(def ^{:dynamic true} *extra-metadata*
+  "Signals to downstream sources which additional information about completion\n  candidates they should attach . Should be a set of keywords."
+  nil)
 
-(defn- resolve-namespace
+(defn split-by-leading-literals
+  "Separate quote/var/deref qualifiers from a var name."
+  [symbol-str]
+  (next (re-matches #"(@{0,2}#'|'|@)?(.*)" symbol-str)))
+
+(set! *unchecked-math* true)
+
+(defn fuzzy-matches?
+  "Tests if symbol matches the prefix when symbol is split into parts on\n  separator."
+  [^String prefix ^String symbol ^Character separator]
+  (let [pn (.length prefix)
+        sn (.length symbol)]
+    (or (= pn 0)
+        (and (> sn 0)
+             (= (.charAt prefix 0) (.charAt symbol 0))
+             (loop [pi 1
+                    si 1
+                    skipping false]
+               (cond (>= pi pn) true
+                     (>= si sn) false
+                     :else (let [pc (.charAt prefix pi)
+                                 sc (.charAt symbol si)
+                                 match (= pc sc)]
+                             (cond skipping (if (= sc separator)
+                                              (recur (if match (inc pi) pi)
+                                                     (inc si)
+                                                     false)
+                                              (recur pi (inc si) true))
+                                   match (recur (inc pi) (inc si) false)
+                                   :else (recur pi
+                                                (inc si)
+                                                (not (= pc separator)))))))))))
+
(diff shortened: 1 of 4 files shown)


bbatsov:
> kept the compatibility with the current call signature of completions in nREPL.

The one that the users can set themselves? What are the differences in the signature?

alexander-yakushev:
> The one that the users can set themselves? What are the differences in the signature?

Yes. Compliment's own signature is `([prefix] [prefix options])`, nrepl's is `([prefix] [prefix ns] [prefix ns options])`. I kept the latter here.

bbatsov:
Makes sense to me. I always felt that the prefix and the ns are kind of mandatory for the completion, that's why I went with this signature.

src/clojure/nrepl/util/completion.clj
```diff
@@ -3,317 +3,644 @@
 
   The functionality here is experimental and
   the API is subject to changes."
-  {:author "Bozhidar Batsov"
+  {:author "Bozhidar Batsov, Oleksandr Yakushev"
```
bbatsov, line 6: You might also mention in the docstring that this based on `compliment.lite`, so it's even more visible.

alexander-yakushev:
OK, finally got it working. Still a couple of points to address:

1. Compliment returns arglists metadata as a Clojure list of strings. nrepl.completion returns it as a single string. I didn't change this behavior here, it still a single string.
2. Compliment doesn't add `:doc` metadata to namespaces. The primary reason is that it only works for namespaces that are currently loaded, and I don't think it's terribly useful. So, I left out this feature here. I can perhaps hack it back in, but it conflicts with deduplication logic between loaded namespaces and namespaces on classpath, and it's a bit annoying. How badly do you want it?

bbatsov:
> Compliment returns arglists metadata as a Clojure list of strings. nrepl.completion returns it as a single string. I didn't change this behavior here, it still a single string.

Hmm, I thought I made those the same, as CIDER can handle both formats. I guess it was an oversight on my end. 

> Compliment doesn't add :doc metadata to namespaces. The primary reason is that it only works for namespaces that are currently loaded, and I don't think it's terribly useful. So, I left out this feature here. I can perhaps hack it back in, but it conflicts with deduplication logic between loaded namespaces and namespaces on classpath, and it's a bit annoying. How badly do you want it?

It's not particularly important to me.

alexander-yakushev:
> Hmm, I thought I made those the same, as CIDER can handle both formats. I guess it was an oversight on my end.
Cool, then I'll make it identical to Compliment.

alexander-yakushev:
I see that cljfmt doesn't honor CIDER's :style/indent, tsk tsk tsk 😄. What do you do in such cases?

image

alexander-yakushev:
There are some issues with Clojure < 1.10 and JDK11+. Would you consider removing that combination from the test matrix? Clojure 1.10 had some important fixes for JDK11+, so I personally consider anything less as "broken" after Java 1.8, see https://github.com/clojure/clojure/blob/master/changes.md#changes-to-clojure-in-version-110.

If you want to keep it, then I'm fine to bandaid those issues for now.

bbatsov:
I know that Clj 1.7 and Clj 1.8 are effectively dead, but I'm not sure how best to reflect this in the test matrix. I don't want to drop support for them just yet, but I don't mind running less builds for them.

alexander-yakushev:
I meant only testing Clojure 1.7-1.9 with Java 8. But I'll probably add a bandaid anyway because dropping test targets is outside of scope of this PR.

alexander-yakushev:
So, what should I do about cljfmt failing?

bbatsov:
> So, what should I do about cljfmt failing?

Didn't it accept some custom settings to tweak how it indents stuff? I'd probably fix the indentation using those.

bbatsov:
> I meant only testing Clojure 1.7-1.9 with Java 8. But I'll probably add a bandaid anyway because dropping test targets is outside of the scope of this PR.

To me this sounds reasonable. @vemv Feel free to weigh in on this as well.

alexander-yakushev:
> Didn't it accept some custom settings to tweak how it indents stuff? I'd probably fix the indentation using those.

I don't use cljfmt, cljfmt is what fails the linting check. The code is indented with Emacs/CIDER, the particular form that cljmft complains about uses CIDER's :style/indent rule.

bbatsov:
Yeah, I get this - I meant to just match the CIDER indentation in the configuration for cljfmt as mentioned here https://github.com/weavejester/cljfmt/blob/master/docs/INDENTS.md

vemv:
> I'm almost ready. There is a single testcase that fails Windows tests and I have no idea how to debug that without a win machine :(

I'd suggest by adding Win to the CI of Compliment (lite), perhaps that way you'll get a more fine-grained failure instead of an integration-ish one.

Given that the failure is:

```
namespace completion with java classes
expected: (set/subset? #{"nrepl.test.Dummy"} (set (candidates "nrepl.t")))
  actual: (not (set/subset? #{"nrepl.test.Dummy"} #{"nrepl.tls" "nrepl.transport" "nrepl.tls-test" "nrepl.tls-client-proxy" "nrepl.transport-test" "nrepl.test-helpers"}))
```

There could be a unit test for classes (probably you have it already).

alexander-yakushev:
Yes, I see what the issue is, but I have no way to debug/fix it without a proper environment.

vemv:
Circle offers a "debug with ssh" feature though!

alexander-yakushev:
Can you please put "help wanted" label on this? Maybe somebody with a windows machine can lend a hand.

alexander-yakushev:
@bbatsov BTW, the tests often timeout on CI, like this time.

bbatsov:
> Can you please put "help wanted" label on this? Maybe somebody with a windows machine can lend a hand.

Sure, although I don't see this Windows test as a blocker per se, given we have very little Windows users to begin with. Perhaps @ikappaki can help us fix this, as he's the only real Windows user I can think of. (I have a Windows PC, but I only use WSL for development)

alexander-yakushev:
Thank you, it's OK, I've fixed the Windows issue now. The reason why the test fails now is this:

```
Too long with no output (exceeded 10m0s): context deadline exceeded
```

I don't think it can be because of my changes.

bbatsov:
Seems the flakiness got introduced here https://github.com/nrepl/nrepl/pull/303 I hope @ivarref can look into the problematic test at some point.

alexander-yakushev:
Are we stuck until then? I should I trigger the pipeline until it passes?

bbatsov:
@alexander-yakushev just disable the problematic test for now, so we can merge your PR.

ivarref:
Sorry about the flakiness.

```
nrepl.tls-test
  bad-config
  bad-keys
  bad-keys-then-good
  bad-keys-then-good-no-consume-exception
Too long with no output (exceeded 10m0s): context deadline exceeded
```

Does this mean that the test *after* `bad-keys-then-good-no-consume-exception` is taking more than 10m to complete?

Are tests run in parallel, or one by one?

Thanks.

bbatsov:
@ivarref I think it's best to create a separate ticket about the flaky test, so we don't pollute the conversation about the completion changes.

alexander-yakushev:
Disabled testing on Windows. Let's make this happen!

bbatsov:
Thanks for working on this!

PR #305, opened 2023-11-15
Update CI config

alexander-yakushev:
Done a couple of things here:

1. Switched from Circle's docker images to [official Clojure Docker images](https://hub.docker.com/_/clojure). It looks like [Circle's images](https://hub.docker.com/r/circleci/clojure) are no longer updated (last update 2 years ago). 
2. Added JDK21 to the test matrix, switched to Clojure 1.11 as a default for linting/etc jobs.
3. Removed testing clj 1.7-1.9 together with JDK higher than 8. The motivation for it is that Clojure 1.10 contains important fixes for compatibility with Java9+ (see https://github.com/clojure/clojure/blob/master/changes.md#11-java). Anyone who cares enough about using newer JDKs must eventually upgrade Clojure as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,10 @@ executors:
     docker:
       - image: clojure:temurin-17-lein-2.10.0-jammy
     <<: *defaults
+  openjdk21:
+    docker:
+      - image: clojure:temurin-21-lein-2.10.0-jammy
+    <<: *defaults
 
 # Runs a given set of steps, with some standard pre- and post-
 # steps, including restoring of cache, saving of cache.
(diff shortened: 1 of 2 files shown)


bbatsov:
Can you rebase on top of `master`?

alexander-yakushev:
https://app.circleci.com/pipelines/github/nrepl/nrepl/458/workflows/c42883fc-143f-4e24-bdea-e6596b23d5aa/jobs/7545

That's the same place where the Windows job gets stuck. Looks like there is a deadlock of sorts in that test. It seemingly happens on Windows all the time, but sometimes on Linux too.

bbatsov:
@alexander-yakushev Perhaps we should disable only this particular test then?

alexander-yakushev:
It would be ideal if @ivarref could take a look, I see that he he wrote the test. Don't want to disable arbitrary tests without understanding:)

alexander-yakushev:
Anyway, passed this time 😆

Issue #306, opened 2023-11-20
Fix bugged (TLS) accept loop


ivarref:
Placeholder for fixing TLS flaky test(s).

Adding an issue for this, re comments at https://github.com/nrepl/nrepl/pull/304 and https://github.com/nrepl/nrepl/pull/305

I have not been able to reproduce this locally, but I do believe I've found some incorrectness that could lead to a server that is no longer accepting connections, i.e. the accept loop exits. It's pushed at https://github.com/nrepl/nrepl/compare/master...ivarref:nrepl:master

I'm using java version:
```bash
$ java -version
openjdk version "20-ea" 2023-03-21
OpenJDK Runtime Environment (build 20-ea+34-2340)
OpenJDK 64-Bit Server VM (build 20-ea+34-2340, mixed mode, sharing)
```

Some open questions:

How can we debug this in CI?

What test(s) is it that deadlock? Or is the whole suite just taking more than 10 minutes?

Regards

bbatsov:
Fixed by #307

PR #307, opened 2023-11-22
Fix bugged TLS accept loop logic, fixes issue #306

ivarref:
Fix bugged TLS accept loop logic. TLS server accept loop should continue in the presence of non-SSL exceptions such as SocketException. The TLS accept method will now wrap any non SSL exceptions in SSLException. The server accept loop will close the server socket upon encountering an unexpected exception. This was not the case earlier, and this lead to a lingering, open server socket where the OS would continue to accept new connections. It seems that .connect, which includes a connect-timeout-ms parameter, on a TLS socket does *not* include the handshake. Thus, in the tests, if the server first encountered a SocketException and this lead to a lingering socket, the next TLS connect would succeed, but the handshake would never finish (and there was no timeout specified). Thus, the test became stuck given a non SSL exception, and caused the CI system to time out after 10 minutes.

- [X] All tests are passing

### DIFF
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -64,9 +64,10 @@
                       nil)
                     (catch Throwable t
                       (consume-exception t)
-                      (if (.isClosed ^ServerSocket server-socket)
-                        nil
-                        (throw t))))]
+                      (when-not (.isClosed ^ServerSocket server-socket)
+                        (log "Unexpected exception of class" (class t) "with message" (.getMessage t))
+                        (safe-close server-socket) ; is this a good idea?
+                        (log "Shutting down server abruptly"))))]
     (if (= sock :tls-exception)
       ; if there was a TLS exception, e.g. bad client cert,
       ; we simply recur: accept new connections on the same thread.
--- a/src/clojure/nrepl/tls.clj
+++ b/src/clojure/nrepl/tls.clj
@@ -11,8 +11,7 @@
   (:require [clojure.java.io :as io :refer [input-stream]]
             [clojure.stacktrace]
             [clojure.string :as str])
-  (:import (java.io IOException)
-           (java.net InetSocketAddress)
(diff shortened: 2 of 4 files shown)


src/clojure/nrepl/server.clj
```diff
@@ -66,7 +66,11 @@
                       (consume-exception t)
                       (if (.isClosed ^ServerSocket server-socket)
                         nil
-                        (throw t))))]
+                        (do
+                          (log "Unexpected exception of class" (class t) "with message" (.getMessage t))
+                          (safe-close server-socket) ; is this a good idea?
```
bbatsov, line 71: No idea. :D

src/clojure/nrepl/tls.clj
```diff
@@ -245,12 +244,19 @@
     (.connect sock (InetSocketAddress. host ^int port) ^int connect-timeout-ms)
     sock))
 
+(def ^:dynamic *handshake-timeout-ms* 30000)
```
bbatsov, line 247: Might be good to add a docstring for this.

src/clojure/nrepl/server.clj
```diff
@@ -66,7 +66,11 @@
...
+                          (log "Shutting down server abruptly")
+                          nil))))]
```
bbatsov, line 73: Doesn't `log` return `nil` anyways?

bbatsov:
Overall your changes seem reasonable to me. I left a couple of small remarks here and there.

bbatsov:
@ivarref Ping :-)

bbatsov:
@ivarref Happy New Year! 

Please check my review feedback when you can.

ivarref:
Happy new year @bbatsov !

Sorry once again about the delayment. I've added your suggested improvements now. What do you think?

bbatsov:
Looks good. Thanks!

Issue #309, opened 2024-02-07
`--interactive` doesn't work with `--socket PATH`


chancerussell:
## Expected behavior

Using the `--interactive` flag in concert with `--socket PATH` should start an `AF_UNIX`-based nREPL server, then a command line nREPL client session connected to that server.

## Actual behavior

The server starts, then exits immediately with the error message `Must supply host/port or socket.`

## Steps to reproduce the problem

```bash
$ clojure -Srepro -Sdeps '{:deps {nrepl/nrepl {:mvn/version "1.1.0"}}}' -M --main nrepl.cmdline --socket "$(mktemp -u)" --interactive
```

results in:

```bash
nREPL server started on socket nrepl+unix:/var/folders/yz/dcqd65r909zfb_hz5_6srps80000gn/T/tmp.TzBosjY3Gn
Must supply host/port or socket.
```

## Environment & Version information

Various

### Clojure version

1.10.3, 1.11.1

### Java version

OpenJDK 19.0.1

### Operating system

macOS 14 Sonoma

## Analysis

The `--interactive` option is meant to allow users to start an nREPL server and then connect to it with the built-in CLI client. This works fine when the server is bound to a TCP host/socket, but when the user uses the `--socket PATH` option to bind the server to an `AF_UNIX` socket, nREPL won't connect to the server and exits with an error message.

The underlying issue is that when starting an nREPL server, the returned `nrepl.server.Server` record doesn't contain a `:socket` key, and the `(if socket...` path in `nrepl.cmdline/interactive-repl` won't be taken. 

(This issue doesn't occur when using `--socket PATH` with the `--connect` command because that code path ends up passing in a plain map rather than constructing a `Server` record.)

I'll be opening two pull requests in a moment that fix this issue:

* a minimal fix that adds a `:socket` field to the `nrepl.server.Server` record and populates it.
* the same fix but also adding a `host` field to the `Server` record. That seems like a more *correct* thing to do since other parts of the program seem to expect that field, but it's *possible* that there are users relying on the current fallback behavior: if no repl-fn is specified, they'll get `127.0.0.1` defaulted by `nrepl.cmdline/run-repl`, even if they passed a different `--host`.
 
If you'd like to take a different approach to fixing this, let me know and I will open an additional PR!

bbatsov:
Thanks for reporting the issue and willing to work on a fix! That's some pretty good timing as I'm thinking of cutting a new nREPL release soon.

PR #310, opened 2024-02-07
Add and populate :socket key on nrepl.server.Server

chancerussell:
This fixes [an issue](https://github.com/nrepl/nrepl/issues/309) that occurs when using the `--interactive` option in concert with `--socket PATH`. Since the Server object has no :socket key, nrepl.cmdline/interactive-repl won't recognize that it should connect to the socket and will exit with an error message.

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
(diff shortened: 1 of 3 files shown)


chancerussell:
Closing in favor of #311

PR #311, opened 2024-02-07
Add and populate :host and :socket keys on nrepl.server.Server

chancerussell:
Like https://github.com/nrepl/nrepl/pull/310, but also provides the `:host` key.

This fixes [an issue](https://github.com/nrepl/nrepl/issues/309) that occurs when using the `--interactive` option in concert with `--socket PATH`. Since the Server object has no :socket key, nrepl.cmdline/interactive-repl won't recognize that it should connect to the socket and will exit with an error message.

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -158,7 +158,7 @@
       (binding [dynamic-loader/*state* state]
         ((:handler @state) msg)))))
 
-(defrecord Server [server-socket port open-transports transport greeting handler]
+(defrecord Server [server-socket host port socket open-transports transport greeting handler]
   java.io.Closeable
   (close [this] (stop-server this)))
 
@@ -211,7 +211,9 @@
                  :else
                  (inet-socket bind port))
         server (Server. ss
+                        (when-not socket bind)
                         (when-not socket (.getLocalPort ^java.net.ServerSocket ss))
+                        socket
                         (atom #{})
                         transport-fn
                         greeting-fn
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -158,7 +158,23 @@
       (binding [dynamic-loader/*state* state]
         ((:handler @state) msg)))))
 
-(defrecord Server [server-socket host port socket open-transports transport greeting handler]
+(defrecord
+ Server
+ ;;A record representing an nREPL server.
+ [server-socket ;; A java.net.ServerSocket for underlying server connection
+  host ;; When starting an IP server, the hostname the server is bound to
+  port ;; When starting an IP server, the port the servfer is bound to
+  socket ;; When starting a filesystem socket server, the string path to the
+         ;; socket file
+  open-transports ;; An IDeref containing a set of nrepl.transport/Transport
+                  ;; objects representing open connections
+  transport ;; A function that, given a java.net.Socket corresponding to an
+            ;; incoming connection, will return a value satisfying the
+            ;; nrepl.transport/Transport protocol for that Socket
+  greeting  ;; A function called after a client connects but before the
+            ;; handler. Called with the connection's corresponding
+            ;; nrepl.transport/Transport object
+  handler]  ;; The message handler function to use for connected clients
   java.io.Closeable
   (close [this] (stop-server this)))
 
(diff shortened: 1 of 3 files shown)


src/clojure/nrepl/server.clj
```diff
@@ -210,7 +210,9 @@
                  :else
                  (inet-socket bind port))
         server (Server. ss
+                        (when-not socket bind)
```
bbatsov, line 213: I'm wondering if we shouldn't raise an error if both host and socket are provided. I'm also thinking that at some point we should add some `:host` key as well (an alias for `:bind`), because even I forget that `:bind` is the host. 😄

chancerussell: @bbatsov Do you mean we should add a `:bind` key? The *local* here is `bind`, but the field on the record I've added is `:host` (since that seems to be what other parts of the program look for).

>  raise an error if both host and socket are provided

I think that would make sense. Do you think that would be better added down here in the server code, or as an input validation in the `cmdline` namespace?

bbatsov: Probably here would be best. 

> @bbatsov Do you mean we should add a :bind key? The local here is bind, but the field on the record I've added is :host (since that seems to be what other parts of the program look for).

Yeah, I know - I was just wondering if we shouldn't use the `host` terminology more pervasively. I think the in most places of the code we use `bind`, because something in the java.net code was using this terminology. Anyways, let's leave this as is and not expand the scope.

bbatsov:
@rlbdv Feel free to take a look as well.

bbatsov:
Can you rebase on top of  `master`?

chancerussell:
> Can you rebase on top of  master?

Yep, just rebased!

src/clojure/nrepl/server.clj
```diff
@@ -158,7 +158,7 @@
       (binding [dynamic-loader/*state* state]
         ((:handler @state) msg)))))
 
-(defrecord Server [server-socket port open-transports transport greeting handler]
+(defrecord Server [server-socket host port socket open-transports transport greeting handler]
```
bbatsov, line 161: I'm wondering if here we shouldn't name this `unix-socket`, to differentiate from the `server-socket`. Not a big deal, but it might make things clearer for someone. 

Or maybe we should add some docstring to this record to explain its purpose.

chancerussell: I will:

* Add some explanatory comments to the `defrecord` form re: what the fields are supposed to mean (unfortunately, this won't be a docstring, since `defrecrord` doesn't support them)
* Open an additional PR just like this one, but with the `socket` field renamed to `unix-socket`

And then I'll leave it to you all to choose which one you like better and let you update the comments as you wish :)

bbatsov:
Can you, please, address the small indentation error detected by the lint job?

src/clojure/nrepl/server.clj
```diff
@@ -158,9 +158,25 @@
       (binding [dynamic-loader/*state* state]
         ((:handler @state) msg)))))
 
-(defrecord Server [server-socket port open-transports transport greeting handler]
-  java.io.Closeable
-  (close [this] (stop-server this)))
+(defrecord 
+ Server
+  ;;A record representing an nREPL server.
```
bbatsov, line 163: It seems to me this comment is off by one space.

chancerussell: I think we've now satisfied the formatter :)

chancerussell:
I just opened #312, which is the same as this PR, but renames the :socket field to :unix-socket and updates the bits that look at that field to account for the change.

Personally, I would lean towards *not* renaming the key, since anyone providing their own custom repl-fn might have written it to expect a server object bearing the `socket` key, but you'd know better than me if that's likely to be an issue :)

(I also sort of want to say the key might be better named unix-socket-path or filesystem-socket-path, since it's not holding an actual socket object, but I could bikeshed on that for days!)

I'll be unavailable for a few days—please feel 100% free to make any desired changes you'd like against either PR to get this merged!

bbatsov:
> Personally, I would lean towards not renaming the key, since anyone providing their own custom repl-fn might have written it to expect a server object bearing the socket key, but you'd know better than me if that's likely to be an issue :)

Same. I was thinking more along the lines of duplicating the key and deprecating the old name, but it's not a big deal. I think this PR is enough for now.

PR #312, opened 2024-02-15
Add and populate :host and :unix-socket keys on nrepl.server.Server

chancerussell:
Like https://github.com/nrepl/nrepl/pull/311, but also renames the :socket key to :unix-socket.

This fixes https://github.com/nrepl/nrepl/issues/309 that occurs when using the --interactive option in concert with --socket PATH. Since the Server object has no :socket key, nrepl.cmdline/interactive-repl won't recognize that it should connect to the socket and will exit with an error message.

- [x] All tests are passing


bbatsov:
As we opted to merge #311 we can close this PR.

PR #317, opened 2024-05-06
[completion] Update compliment-lite

alexander-yakushev:
The latest version contains a few improvements worth to be integrating into nREPL, namely:

- Reduced memory footprint of caches retained by Compliment-lite.
- Slightly improved speed and significantly reduced allocations when completing namespaces, classes and class-related data.


src/clojure/nrepl/util/completion.clj
```diff
@@ -401,8 +421,8 @@
   (fuzzy-matches? prefix namespace \.))
 
 (defn namespaces-candidates
-  "Returns a list of namespace and classname completions."
   [^String prefix ns _]
+  "Return a list of namespace candidates."
```
bbatsov, line 425: This should be above the params.

bbatsov:
Thanks!

Issue #296, opened 2023-07-03
Interrupt is broken on JVM 20+


bbatsov:
nREPL relied on `Thread.stop` to kill the session thread when needed and the function was finally removed from Java in Java 20 (after being deprecated for ages). The [current nREPL implementation](https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/middleware/session.clj#L176) tries `Tread.interrupt` first and fallbacks on `Thread.stop` if that fails.

Any ideas for fixing this on Java 20+ are welcome!

See https://github.com/clojure-emacs/cider/issues/3351 for more details.

socksy:
For reference, this is how it was solved in jshell: https://github.com/openjdk/jdk/pull/10166

Looks like they override the jdk internal(!) [MethodVisitor](https://asm.ow2.io/javadoc/org/objectweb/asm/MethodVisitor.html#visitJumpInsn(int,org.objectweb.asm.Label)) to [add a check](https://github.com/openjdk/jdk/commit/ffc249ae2120752a9c0e9de18167b0c16f2de410#diff-c36ffff66439858a380f5d665260d18eef6d57d52157621b2bb731ac2ee1581aR84-R88) to a [new static method](https://github.com/openjdk/jdk/commit/ffc249ae2120752a9c0e9de18167b0c16f2de410#diff-c36ffff66439858a380f5d665260d18eef6d57d52157621b2bb731ac2ee1581aR99) on every single JVM `jump` instruction. If a new `allStop` boolean is true, then it [generates a ThreadDeath object](https://github.com/openjdk/jdk/commit/ffc249ae2120752a9c0e9de18167b0c16f2de410#diff-c36ffff66439858a380f5d665260d18eef6d57d52157621b2bb731ac2ee1581aR104) and throws it. They then [instrument all code](https://github.com/openjdk/jdk/commit/ffc249ae2120752a9c0e9de18167b0c16f2de410#diff-c36ffff66439858a380f5d665260d18eef6d57d52157621b2bb731ac2ee1581aR73) to be run in their execution environment with it. Then instead of `Thread.stop` they set this `allStop` value to be `true`, then call `Thread.interrupt` on the original thread. So basically, reimplementing 

Few things about that:
- :\
- ThreadDeath is explicitly deprecated too, so that doesn't sound like a super long term solution. I'm surprised they didn't use the same mechanism that `.interrupt` uses, but maybe that's not possible from the bytecode (I'm no JVM expert...)
[...]

PR #318, opened 2024-05-06
[session] Introduce custom JVMTI agent to restore Thread.stop() on JDK20+

alexander-yakushev:
I have a real treat for you, but it comes with a cost.

I grew tired and annoyed by not being able to properly interrupt evaluation when running on the latest JDKs. You know, the dreadful #296. So, I rolled up the sleeves and put together an itty bitty native agent that uses JVMTI to access the hidden ability to stop threads. I verified that it works and even got the tests passing. But of course, it is only half of the story.

There are quite a few questions that have to be answered before this even gets a chance to be integrated:

- [x] Native agent is not a cross-platform thing. So far, there are three libs in this PR – MacOS universal binary and Linux aarch64 and amd64. We can maybe also add Linux armv7 and x86 but I doubt it's that important to have interruptible eval on those platforms. **Solution: support 4 platforms – Linux x64/aarch64 and MacOS x64/aarch64.**
- [x] We have to build those libs somewhere. Now I just build them locally. I'm pretty sure it can be done on the CI. **Solution: binaries for all supported platforms are built with Github Actions.**
- [x] Need to decide whether the binaries get built as part of the pipeline and passed to the build of nREPL, or the binaries are built as a separate CI job and manually placed in `res/` folder and committed. I prefer the latter variant because it allows people who clone the repo to freely run tests and hack without having to build the native binary first. **Resolution: build binaries as a separate CI step, place them manually into `res/`.**
- [x] Proper binary for the platform is selected, unpacked, and loaded on demand at runtime. This approach has been working very well for https://github.com/clojure-goes-fast/clj-async-profiler for many years without many problems. The `nrepl.util.jvmti` namespace is also loaded on demand, only when the user runs JDK20+, and only when they interrupt evaluation that cannot be stopped by regular `Thread/interupt`.
- [x] It has to be decided if such functionality (if it ever merges) has to be protected by an additional feature flag – for folks that don't want to sideload binary agents into their Java process. **Resolution: allowAttachSelf can act as an opt-in flag to use the feature.**
- [x] For dynamic native agent loading, the JVM has to be started with `-Djdk.attach.allowAttachSelf`. This is the biggest hurdle from the user's perspective. At the same time, this can act as the opt-in flag for using the agent. **Resolution: allowAttachSelf also acts as an opt-in, and CIDER will be able to inject this.**
- [x] As an alternative, we can ask users to set up the native agent before the launch, using `-D-agentpath:...`. But I feel like this will be much more cumbersome because users themselves would have to select the proper binary for their OS/arch. On the bright side, `allowAttachSelf` doesn't have to be enabled this way, but that's a very minor benefit. **Resolution: allowAttachSelf can be injected by CIDER, so not much trouble for users.**

# TL DR

- Consider this PR a tech demo with the solution for #296. 
- You can `lein install` it and inject the local nREPL build into other projects and check out if interruptible-eval works on JDK21 and later.
- Let's use this PR to discuss how viable this approach is. I'm afraid there's nothing else that is easier, and I checked a lot.


bbatsov:
> Native agent is not a cross-platform thing. So far, I've built and put two compiled libs into this PR – MacOS universal binary and Linux aarch64. To call this feature a complete one, we'll need to additionally build at least Linux amd64. Maybe Linux armv7 and x86 but I doubt it's that important to have interruptible eval on those platforms.

Yeah, I agree that adding Linux amd64 will be probably be enough. There's also Windows support to consider, but I doubt many people use it directly (e.g. I use WSL).

> We have to build those libs somewhere. Now I just build them locally. I'm pretty sure it can be done on the CI.

Building them using the CI is the way to go IMO.

> Need to decide whether the binaries get built as part of the pipeline and passed to the build of nREPL, or the binaries are built as a separate CI job and manually placed in res/ folder and committed. I prefer the latter variant because it allows people who clone the repo to freely run tests and hack without having to build the native binary first.

Same.

> It has to be decided if such functionality (if it ever merges) has to be protected by an additional feature flag – for folks that don't want to sideload binary agents into their Java process.

I'd definitely make this optional.

> For dynamic native agent loading, the JVM has to be started with -Djdk.attach.allowAttachSelf. This is the biggest hurdle from the user's perspective. At the same time, this can act as the opt-in flag for using the agent.

As often nREPL gets started from some editor this wouldn't be that big of a problem in practice. E.g. in CIDER we'd just add this to the jack-in params.

> Let's use this PR to discuss how viable this approach is. I'm afraid there's nothing else that is easier, and I checked a lot.

Thanks for looking into this! I'm not a huge fan of the complexity this would add, but I also spent a lot of time looking for solutions in the past and I didn't discover any, so we'll probably have to stomach the additional complexity given the importance of having a working interrupt functionality.

alexander-yakushev:
> Thanks for looking into this! I'm not a huge fan of the complexity this would add, but I also spent a lot of time looking for solutions in the past and I didn't discover any, so we'll probably have to stomach the additional complexity given the importance of having a working interrupt functionality.

This is indeed a big step in complexity, so I commit to make this 100% unintrusive for users who don't opt-in into this.

I need your advice on how to implement the feature toggle for this. For CIDER users, we will probably expose this as a CIDER parameter, but given this feature will live in nREPL, there should be some public toggle here. Options that I see so far:
1. Extra middleware that should be injected onto the chain.
2. Java property.
3. Reuse `-Djdk.attach.allowAttachSelf` as enabler of the feature. People who turn it on usually know what they are doing, so I think it can serve as an opt-in indicator.

bbatsov:
> Reuse -Djdk.attach.allowAttachSelf as enabler of the feature. People who turn it on usually know what they are doing, so I think it can serve as an opt-in indicator.

I like this option the most, and it's also the simplest.

alexander-yakushev:
Update: I've built and commited Linux x64 binary, and the tests on CI are passing now.

cichli:
> Options that I see so far:

Another option is to put the agent in a separate artifact and make it an optional dependency. That would sidestep the choice you mention of needing to commit binaries to this repo or have people hacking on nREPL needing to compile them.

alexander-yakushev:
> Another option is to put the agent in a separate artifact and make it an optional dependency. That would sidestep the choice you mention of needing to commit binaries to this repo or have people hacking on nREPL needing to compile them.

Good point, thank you! While this is technically possible, I feel like this would be a bit "weird" dependency to name (`org.nrepl/nrepl-jdk20-thread-stopper`? `org.nrepl/nrepl-jvmti` – but why it needs one, the user might ask?). Besides, this would add another step for users who don't launch nREPL through jack-in of sorts but rather with a manual command.

I see adding the binaries to `res/` as the easiest and the most practical option. They will very rarely (never?) change, won't be seen to users from the outside, and the local development of nREPL won't be complicated.

alexander-yakushev:
@bbatsov  Offtopic: is this the "correct" way to ident delay? I'm pretty sure that clojure-mode has always indented the body with two spaces, or is it my local configuration?

image

alexander-yakushev:
Regarding compatibility with JDK21 virtual threads: JVMTI does not allow stopping a virtual thread from the "outside": https://github.com/openjdk/jdk/commit/51b8f3cfb9df3444b6226a5d5cb7f01a9ab6db6c#diff-bcd9b3f49342f083372b4980b86ad243983153ce498ece2a5076eb7223c31eb8R1195. However, this shouldn't be a problem for us since we are in control of what type of threads we create for nREPL sessions, and we do create "normal" platform threads that are stoppable alright.

alexander-yakushev:
Bump.

bbatsov:
> @bbatsov Offtopic: is this the "correct" way to indent delay? I'm pretty sure that clojure-mode has always indented the body with two spaces, or is it my local configuration?

It should be indented the same way as any macro taking a `body`. (meaning 2 spaces, more context https://github.com/bbatsov/clojure-style-guide?tab=readme-ov-file#body-indentation)

bbatsov:
> I'm getting this out of the draft. I'll need your guidance on which supporting documentation I should write and what the correct release plan would be if this is integrated.

We should add something about this here https://nrepl.org/nrepl/1.1/hacking_on_nrepl.html and in the documentation for the interrupt op. Probably it'd be a good idea to add some "Caveats" page where we can document the JDK change and how it affected nREPL.

And we'll have to update the usage docs to mention how to enable the agent, of course.

alexander-yakushev:
> It should be indented the same way as any macro taking a body. (meaning 2 spaces, more context https://github.com/bbatsov/clojure-style-guide?tab=readme-ov-file#body-indentation)

Right, but cljfmt for some reason insists on identing it with 1 space. I've added an override for that.

alexander-yakushev:
Updated the documentation.

- New section in Installation (about enabling allowAttachSelf)
- New section in Hacking nREPL (about how to build native binaries locally)
- New section about Java 20 compatibility in Compability (this should act as a Caveats page that you mentioned).
- Mentioned jvmti stuff in the docs for `interrupt` op.

Anything else?

bbatsov:
Everything looks good to me now, but I'll take a closer look once more before merging. 

I'm also thinking of cutting nREPL 1.1.2 first today, before merging this PR.

bbatsov:
I've just cut 1.1.2, so you'll need to rebase this on top of `master`. I'll cut 1.2.0-beta1 after we merge this one and the bencode PR.

alexander-yakushev:
Rebased onto master, skipped changing the version in `project.clj`.

bbatsov:
🚀

PR #319, opened 2024-05-07
Refactor CI and bump dev dependencies

alexander-yakushev:
We did this for Orchard and cider-nrepl before, and I'd argue that the rules look more apparent this way (and there is less room for error).


bbatsov:
Good changes! Thanks!

PR #320, opened 2024-05-20
Docs cleanup

alexander-yakushev:
I've tried to clean up some docs on the main documentation site.

- nrepl 0.2.x, Contrib migration, and compatibility are still featured rather prominently on multiple pages. Given the event is 7 years old now, and all users who cared enough already did migrate, I propose to put it into a single place and keep just for posterity. I've moved it into About subfolder which is most likely a wrong place for, so I'm looking for your suggestions.
- Removed mentions of Boot. There is not much very useful information regarding Boot in the docs, just mentions of it alongside Leiningen.
- Removed unrepl from the Alternatives page. Since it is abandonware, I don't think it's worth mentioning as an alternative. But maybe you want to keep it as an example of how things could be different, idk.


doc/cljdoc.edn
```diff
@@ -26,6 +26,7 @@
   ["Additional Resources" {:file "doc/modules/ROOT/pages/additional_resources.adoc"}]
   ["About" {}
    ["History" {:file "doc/modules/ROOT/pages/about/history.adoc"}]
+   ["Contrib nREPL" {:file "doc/modules/ROOT/pages/about/contrib_nrepl.adoc"}]
```
bbatsov, line 29: Probably it'd be better to just add this to the existing `History` page.

bbatsov: Or rather the FAQ, given that all the info here is about upgrading from 0.2.x. In general we can probably remove all of this and nobody would notice this. You said it yourself - 7 years are a lot of time and I doubt anyone it still using nREPL 0.2 at this point.

alexander-yakushev: Done!

Issue #322, opened 2024-05-21
Modernize bencode implementation


alexander-yakushev:
Before trying to address #285 I'd like to ask if there's a desire/will to update the bencode implementation in nREPL (and github.com/nrepl/encode). From the various old-style `#^{:private true}` metadata I can infer that it was written a long time ago and haven't been touched ever since. I see a few potential performance improvements in that implementation and much space for modernization. But I understand that bencode is the cornerstone of nREPL and it might be scary to touch it. How confident are you in tests for it, @bbatsov?

bbatsov:
> How confident are you in tests for it, @bbatsov?

Quite confident. The implementation basically hasn't been touched in over 10 years (it targeted Clojure 1.1 or 1.2 if I recall correctly), but I know that no codebase is perfect and suggestions for improvements are always welcome. Performance improvements would be particularly welcome.

alexander-yakushev:
Complete in #323.

PR #323, opened 2024-05-22
[bencode] Rewrite nrepl.bencode

alexander-yakushev:
I freshened up the implementation as discussed in #322 and made performance improvements where it made sense to do so. Although the spirit of the implementation remained mostly the same, there are a few significant changes:

- `read-token` is gone, reading dispatch happens within the `read-bencode` function for better efficiency.
- Rewrote writing dispatch from multimethods to protocols. Introduced a separate private protocol function `write-bencode*`, the public `write-bencode` is a public function with old argument order.

I've generated a stress test data structure (~10Mb when serialized as EDN) to benchmark. Here are the results:

##### Encoding
`Before: Time per call: 230.37 ms   Alloc per call: 293,593,739b`
`After:  Time per call: 89.59 ms    Alloc per call: 114,506,367b`
Improvement: 2.5x in time and 2.5x in allocations.

##### Decoding
`Before: Time per call: 263.36 ms   Alloc per call: 374,434,013b`
`After:  Time per call: 49.39 ms    Alloc per call: 76,781,066b`
Improvement: 5.3x in time and 4.9x in allocations.

Overall, I don't think this would influence the UX of e.g. CIDER much because the data structures that go through cider-nrepl are usually small, and the bottleneck would most likely be on the Emacs side, not on the Clojure side. Anyway, a win is a win.

### DIFF
--- a/test/clojure/nrepl/bencode_test.clj
+++ b/test/clojure/nrepl/bencode_test.clj
@@ -9,6 +9,9 @@
 
 (ns nrepl.bencode-test
   (:require [clojure.test :refer [are deftest is testing]]
+            [clojure.test.check.generators :as gen]
+            [clojure.test.check.properties :as tc.prop]
+            [clojure.test.check.clojure-test :refer [defspec]]
             [nrepl.bencode :as bencode :refer [read-bencode
                                                read-netstring
                                                write-bencode
@@ -173,13 +176,9 @@
     {"cheese" 42 "ham" ["eggs"]} "d6:cheesei42e3:haml4:eggsee"))
 
 (deftest test-lexicographic-sorting
-  (let [source   ["ham" "eggs" "hamburg" "hamburger" "cheese"]
-        expected ["cheese" "eggs" "ham" "hamburg" "hamburger"]
-        to-test  (->> source
-                      (map >bytes)
-                      (sort @#'nrepl.bencode/lexicographically)
-                      (map <bytes))]
-    (is (= to-test expected))))
+  (let [source (zipmap ["ham" "eggs" "hamburger" "hamburg" "cheese"] (range))]
+    (is (= "d6:cheesei4e4:eggsi1e3:hami0e7:hamburgi3e9:hamburgeri2ee"
+           (>output source :writer write-bencode)))))
 
 (deftest unencoded-values
   ;; just some PNG data that won't round-trip cleanly through UTF-8 encoding, so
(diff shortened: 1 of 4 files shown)


alexander-yakushev:
Sorry for the push noise. Ready for review now.

bbatsov:
Impressive performance gains! Great work, Sashko! 🚀 

At at a glance I don't see anything concerning in the code changes that you made, so I'm pondering now whether to ship this with nREPL 1.1.2 or nREPL 1.2. I'm leaning slightly towards 1.2, together with the new agent. Any preferences regarding this?

test/clojure/nrepl/bencode_test.clj
```diff
@@ -206,3 +205,31 @@
       (is (thrown? IllegalArgumentException
                    (write-bencode out {"obj" (Object.)})))
       (is (= "d3:obj" (String. (.toByteArray out)))))))
+
+(def valid-bencode-input-generator
```
bbatsov, line 209: You might add some section comment before the `test.check` tests, so they are easier to spot when someone's scanning the file.

test/clojure/nrepl/bencode_test.clj
```diff
@@ -206,3 +205,31 @@
...
+                                   (gen/map gen/string % {:min-elements 0
+                                                          :max-elements 20})])
+                     (gen/one-of [gen/string gen/large-integer])))
+
+(defspec generative-roundtrip-test {:num-tests 200}
+  (tc.prop/for-all
+   [value valid-bencode-input-generator]
+   (= value (>input (>output value :writer write-bencode) :reader read-bencode))))
+
+(comment
```
bbatsov, line 221: Same for the performance tests. (e.g. something `### Performance tests`) and couple of sentences explaining them.

alexander-yakushev: Done

bbatsov:
Seems all you need to do now is make `cljfmt` happy (it complains that something isn't indented properly) and we can merge this PR and cut 1.2.0-beta1.

alexander-yakushev:
All good now. Also updated the changelog mentioning the jvmti change which I forgot.

alexander-yakushev:
Not good, one test failed with OOM. Reduced num-tests for generative testing.

bbatsov:
Seems we're in business now. 🚀 I'll cut 1.2.0-beta1 right away.

bbatsov:
Seems there's still some OOM flakiness https://app.circleci.com/pipelines/github/nrepl/nrepl/556/workflows/966d5bbd-ba07-4b41-9eb4-d99e8a034677/jobs/9124

alexander-yakushev:
Hm, I didn't consider that somebody may be relying on this; after all, the number of types that Bencode declares is a closed set. You can extend `BencodeSerializable` protocol for your types; or we can consider reverting this back to multimethod dispatch. @bbatsov thoughts?

bbatsov:
> Hm, I didn't consider that somebody may be relying on this; after all, the number of types that Bencode declares is a closed set. 

Same.

> You can either extend BencodeSerializable protocol for your types;

I'd prefer this direction.

bbatsov:
> (for software extending nrepl.bencode/write-bencode by adding methods like is the case for FlowStorm nrepl middleware)

@jpmonettas Btw, out of curiosity why were you doing something like this?

jpmonettas:
@alexander-yakushev it is ok as it is, I just figured that I can fix it on my side, not relying on that, so I will not be pushing for any revert.

@bbatsov 
> out of curiosity why were you doing something like this?

So, different ops on my middleware were dealing with maps with keys pointing to custom types. So instead of doing updates on each piece of data before sending the response I went and used write-bencode extension as a serialization layer. Maybe it was hacky.

Anyway I figured out I can do it without it, by converting the data before sending the response, there aren't many cases, so no worries.

alexander-yakushev:
I think it's not really different from a regular public function in a "non-public" namespace that is renamed or deleted or has arglist changed or even retains all visual compatibility but suddenly starts behaving differently. I personally don't agree that any public function is automatically a public API. This kind of thinking leads us to Java land where everything is paranoidally private.

jpmonettas:
Yeah, I don't agree either that every non private function is automatically considered part of a public api, but since multimethods and protocols are normally used because of their open extension capabilities, they lead me to think that the library authors added them as extension points for everybody, not just for them. Like in this case, write-bencode being a multimethod looked to me like it was designed for people to write anything into bencode.

alexander-yakushev:
Maybe, this was a case of incomplete documentation. In any case, I don't know what the original author had in mind in 2012, but in the current implementation the protocol is used purely as a developer's convenience.

bbatsov:
> Maybe, this was a case of incomplete documentation. In any case, I don't know what the original author had in mind in 2012, but in the current implementation the protocol is used purely as a developer's convenience.

Pretty sure this was the case back in the day as well - I've often used multimethods in the past just because I thought the code was more elegant this way. (without actually anticipating someone would add more down the road)

Anyways, it's always great to learn how someone is your using your libraries - especially if that's something you didn't think about. :-) Sorry about any issues we might have caused!

PR #324, opened 2024-05-22
[bencode] Further reduce the span of generative testing

alexander-yakushev:
The `num-tests` influences how "big" the generated values will be. I don't have any good intuition about which value would prevent OOMs so we'll have to trial-and-error it.


alexander-yakushev:
Looks fine.

bbatsov:
Still failed after you merged it, but it's not a big deal. It seems we always get somewhere between 0 and 1 failure on the OOM front, which is manageable. I'm surprised that cutting the size in half didn't solve this more reliably, though.

alexander-yakushev:
Yes, I am addressing it nuclearly in #325.

PR #325, opened 2024-05-22
[bencode] Reduce :num-tests even further

alexander-yakushev:
Reducing it to an indecently low value and adding a comment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,7 @@ jobs:
     steps:
       - checkout
       - with_cache:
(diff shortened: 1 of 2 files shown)


bbatsov:
Hmm, maybe we should try with at least 50, so at least there's some point in running the test. :D Or leave it at 100 and observe how frequently this will cause failures for a while.

alexander-yakushev:
Changed it to 30. Unrelated CI failures are very annoying :). I think this should be enough to sanity-check.

PR #326, opened 2024-05-23
Add explicit opt-out for libnrepl agent

alexander-yakushev:
Context: https://clojurians.slack.com/archives/C0617A8PQ/p1716406878253329

I would think that it would be quite rare for the user to enable allowAttachSelf for unrelated reasons (but most probably to load other native agents at runtime) but bail out of this one. Still, this PR gives them the ability to do so.


alexander-yakushev:
@phronmophobic

bbatsov:
Should we also add some cmd-line flag for nREPL's CLI related to this?

PR #327, opened 2024-05-27
[interruptible-eval] Fix classloader chain growth (again)

alexander-yakushev:
I've noticed that when using nREPL, the classloader chain has been growing again, even if that problem has been fixed before (multiple times, it seems). See #8.

The reason why the chain is growing was laid out in great detail in https://github.com/clojure-emacs/cider/issues/3049, specifically in this comment: https://github.com/clojure-emacs/cider/issues/3049#issuecomment-920183844. The culprit is nREPL's usage of `clojure.main/repl` which should know better not to spam DCLs, but yet it does. `with-session-classloader` which is supposed to fix some things around classloaders would only do so when using the sideloader, and besides, `w-s-c` is used in the wrong place in interruptable-eval to fix this issue anyway.

Funnily enough, the classloader chain did not grow on Java 8 because of some fix for the unrelated issue (actually, semirelated. All classloader problems are related because they all are the hideous).

The suggested fix does really restore the context classloader (to @plexus's disgust) once `clojure.main/repl` finishes. But I've added a check to see if the final classloader is precisely one level deeper then the initial one and is a DCL. If it is, then we know this new classloader is the redundant one created by `c.m/repl`, discard it and restore the initial one. **Otherwise** it means that user code has installed a different classloader 😱, and we kinda have to roll with it.

Ideally, this should be fixed by dropping `c.m/repl` and using something that we actually control, like Chas originally wanted. But I don't know if it's good/safe enough to do that after 1.0.


bbatsov:
The suggested fix seems reasonable and simple (always a good thing) to me. 

> Ideally, this should be fixed by dropping c.m/repl and using something that we actually control, like Chas originally wanted. But I don't know if it's good/safe enough to do that after 1.0.

Well, given that swapping will be mostly transparent for users something like this is always in the cards. Potentially as something optional at first - e.g. our custom REPL would would have to be specified in the config. A custom implementation would also be handy for subrepls (https://github.com/nrepl/nrepl/issues/45)

alexander-yakushev:
Fixed tests. There was a flaky test that relied on very crudely parsing object identity hashcodes from a string. Instead, now the values are passed via a shared atom.

bbatsov (approved):
Everything looks good to me.

PR #328, opened 2024-06-03
[interruptible-eval] Improve error message if JVMTI agent cannot not be loaded

alexander-yakushev:
Because the user may see this error message without following the whole debacle, it makes sense to add some more context to lead the user towards the solution.

The only issue is that the link will 404 until nREPL docs are regenerated (which would probably happen with 1.2.0 release). Is this OK?

### DIFF
--- a/src/clojure/nrepl/middleware/session.clj
+++ b/src/clojure/nrepl/middleware/session.clj
@@ -176,6 +176,19 @@
 (defn- jvmti-stop-thread [t]
   ((misc/requiring-resolve 'nrepl.util.jvmti/stop-thread) t))
 
+(defn- try-stop-thread [^Thread t]
+  (cond
+    (<= misc/java-version 19) (.stop t)
+    ;; Since JDK20, Thread.stop() no longer works. We must resort to using
+    ;; JVMTI native agent which luckily still supports Stop Thread command.
+    ;; Whether this is more dangerous than calling Thread.stop() in earlier
+    ;; JDKs is unknown, but assume the worst and never use this if you can't
+    ;; take the risk!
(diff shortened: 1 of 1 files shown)


bbatsov:
> The only issue is that the link will 404 until nREPL docs are regenerated (which would probably happen with 1.2.0 release). Is this OK?

The link points to the `master` docs, so this will work fine when I regen them in a bit.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -176,6 +176,19 @@
...
+    (<= misc/java-version 19) (.stop t)
+    ;; Since JDK20, Thread.stop() no longer works. We must resort to using
+    ;; JVMTI native agent which luckily still supports Stop Thread command.
+    ;; Whether this is more dangerous than calling Thread.stop() in earlier
+    ;; JDKs is unknown, but assume the worst and never use this if you can't
+    ;; take the risk!
+    (misc/jvmti-agent-enabled?) (jvmti-stop-thread t)
+
+    (not (misc/attach-self-enabled?))
+    (misc/log "Cannot stop thread on JDK20+ without -Djdk.attach.allowAttachSelf enabled, see https://nrepl.org/nrepl/installation.html#jvmti.")))
```
bbatsov, line 190: It'd be good if we added something like https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/util/nrepl.clj#L8 to nREPL, so we can actually notify the clients about such things. Basically today we're missing a mechanism to push notifications to clients (e.g. warnings, deprecation messages, etc)

bbatsov:
I'll update the docs now.

PR #329, opened 2024-06-03
[docs] Minor fixes

alexander-yakushev:
Bringing JVMTI section one level up, so it appears in the TOC and thus is more visible.

### DIFF
--- a/doc/modules/ROOT/pages/design/overview.adoc
+++ b/doc/modules/ROOT/pages/design/overview.adoc
@@ -8,11 +8,7 @@ and asynchronous (in contrast to most REPLs that build on top of streams
 provided by e.g.  terminals).
 
 TIP: If you are interested in historical context, check the
-https://docs.google.com/document/edit?id=1dnb1ONTpK9ttO5W4thxiXkU5Ki89gK62anRqKEK4YZI&authkey=CMuszuMI&hl=en#[original
-design notes], https://github.com/clojure/tools.nrepl/wiki/nREPL.Next[nrepl
-0.2.0 notes], and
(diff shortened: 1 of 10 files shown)


load-file-test/nrepl/load_file_sample2.clj
```diff
@@ -1,3 +1,3 @@
 (ns nrepl.load-file-sample2)
 
-(Thread/sleep 10000)
+(Thread/sleep 20000)
```
bbatsov, line 3: Did you mean this for this branch or another one?

doc/modules/ROOT/pages/design/overview.adoc
```diff
@@ -8,11 +8,7 @@ and asynchronous (in contrast to most REPLs that build on top of streams
 provided by e.g.  terminals).
 
 TIP: If you are interested in historical context, check the
-https://docs.google.com/document/edit?id=1dnb1ONTpK9ttO5W4thxiXkU5Ki89gK62anRqKEK4YZI&authkey=CMuszuMI&hl=en#[original
-design notes], https://github.com/clojure/tools.nrepl/wiki/nREPL.Next[nrepl
-0.2.0 notes], and
-http://groups.google.com/group/clojure-dev/browse_frm/thread/6e366c1d0eaeec59[discussion]
```
bbatsov, line 14: I think this link is missing the in the additional resources, so it'd be nice to add it there

load-file-test/nrepl/load_file_sample2.clj
alexander-yakushev: I meant it for here because this test was failing every other push. _I should also push less frequently._

bbatsov: Got it. Push as often as you want.

Issue #333, opened 2024-06-04
State of sideloading and middleware injection?


alexander-yakushev:
What is the current stance on these features? Obviously, they are not going anywhere, but is there any current interest in them? Do any ecosystem tools started using those and depend on them? Is there will and interest to develop, improve, and eventually converge them with the default middleware setup? I'm trying to understand if these features were done and now fulfill their purpose or they were left in semi-baked experimental state and never took off.

bbatsov:
I know that only iced-vim is currently fully supporting sideloading at the client side. Not sure if they are actually using it for something. I still think that could be an useful way for editors to load some libraries or middleware on the fly, which would be nice for people using `cider-connect` or loading CIDER plugins on demand.

The way I see there are 2 main problems today:

1. It's kind of hard to use this, as you can't just instruct it to load some jar
2. Potentially `add-lib` can provide a simpler way to do this (as discussed here https://github.com/nrepl/nrepl/pull/246)

>  they were left in semi-baked experimental state and never took off.

That's the current state of affairs for sure. My initial intention was to make it easier for clients to use nREPL without having to rely on a lot of external middleware. E.g. they'd just sideload some Clojure library (e.g. for profiling) and eval code from it directly. Then we extended the scope to sideloading middleware, but never got both features to a state in which they were easy to use and somewhere there the work got stuck.

bbatsov:
Actually, this issue best summarizes the state of affairs https://github.com/clojure-emacs/cider/issues/3037 Basically, we were pretty close to making it possible to sideload CIDER's middleware, but didn't get to the finish line.

alexander-yakushev:
Thank you for the context!

I can't help but wonder if sideloading middleware is a tradeoff worth doing overall. I can clearly see the benefit for the user – not having to pre-specify the middleware dependencies like cider-nrepl and the middleware itself is useful. But for the developers, it seems to be adding a hell lot of extra complexity and points of failure. Currently, we are at least sure (and can check) that cider-nrepl dependency is on the classpath, and the real classpath, not some artificial emulation of it. If the wrong thing is loaded, again, this can be debugged using the usual methods that the user already knows from regular Clojure dependency management.

Speaking subjectively here, I as a user would prefer the current way of loading things, and maybe would appreciate an option to sideload for a remote prod REPL scenario. But I personally don't find value in sideloading for a local REPL that I have full control of.

[...]

PR #335, opened 2024-06-11
Remove support for sideloader and wrap-sideloader middleware

alexander-yakushev:
Please, consider this a proposal with an example attached.

After the discussion in #333 and on Slack, I came to the conclusion that the sideloader in its current state and implementation is not worth keeping for now. I considered the following observations:

1. Nobody came out to claim that they use the current sideloader functionality.
2. Of all public nREPL clients, only vim-iced has any support for current sideloading.
3. Many people claim that sideloading may indeed be nice for some usecases. However, people usually want the behavior provided by `add-lib` or Pomegranate, not class-by-class feeding.
4. Current sideloading approach has significant overhead and delay as each class (or Clojure file) is independently transferred from client to server on-demand.
5. Making the current sideloading work bears quite a lot of complexity on nREPL internals. Minus 200 LOC in `src/` from this PR is quite nice.
6. Sideloading has been marked as an experimental feature in the docs, so we are less committed to preserving it (in its current form).
7. If we want to resurrect sideloading functionality in future, it most likely will be done from scratch and will use simpler mechanisms akin to `add-lib`.
8. Finally, getting rid of it now simplifies improvements and refactoring that I envision for 1.3.


src/clojure/nrepl/misc.clj
```diff
@@ -78,30 +78,13 @@
...
-       (with-bindings {clojure.lang.Compiler/LOADER cl#}
-         ~@body)
-       (do
-         (.setContextClassLoader (Thread/currentThread) cl#)
-         (try
-           (with-bindings {clojure.lang.Compiler/LOADER cl#}
-             ~@body)
-           (finally
-             (.setContextClassLoader (Thread/currentThread) ctxcl#)))))))
+(defmacro bind-classloader
```
bbatsov, line 81: Why `bind-classloader` instead of `with-classloader`? I thought `with-*` names were more common when a macro has a body.

bbatsov:
> Sideloading has been marked as an experimental feature in the docs, so we are less committed to preserving it (in its current form).

Indeed. It was always implied that this might be removed if it doesn't work out, so no issues on this front.

Your proposal is fine by me and I'm good with removing the sideloader. Less is more. :-)

src/clojure/nrepl/misc.clj
alexander-yakushev: That's true. I'm pretty sure this macro will also be gone in the future. But anyway, I'll rename it to keep the convention for now.

bbatsov:
> BTW, I also considered removing dynamic-loader middleware (which allows to manipulate middleware – quite awkward naming) because without the sideloading requirement it can be replaced just with eval calls. But it can live for now, maybe we can turn it into something more useful.

Fair enough. It's also OK to change its name down the road - "dynamic-loader" can definitely a different thing to Java devs. :-)

Issue #334, opened 2024-06-04
Dynamic variables in nrepl.middleware.print are not working


alexander-yakushev:
Setting or binding dynamic variables `*print-fn*`, `*stream?*`, `*buffer-size*`, `*quota*` does not change the behavior of the `print` middleware. This happens because `interruptible-eval` and `print` middleware run on different threads, and thus changing the dynvars with `set!` within an eval command are not visible to the printer thread.

At this point, it is important to understand if this is worth fixing. Currently, `print` middleware allows customizing itself via the options in either request of response map. This is reliable and convenient to do for tooling authors that control the interaction with nREPL. Setting the parameters through dynvars was probably intended for end users who want to change the behavior of the printer directly from the REPL, bypassing the IDE/tooling . Do we want to support this usecase?

bbatsov:
> Setting the parameters through dynvars was probably intended for end users who want to change the behavior of the printer directly from the REPL, bypassing the IDE/tooling . Do we want to support this usecase?

Yeah, you're right about the intended use-case. Ideally we want to support it, as it'd be convenient for the end users. If that's not feasible (easily doable and not particularly complex) we should remove them, so we're not confusing people. //cc @cichli

alexander-yakushev:
If we want to support the end-user usecase, then I think it's better to provide explicit global setter functions like `set-print-fn`, `set-buffer-size`, etc., or a single `set-options`, thus dropping the illusion that this is achieved through dynamic variables and can be limited to a thread, a session, or something. Because making it work through dynamic variables would require a lot of gymnastics, and in the end, I doubt any users care if it is achieved through dynvars.

bbatsov:
Let's see first what @cichli will say about this given he introduced them https://github.com/nrepl/nrepl/commit/2ddf5361bdb740cb463c4679fbafb67565a70174 I'm wondering if this ever worked in the first place. If it didn't I'd probably just remove it. 

But I kind of like the idea of having some general purpose `set-options` function or something along those lines, so people would interactively explore the effect of certain options before changing the settings of their server or client.

alexander-yakushev:
This isn't the only place where dynamic variables do nothing in nREPL. For instance, the docs say that you can do:
```clj
(set! nrepl.middleware.lookup/*lookup-fn* my.namespace/my-lookup)
```
[...]

PR #336, opened 2024-06-13
Unify configuration

alexander-yakushev:
This is an attempt to make sense of the multitude of ways nrepl can be configured and fix a few tangential issues.

_Consider this is a WIP. There are plenty of things I'm not sure about, need feedback._

So, nREPL can be currently configured using the following (note that most options can be configured by only one or some of the below):

1. Global and local configuration files – which are EDN (data).
2. Command-line options which are strings – need to be parsed.
3. Sporadic dynamic variables which are meant to be `set!` at runtime, usually to resolved functions (vars). Some of those dynvars don't actually work when `set!` because they are not dynamically bound higher up in the stack (`*lookup-fn*`, `*complete-fn*`).
4. There are also dynvars that do work when `set!` but that requires some unobvious preparatory work from nrepl (see discussion in #334).
5. We now also have `:enable-jvmti-agent` which can only be set through config files.

I'd like to unify all those and make it more obvious how the user can change any of the options. The proposed solution upgrades `nrepl.config` to be more than a file reader. It is inspired by https://github.com/grammarly/omniconf (but greatly simplified). The config blueprint is first declared that lists all configuration values that nREPL supports, with types (and parsers) and default values. The blueprint then allows to parse the values from files, from env variables (with `NREPL_` prefix), Java properties (with `-Dnrepl.`prefix), and at runtime using `set-value` function.

Configuration values can be of simple data types (number, boolean), or a deferred resolved var. This is similar to how `nrepl.cmdline` works. The latter is now more tightly integrated `nrepl.config`. The user-supplied CLI args are transformed to config-like keys, merged with config, and then uniformly used throughout `nrepl.cmdline`.

Open questions:

1. Is this approach viable? Is it necessary?
2. Is it useful to be able to configure nREPL with env variables and properties? That part can be easily removed.
3. I wasn't too careful about not breaking compatibility for public functions in `nrepl.cmdline` and `nrepl.config` namespaces. Is compatibility there worth keeping?
4. I also removed `*lookup-fn*` and`*complete-fn*` vars but they were in experimental namespaces, and `set!`ing them by default throws an exception anyway.

### DIFF
--- a/src/clojure/nrepl/cmdline.clj
+++ b/src/clojure/nrepl/cmdline.clj
@@ -68,10 +68,8 @@
 (def running-repl (atom {:transport nil
                          :client nil}))
 
-(defn- done?
-  [input]
-  (some (partial = input)
-        ['exit 'quit '(exit) '(quit)]))
+(defn- done? [input]
+  (contains? #{'exit 'quit '(exit) '(quit)} input))
 
 (defn repl-intro
   "Returns nREPL interactive repl intro copy and version info as a new-line
@@ -121,14 +119,16 @@ Exit:      Control+D or (exit) or (quit)"
 (defn- run-repl
   ([{:keys [server options]}]
    (let [{:keys [host port socket] :or {host "127.0.0.1"}} server
-         {:keys [transport tls-keys-file tls-keys-str] :or {transport #'transport/bencode}} options]
+         {:keys [transport-fn tls-keys-file tls-keys-str]
+          :or {transport-fn #'transport/bencode}} options]
      (run-repl-with-transport
       (cond
         socket
-        (nrepl/connect :socket socket :transport-fn transport)
+        (nrepl/connect :socket socket :transport-fn transport-fn)
 
         (and host port)
-        (nrepl/connect :host host :port port :transport-fn transport :tls-keys-file tls-keys-file :tls-keys-str tls-keys-str)
+        (nrepl/connect :host host :port port :transport-fn transport-fn
+                       :tls-keys-file tls-keys-file :tls-keys-str tls-keys-str)
 
         :else
         (die "Must supply host/port or socket."))
@@ -169,10 +169,16 @@ Exit:      Control+D or (exit) or (quit)"
   (map (fn [arg] (or (option-shorthands arg) arg)) args))
 
 (defn- keywordize-options [options]
-  (reduce-kv
-   #(assoc %1 (keyword (str/replace-first %2 "--" "")) %3)
-   {}
-   options))
+  (let [options (reduce-kv
+                 #(assoc %1 (keyword (str/replace-first %2 "--" "")) %3)
+                 {}
+                 options)
+        ;; Rename CLI arg --transport canonical *-fn keys.
+        transport-fn (:transport options)
+        ack-port (:ack options)]
+    (cond-> (dissoc options :transport)
+      transport-fn (assoc :transport-fn transport-fn)
+      ack-port (assoc :ack-port ack-port))))
 
 (defn- split-args
   "Convert `args` into a map of options + a list of args.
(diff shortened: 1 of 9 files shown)


alexander-yakushev:
@cichli, do you think this can potentially replace dynvars in the print namespace? Note that all of this stands for global configuration, not a per-session one. However, I'm not sure how useful the fact that vars are changed per session.

bbatsov:
> I'd like to unify all those and make it more obvious how the user can change any of the options. The proposed solution upgrades nrepl.config to be more than a file reader. It is inspired by https://github.com/grammarly/omniconf (but greatly simplified). The config blueprint is first declared that lists all configuration values that nREPL supports, with types (and parsers) and default values. The blueprint then allows to parse the values from files, from env variables (with NREPL_ prefix), Java properties (with -Dnrepl.prefix), and at runtime using set-value function.

Cool approach! Admittedly I didn't think too much about the configuration when I introduced it (mostly 1) and 2)) and this shows today. 

> Is this approach viable? Is it necessary?

Seems viable to me. It's probably not necessary (few things ever are), but seems like an improvement to me. 

> Is it useful to be able to configure nREPL with env variables and properties? That part can be easily removed.

I definitely think this is useful.

> I wasn't too careful about not breaking compatibility for public functions in nrepl.cmdline and nrepl.config namespaces. Is compatibility there worth keeping?

I doubt anyone's using those directly, so probably that's fine. Just do some GitHub search to be on the safe side. 

> I also removed *lookup-fn* and*complete-fn* vars but they were in experimental namespaces, and set!ing them by default throws an exception anyway.

Fine by me, with the same advice as above - if it turns out they have known uses, we should probably deprecate them first for a smoother transition.

bbatsov:
> @cichli, do you think this can potentially replace dynvars in the print namespace? Note that all of this stands for global configuration, not a per-session one. However, I'm not sure how useful the fact that print options are changed per session.

Btw, #193 is kind of related to this as well, but it never got finished. I guess the work you're doing here will render the other PR obsolete and we can finally close it.

cichli:
I think it definitely makes sense to unify the config required for _starting the server_ like `port` / `bind` / and friends. I agree that supporting all four of CLI options / properties / env vars / config file is probably complete overkill – especially given that every project and build tool I've used that embeds nREPL just uses the Clojure API to start a server anyway – but it's what we have now.

What I don't think makes sense is complecting this server config with config that relates to _client expectations_ (i.e. `lookup-fn` / `complete-fn` / the print config you mention). I know that the majority use case for nREPL is for local development, where there is just a single client connecting to a local server, but it's still very much a supported use case (and one that _does_ happen in the wild) to have a remote server that accepts connections from multiple clients. In that case it would (to me) seem very strange to impose this kind of config on the server side.

Could we possibly just remove the global `lookup-fn` and `complete-fn` config entirely? Since they're new ops, any client that adds support for them could just support setting them in the request. The only reason we added the dynvars to the `print` and `caught` middleware is because those middleware enhance the eval op rather than add new ops, and we wanted them to still be configurable even for users of older clients.

cichli:
Having said that, it wouldn't exactly be the end of the world to have the `print` and `caught` config here too. We should just make abdundantly clear in the docs that these are the server-side defaults for client options, and that it's very much the expectation that modern clients are letting users set them.

If we removed the dynvars, the only users losing out would be those using old clients that don't support the new options (e.g. reply) to connect to remote REPLs, which is probably a small cross-section of nREPL users at this point.

alexander-yakushev:
Thank you for the detailed explanation! Indeed, I see your point about the distinction between server-side and client-side config. And the print and caught options being special because they change the behavior of the primary `eval` op also makes sense.

I'm not as keen on removing some dynamic variables as I try to fit the preferred mental model into my tiny brain. For example, it is possible to make `nrepl.config/set-config` only change the values within the session if it is executed from the `eval` op. And, on paper, changing per-session configuration sounds logical on a multi-user nREPL server. But how convenient is doing that on every re-connect? Does it follow the principle of least surprise?

> If we removed the dynvars, the only users losing out would be those using old clients that don't support the new options (e.g. reply) to connect to remote REPLs, which is probably a small cross-section of nREPL users at this point.

Yeah, it is something that bothers me a little about all this. Clients can already override the behavior by sending the correct response keys; server can (or will be able to) configure the defaults for those same options; and then, the user wants to be able to change these requests just for themself, not touching the server and circumventing the client. I suppose it is a valid usecasel; but just the sheer number of places where such a mere thing as a print function can be overridden from makes my head spin.

cursive-ide:
> I doubt anyone's using those directly, so probably that's fine. Just do some GitHub search to be on the safe side.

I use `nrepl.cmdline` in Cursive, so I'd be interested in the changes. That said, my use of it is very simple, I currently only use `--ack` and that's about to go away, so it'll just be bare `nrepl.cmdline` at that point.

alexander-yakushev:
> I use nrepl.cmdline in Cursive, so I'd be interested in the changes. That said, my use of it is very simple, I currently only use --ack and that's about to go away, so it'll just be bare nrepl.cmdline at that point.

The changes are only in the public functions that are not part of the "API" (if we assume that the API of `nrepl.cmdline` is `-main` and all of the supported configuration flags, those remain unchanged).

bbatsov:
> What I don't think makes sense is complecting this server config with config that relates to client expectations (i.e. lookup-fn / complete-fn / the print config you mention). I know that the majority use case for nREPL is for local development, where there is just a single client connecting to a local server, but it's still very much a supported use case (and one that does happen in the wild) to have a remote server that accepts connections from multiple clients. In that case it would (to me) seem very strange to impose this kind of config on the server side.

Those are fair points. For `lookup-fn` and `complete-fn` I recall that my idea was to make it possible to potentially auto-detect the presence of certain libraries (e.g. compliment and orchard) and have the middleware switch to using them, but I never did it (as I also thought this auto-upgrade might be a bit confusing). Today I'm thinking it's probably best to just making things simpler/uniform and only rely on client request for such type of configuration. I also thought those might be useful in the REPL for experimentation purposes, but I guess in reality no one ever used them at all. 

In #93, similarly, the idea was to be able to make the default printer configurable, in case someone just wanted to use that global always, but no one really asked for this ever, so in hindsight I think it was a mistake on my part to start thinking in this direction.

> It does make sense to remove the dynvars for lookup-fn and complete-fn – since they're new ops, any client that adds support for them could just support setting them in the request. The only reason we added the dynvars to the print and caught middleware is because those middleware enhance the eval op rather than add new ops, and we wanted them to still be configurable even for users of older clients.

That'd be fine by me - again those weren't really meant to be used by clients directly anyways, so it won't make much difference for them.

bbatsov:
One more thing - as noted in https://github.com/nrepl/nrepl/pull/66, much of my reasoning around the introduction of `.nrepl.edn` was to replace the need to have some server defaults in Lein or Boot's configuration, as they both had some custom solutions to do this. Similarly, `nrepl.cmdline`'s options were mostly driven by how people were using nREPL with `tools.deps`. 

Admittedly, I never stopped to think really carefully about what makes sense to be configured at the server-level vs at the request-level, which introduced some of the weirdness in the configuration design that we're discussing today. For a while I was thinking that some users might want to be able to centrally configure much of nREPL's behavior, so they can get similar experience without having to tweak client settings for several clients. As mentioned above - I don't recall users actually wanting to configure middleware via `.nrepl.edn` or command-line flags, so probably I was thinking in the wrong direction there. 

At the end of the day, most inconsistencies usually have the same source - different ideas from different stages of a project's evolution that were implemented without enough hammock time (consideration). I'm often struggling to remember what I was thinking and why I did something in a certain way years ago, and for the config it took me a couple of days to start remember (a bit of my reasoning) from the early days on maintaining nREPL.

alexander-yakushev:
Guess I'll have to backpedal on this one. It turned out to be more complicated than I anticipated. I'll restate the problem domain so that it is more clear what's going on in the future:

1. nREPL is a client/server application, so it (obviously) has a server and a client. We want certain aspects of the **server** behavior to be customizable.
2. The most straightforward, obvious, and utilized way to configure an nREPL server is to provide proper arguments to `nrepl.server/start-server` function. Somewhat similarly, `nrepl.cmdline` can be used to start the server, and equivalent command-line arguments fulfill the same role of configuring the server when started. `nrepl.cmdline` was mainly introduced as a REPL starter in tools.deps-based projects.
3. Leiningen ~and Boot~ (not relevant anymore) use `nrepl.server/start-server` but don't expose [enough] options for the **user** to change; thus, `nrepl.config` was introduced to configure the server bypassing the build tool black-box (and also, to give the user the ability to unify server config between different build tools).
4. Some **server** configuration options can be also changed on the **client** side. The most straightforward way to do that is by supplying appropriate keys together with the nREPL message.
5. For cases when the **client implementation** can't or won't pass the appropriate configuration keys, there is also an ability for **the user** to execute certain commands at the REPL (`set!` some variables) to change the behavior of the server only for the **current session**.

And this is my subjective feeling about this (numbers taken out of the butt):
- 95% of nREPL users either never explicitly configure anything in nREPL or they use cider-nrepl which overrides most of that configuration anyway with its custom handlers for exceptions, printing, lookup, etc.
- 90% of nREPL usecases have server/client collocated and started by the same process. 9% of usecases connect to a remote nREPL server; yet, the person connecting ot it has access to server configuration and can change it there, at server launch time. So, the percentage of cases where client-side user-initiated configuration is the preferred and best way to achieve the desired behavior is minimal.
- I have no idea how the user wanting to change some nREPL behavior for the first time is supposed to understand where the hell they should configure something. I've spent ~15 hours on this topic at this point and I can still barely wrap my head around this topic.

I'll go do something else for now since trying to simplify this while retaining the all possible scenarios and usecases feels like moving through a swamp 😄.

bbatsov:
Yeah, there's a lot going on here - that's for sure. It's often a good idea to step away from some tasks for a while and let your brain work on it in the background.

bbatsov:
You can also tackle this in smaller increments - e.g. first removing only problematic dynvars, refactoring a bit `nrepl.config` and `nrepl.cmdline` and we'll get to the full unification eventually step by step.

Issue #330, opened 2024-06-03
Replace usage of clojure.main/repl with custom REPL implementation


alexander-yakushev:
`clojure.main/repl` has been causing quite a few woes, especially with its classloader mismanagement (#327). Given that it is:

1. Pretty simple and short in its implementation to be adopted;
2. Contains logic that is fundamental to nREPL and is something that nREPL would like to control more;
3. Is already somewhat incompatible with what nREPL does (`clojure.main/repl` at its core has a streaming API, nREPL uses it to emulate message-oriented logic).

It makes sense to take over that code, repurpose and fit it better to our needs.

bbatsov:
That would be really great to address, as we had to work around quite a few issues with `clojure.main/repl`. You probably remember that Chas himself regretted the choice of `clojure.main/repl` on a few occasions. I'll add here related tickets & additional discussions (if I manage to recall where they happened).

alexander-yakushev:
Closed by #339.

PR #339, opened 2024-06-17
Introduce custom REPL implementation instead clojure.main/repl

alexander-yakushev:
See #330 for context.

The introduction of custom REPL is split into two commits:

1. The first one inlines `clojure.main/repl` into `nrepl.middleware.interruptible-eval/evaluate` almost as is.
2. The second commit refactors the inlined implementation to remove redundant parts and make it more idiomatic as a whole.

This approach makes it easier to see which exact changes we made to`clojure.main/repl`. There is only one meaningful change so far:
- Don't touch the context classloader, as it is already correctly managed by session middleware (see note below).

The PR is just the first step in rethinking the eval pipeline, yet it can be merged as an independent part.

**NOTE BELOW**

By looking at how interruptible-eval is implemented and how it is tied to `session` middleware, it seems like it was made  so that `interruptible-eval` can work without `session` in the middleware chain, albeit poorly. Is this compatibility mode still desired? The two middleware are very intertwined, and it makes sense to make one depend on the other in order to make further improvements.

**COMPATIBILITY NOTICE**

This PR should not break any public APIs of ours. However, some development tooling might rely on `clojure.main/repl` being on the stack to tell if the code is running in a REPL context. I know at least one place which does that. Anyway, relying on such things is expected to be unstable. Clojure 1.12 is about to introduce `*repl*` dynvar to make the notion of REPL more explicit, but until everyone upgrades to it in 10 years, I would expect tooling authors to be aware that the look of the stacktrace is not part of the API contract.

--- 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changes
 
 * [#335](https://github.com/nrepl/nrepl/pull/335): Remove support for sideloading and `wrap-sideloader` middleware.
+* [#339](https://github.com/nrepl/nrepl/pull/339): Introduce custom REPL implementation instead `clojure.main/repl`.
 
 ### Bugs fixed
 
--- a/doc/modules/ROOT/pages/design/middleware.adoc
+++ b/doc/modules/ROOT/pages/design/middleware.adoc
@@ -142,6 +142,17 @@ unhandled messages will always produce a response message with an `:unknown-op`
 `linearize-middleware-stack` to obtain a ordered sequence of middleware vars)
 should do the same, or use a similar alternative base handler.
 
+== Evaluation
+
+Core evaluation functionality is provided by
+`nrepl.middleware.interruptible-eval/interruptible-eval` middleware. It drives
+the REPL process using custom implementation instead of `clojure.main/repl`.
+When a message with `eval` op reaches this middleware, the forms in `:code` are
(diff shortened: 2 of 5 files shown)


bbatsov:
> By looking at how interruptible-eval is implemented and how it is tied to session middleware, it seems like it was made so that interruptible-eval can work without session in the middleware chain, albeit poorly. Is this compatibility mode still desired? The two middleware are very intertwined, and it makes sense to make one depend on the other in order to make further improvements.

I definitely don't remember any reasons for interruptible-eval to work without session. If I had to guess that's probably a remnant of nREPL's original design (the one that never made it to production).

bbatsov:
Overall, I'm good with the proposed changes. A bit of general feedback:

- Does it make sense to move the REPL to a separate ns/fn and potentially make this configurable? Probably make it configurable will be an overkill, as I doubt anyone would ever want to change this.
- Should we document this somewhere (e.g. in "Design")?
- You might want to add a few comments here and there in the REPL implementation, so someone won't change accidentally something important down the road (e.g. for the classloader optimization, etc)

@cichli Any feedback from you?

alexander-yakushev:
Rebased. Should we move forward with this? I've been using it (and other changes) for some time, no issues so far.

bbatsov:
Yeah, let's move forward with it.

PR #340, opened 2024-06-18
Use provided host in server-started-message

### LINKED ISSUE #302: nrepl wrongly prefers ip4 on dual stack systems
Starting nrepl with default params (no `-b`) launches it on `127.0.0.1`. It happily reports that it started/bound to `localhost:99999` . Other software (calva) connect via printed message and fails because `::1:99999` (ipv6) is not listened on.

(this started as https://github.com/BetterThanTomorrow/calva/issues/2310)

---

since nrepl binds to a port, it should print the address instead of the host  

## Expected behavior

banner text to print connectable address

`nREPL server started on port 36009 on host localhost - nrepl://127.0.0.1:36009`

or

`nREPL server started on port 36009 on host localhost - nrepl://::1:36009`

## Actual behavior

banner text to port not listed on

`nREPL server started on port 36009 on host localhost - nrepl://localhost:36009`

1. have a dualstack system (ip4 + ip6)
2. start nrepl (without `-b`)
3. nrepl starts ([hardbinds](https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/socket.clj#L37) to `127.0.0.1`)
4. nrepl prints `nrepl://localhost:36009`
5. calva/some software  wants to connect - resolves localhost - `::1`
6. and finally `nREPL connection failed: Error: connect ECONNREFUSED ::1:36009`

### Clojure version

1.11.1

### Java version

```
openjdk version "21" 2023-09-19
OpenJDK Runtime Environment (build 21+35)
OpenJDK 64-Bit Server VM (build 21+35, mixed mode, sharing)
```

### Operating system

arch linux

---

quickfix:

```diff
diff --git a/src/clojure/nrepl/socket.clj b/src/clojure/nrepl/socket.clj
--- src/clojure/nrepl/socket.clj
+++ src/clojure/nrepl/socket.clj
@@ -207,9 +207,9 @@
           (URI. (str transport-scheme
                      (when (instance? SSLServerSocket sock)
                        "s"))
                 nil ;; userInfo
-                (-> sock .getInetAddress .getHostName)
+                (-> sock .getInetAddress .getHostAddress)
                 (.getLocalPort sock)
                 nil       ;; path
                 nil       ;; query
                 nil)))))) ;; fragment
```

cichli:
Fixes #302.

I had to lift the logic that defaults `port` and `bind` from `inet-socket` up into `start-server` – otherwise they were nil on the server record. That does mean that they're now non-nil when `socket` is provided, but it shouldn't cause an issue as we always check `socket` first when it's relevant.

The last commit changes `as-nrepl-uri` to prefer `cond` over multiple nested `if-let`s; I recommend reviewing the commits one-by-one because of this.

### DIFF
--- a/src/clojure/nrepl/cmdline.clj
+++ b/src/clojure/nrepl/cmdline.clj
@@ -14,9 +14,7 @@
    [nrepl.server :as nrepl-server]
    [nrepl.socket :as socket]
    [nrepl.transport :as transport]
-   [nrepl.version :as version])
-  (:import
-   [java.net URI]))
+   [nrepl.version :as version]))
 
 (defn- clean-up-and-exit
   "Performs any necessary clean up and calls `(System/exit status)`."
@@ -440,16 +438,15 @@ Exit:      Control+D or (exit) or (quit)"
   connection details from.
   Takes nREPL server map and processed CLI options map.
   Returns connection header string."
-  [server options]
-  (let [transport (:transport options)
-        ^java.net.ServerSocket ssocket (:server-socket server)
-        ^URI uri (socket/as-nrepl-uri ssocket (transport/uri-scheme transport))]
-    ;; The format here is important, as some tools (e.g. CIDER) parse the string
-    ;; to extract from it the host and the port to connect to
-    (if-let [host (.getHost uri)]
+  [{:keys [host port socket] :as server}
+   {:keys [transport]}]
+  (let [uri (socket/as-nrepl-uri server (transport/uri-scheme transport))]
+    ;; The format here is important, as some tools (e.g. CIDER) parse the
+    ;; string to extract from it the host and the port to connect to
+    (if socket
+      (str "nREPL server started on socket " (.toASCIIString uri))
       (format "nREPL server started on port %d on host %s - %s"
-              (.getPort uri) host uri)
-      (str "nREPL server started on socket " (.toASCIIString uri)))))
+              port host uri))))
 
 (defn save-port-file
   "Writes a file relative to project classpath with port number so other tools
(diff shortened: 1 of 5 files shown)


bbatsov:
Looking at the CI failures it seems to me that IPv6 is probably not available on the test instances. Perhaps we can make such tests conditional?

bbatsov:
Also - should we document somewhere this behavior?

src/clojure/nrepl/cmdline.clj
```diff
@@ -440,16 +438,15 @@ Exit:      Control+D or (exit) or (quit)"
...
-    ;; The format here is important, as some tools (e.g. CIDER) parse the string
-    ;; to extract from it the host and the port to connect to
-    (if-let [host (.getHost uri)]
+  [{:keys [host port socket] :as server}
+   {:keys [transport]}]
+  (let [uri (socket/as-nrepl-uri server (transport/uri-scheme transport))]
+    ;; The format here is important, as some tools (e.g. CIDER) parse the
+    ;; string to extract from it the host and the port to connect to
+    (if socket
+      (str "nREPL server started on socket " (.toASCIIString uri))
```
bbatsov, line 447: Why is this preferable over `.getHost`? We want to avoid unicode hostnames or what?

bbatsov:
I took a closer look at the code and all of the changes look good to me. Feel free to merge, after you get the CI to pass.

alexander-yakushev:
I'll review it once the tests pass. When I was doing the configuration work, I had some issue with defaulting `port` to 0 early, but I don't quite remember what that was. Probably something in the tests.

bbatsov:
@cichli I plan to cut a new nREPL release early next week, so if you want this to be a part of it, you should try to wrap up the PR over the weekend.

bbatsov:
@cichli Can you update this PR?

bbatsov:
Superseded by #423, which has been merged.

PR #341, opened 2024-06-18
Make session middleware handle all necessary dynamic bindings

alexander-yakushev:
## The problem

The relation between `session` and `interruptible-eval` middleware is quite intertwined. The eval middleware binds REPL-relevant dynamic variables around the evaluated expression so that the evaluated expression can `set!` those. This made sense until `session` middleware started persisting these dynvars (to restore them after interrupts). To facilitate this, `interruptible-eval` performs some head-scratching hacks to let `session` mw see the changes made by the evaluated code. Some of the dynvars (e.g., `*in*`) are bound by `session`, some (e.g., `*out*`) are rebound with every new message (but still saved into the session); `*msg*` lives in `interruptible-eval`, is bound by `session`, is prevented from saving into the session atom by `interruptible-eval`. Messy and very confusing.

## The proposal

Let `session` handle all the dynvars required by `interruptible-eval` and the evaluated code. The `session` mw gathers all the dynvars, binds them in its `:exec` implementations, gets them back and persists (only in non-ephemeral sessions). This requires `session` mw to know a bit more about evaluation than it should, but it is still better than what we have now. The non-standard bindings like `*in*` and `*out*` are also bound by `session` but the mechanics of it is more explicit and apparent. The `interruptible-eval` mw becomes dumber and simply hopes that `session` mw sets up all the necessary dynvars.

I had to move some tests from sanity_check to core. Sanity check tests relied on calling `nrepl.middleware.interruptible-eval/evaluate` directly. I kept those tests but made them more e2endy.

---

### DIFF
--- a/src/clojure/nrepl/middleware/session.clj
+++ b/src/clojure/nrepl/middleware/session.clj
@@ -4,7 +4,9 @@
   (:require
    clojure.main
    [nrepl.middleware :refer [set-descriptor!]]
-   [nrepl.middleware.interruptible-eval :refer [*msg* evaluate]]
+   [nrepl.middleware.interruptible-eval :refer [*msg*]]
+   [nrepl.middleware.print :as print]
+   [nrepl.middleware.caught :as caught]
    [nrepl.misc :as misc :refer [noisy-future uuid response-for]]
    [nrepl.transport :as t])
   (:import
@@ -87,17 +89,6 @@
 
 (def default-executor "Delay containing the default Executor." (delay (configure-executor)))
 
-(defn default-exec
-  "Submits a task for execution using #'default-executor.
-   The submitted task is made of:
-   * an id (typically the message id),
-   * thunk, a Runnable, the task itself,
-   * ack, another Runnable, ran to notify of successful execution of thunk.
-   The thunk/ack split is meaningful for interruptible eval: only the thunk can be interrupted."
-  [_id ^Runnable thunk ^Runnable ack]
-  (let [^Runnable f #(do (.run thunk) (.run ack))]
-    (.submit ^ExecutorService @default-executor f)))
-
 (defn- session-in
   "Returns a LineNumberingPushbackReader suitable for binding to *in*.
    When something attempts to read from it, it will (if empty) send a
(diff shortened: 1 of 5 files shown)


bbatsov:
I think the proposal makes a lot of sense, although the solution is Clojure-specific as the eval for ClojureScript (in Piggieback) will still have to setup its own dynamic bindings. Still, I think for Clojure this would be beneficial overall - at the very least all middleware depending on `session` will see all the common dynamic bindings. 

At any rate - I'll defer to @cichli here, as he spent more time working on this part of the code.

bbatsov:
You might want to rebase this PR now that the custom REPL PR was merged. I'll put aside some time for a closer look here, but in general I don't see anything concerning in the implementation and I agree that's an improvement:

> The non-standard bindings like *in* and *out* are also bound by session but the mechanics of it is more explicit and apparent. The interruptible-eval mw becomes dumber and simply hopes that session mw sets up all the necessary dynvars.

bbatsov:
A cursory look at https://github.com/nrepl/piggieback/blob/master/src/cider/piggieback_impl.clj indicates minimal dependencies to ieval (just for `*msg*`) and none on `session`, so it seems the proposed changes won't affect Piggieback.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -216,60 +292,80 @@
...
+  - `:exec` - takes an id (typically a msg-id), a thunk and an ack runnables,
+    and an optional messsage (see `make-ephemeral-exec-fn` for ampler context).
+    Executions are serialized and occurs on a single thread.
+  - `:interrupt` - takes an id and tries to interrupt the matching execution
+    (submitted with :exec above). A nil id is meant to match the currently
+    running execution. The return value can be either: `:idle` (no running
+    execution), the interrupted id, or nil when the running id doesn't match the
+    id argument. Upon successful interruption the backing thread is replaced.
+  - `:close` - terminates the backing thread."
+  [session]
```
bbatsov, line 306: Why did you decide to switch from session id to session here?

alexander-yakushev: Because session is what contains all the saved dynamic bindings, and this function establishes those bindings for the code being run.

bbatsov: Yeah, I get this - I was mostly wondering if it might break some client calling `session-exec` directly for some reason, as it's public API. However, this seems fairly unlikely to me.

alexander-yakushev: Github search shows nothing. And public functions in the namespace that defines a middleware are not really part of the API, only the middleware itself is.

alexander-yakushev:
Rebased. No rush merging this, it's a bit intricate. Maybe @cichli still finds the time to take a look.

alexander-yakushev:
Updated the changelog. Let's make this happen, there are more PRs coming up.

bbatsov:
No arguments from me. :-) The other PRs looked quite interesting.

PR #343, opened 2024-06-20
Minor improvements

alexander-yakushev:
Nothing important here. Java versions are just major after 1.8, and the rest is not spec-declared, so it's no much sense to parse it out.


bbatsov:
Good improvements. I think I had copied that Java version parsing from Clojure, but I agree it's an overkill today.

Issue #332, opened 2024-06-03
Make threading usage more organized


alexander-yakushev:
Currently, nREPL is qiute haphazard about how it uses multi-threading. While the session middleware has a dedicated named threadpool and carefully manages session threads with it, in other places (reading/writing over the transport, killing stuck threads, etc.) a generic `noisy-future` function is used. This has two problems:

1. Putting that work on the Futures executor obscures what nREPL related-code is running/waiting and where.
2. By touching the Clojure futures, nREPL initializes their threadpool and potentially prevents the application from shutting down (since Future threads are non-daemon ones).

Ideally, we want all nREPL threading to be neatly subdivided into threadpools with clear separation what happens where.

bbatsov:
Yeah, I totally agree. The code there hasn't seen much changes since deciding to map evaluation threads to sessions (https://github.com/nrepl/nrepl/pull/98 and https://github.com/nrepl/nrepl/pull/127) and we definitely haven't thought much about the bigger picture.

alexander-yakushev:
Fixed in #345.

PR #345, opened 2024-07-09
Replace noisy-future with customized executors

alexander-yakushev:
Addresses #332.

This PR replaces relying on Clojure futures to perform nREPL internal asynchronous tasks with custom-branded executors. Different stages of the nREPL process get different threads. The futures pool is no longer touched, and thus nREPL doesn't anymore leave hanging non-daemon threads that can prevent the application from stopping normally.

This is what the result looks like:

image

Tangentially, this PR extracts classloader-related and thread-related code into dedicated namespaces out of `session.clj`. So the middleware namespace got a bit lighter.

This code is fresh, so I will be dogfooding it for a while.

---

### DIFF
--- a/src/clojure/nrepl/middleware/session.clj
+++ b/src/clojure/nrepl/middleware/session.clj
@@ -7,16 +7,14 @@
    [nrepl.middleware.interruptible-eval :refer [*msg*]]
    [nrepl.middleware.print :as print]
    [nrepl.middleware.caught :as caught]
-   [nrepl.misc :as misc :refer [noisy-future uuid response-for]]
-   [nrepl.transport :as t])
+   [nrepl.misc :as misc :refer [uuid response-for]]
+   [nrepl.transport :as t]
+   [nrepl.util.classloader :as classloader]
+   [nrepl.util.threading :as threading])
   (:import
-   (clojure.lang Compiler DynamicClassLoader LineNumberingPushbackReader)
+   (clojure.lang Compiler LineNumberingPushbackReader)
    (java.io Reader)
-   (java.util.concurrent.atomic AtomicLong)
-   (java.util.concurrent BlockingQueue LinkedBlockingQueue SynchronousQueue
-                         ExecutorService
-                         ThreadFactory ThreadPoolExecutor
-                         TimeUnit)
+   (java.util.concurrent Executors ExecutorService LinkedBlockingQueue)
    (nrepl SessionThread)))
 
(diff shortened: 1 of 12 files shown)


alexander-yakushev:
I get weird failures with 1.7 even though that part of the code is pretty much moved from nrepl.middleware.session to another place. Gotta investigate later.

alexander-yakushev:
Fixed the 1.7 issue (most probably it's a bug in 1.7, but the fix is not too demanding for this to be a burden). This is now ready for review.

alexander-yakushev:
Got it. Yeah, moving it to util sounds reasonable.

src/clojure/nrepl/classloader.clj
```diff
@@ -0,0 +1,25 @@
+(ns nrepl.classloader
+  "Creating and managing classloaders supplied to evaluated code."
+  (:import
+   clojure.lang.DynamicClassLoader))
+
+;; TODO : test
```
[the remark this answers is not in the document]
alexander-yakushev: Done!

src/clojure/nrepl/middleware/session.clj
```diff
@@ -3,20 +3,18 @@
   {:author "Chas Emerick"}
   (:require
    clojure.main
+   [nrepl.util.classloader :as classloader]
```
bbatsov, line 6: Don't forget to sort the namespaces after renaming them. I think the linter might complain about this as well.

src/clojure/nrepl/server.clj
```diff
@@ -10,8 +10,9 @@
    nrepl.middleware.load-file
    nrepl.middleware.lookup
    nrepl.middleware.session
-   [nrepl.misc :refer [log noisy-future response-for returning]]
+   [nrepl.misc :refer [log log-exceptions response-for returning]]
    [nrepl.socket :as socket :refer [inet-socket unix-server-socket]]
+   [nrepl.util.threading :as threading]
```
bbatsov, line 15: Ditto.

src/clojure/nrepl/util/classloader.clj
```diff
@@ -0,0 +1,27 @@
+(ns nrepl.util.classloader
```
bbatsov, line 1: You might mention in the new namespaces that they were introduced in nREPL (`:added`).

alexander-yakushev:
All done. Also tried to address the flaky interrupt-load test again.

Issue #331, opened 2024-06-03
Trim the common path (eval) stack


alexander-yakushev:
A bare no-extra-middleware stack of an evaluated expression in nREPL currently looks like this:

```
$  clojure -Sdeps '{:deps {nrepl/nrepl {:mvn/version "1.1.2"}}}' -m nrepl.cmdline -i                                                                         ~

user=> (Exception.)
#error {
 :cause nil
 :via
 [{:type java.lang.Exception
   :message nil
   :at [user$eval2608 invokeStatic "NO_SOURCE_FILE" 1]}]
 :trace
 [[user$eval2608 invokeStatic "NO_SOURCE_FILE" 1]
  [user$eval2608 invoke "NO_SOURCE_FILE" 1]
  [clojure.lang.Compiler eval "Compiler.java" 7534]
  [clojure.lang.Compiler eval "Compiler.java" 7489]
  [clojure.core$eval invokeStatic "core.clj" 3230]
  [clojure.core$eval invoke "core.clj" 3226]
  [nrepl.middleware.interruptible_eval$evaluate$fn__1477$fn__1478 invoke "interruptible_eval.clj" 87]
  [clojure.lang.AFn applyToHelper "AFn.java" 152]
  [clojure.lang.AFn applyTo "AFn.java" 144]
  [clojure.core$apply invokeStatic "core.clj" 667]
  [clojure.core$with_bindings_STAR_ invokeStatic "core.clj" 1990]
  [clojure.core$with_bindings_STAR_ doInvoke "core.clj" 1990]
  [clojure.lang.RestFn invoke "RestFn.java" 428]
  [nrepl.middleware.interruptible_eval$evaluate$fn__1477 invoke "interruptible_eval.clj" 87]
  [clojure.main$repl$read_eval_print__9216$fn__9219 invoke "main.clj" 437]
  [clojure.main$repl$read_eval_print__9216 invoke "main.clj" 437]
  [clojure.main$repl$fn__9225 invoke "main.clj" 459]
  [clojure.main$repl invokeStatic "main.clj" 459]
  [clojure.main$repl doInvoke "main.clj" 368]
  [clojure.lang.RestFn invoke "RestFn.java" 1526]
  [nrepl.middleware.interruptible_eval$evaluate invokeStatic "interruptible_eval.clj" 84]
  [nrepl.middleware.interruptible_eval$evaluate invoke "interruptible_eval.clj" 56]
[...]

Issue #196, opened 2020-06-17
Improve handling of boolean values in responses


bbatsov:
After some recent improvements to the encoding logic now an error is rightfully signalled when a response contains a boolean value. (sidenote - I wonder what was happening with them in the past, as I doubt we managed to filter them out everywhere). 

I've been wondering what to do about booleans in general:

a) nothing - leave it to middleware authors to sanitize the responses
b) convert them to strings (e.g. in bencode or with some middleware dedicated to sanitizing responses). I like the middleware approach more
c) disjoin them from responses, similar to the b) approach
d) come up with some extra type encoding info for bencode? (probably an overkill)
e) something else?

I was reminded about this mostly in the context of the lookup middleware and the fact that I missed some potential booleans that `meta` can return. I'll just add some form of handling this to the particular middleware, but I was wondering if we can do better in general. 

@shen-tian what do you think about this?

shen-tian:
It feels like this is related to the question of "what is the nREPL protocol?" that has come up a few times recently. In that context:

- Does the protocol contain a boolean type?
- If so, how is it encoded in bencode?

Right now, I'd say the answer is "No" and "n/a", but worth questioning. 

Previously, I think it just crashed the bencode transport, and some times silently.

bbatsov:
Yeah, probably you're right. Explicit certainly beats implicit. :-)

> Does the protocol contain a boolean type?

Currently it doesn't.

> If so, how is it encoded in bencode?

N/A

> Right now, I'd say the answer is "No" and "n/a", but worth questioning.

Seems we're like-minded here. :-)

That's not a big deal, but I guess I'll have to at least write some notes about client/extension authors - e.g. use strings instead of booleans if you really need them or something along those lines.

shen-tian:
I'd like to have a go at writing the core nREPL specification at some stage. Just a bit snowed under right now.

bbatsov:
And I'll add here years later that we obviously have a similar problem with requests, as it's hard to represent boolean flags without a boolean type in bencode.

PR #347, opened 2024-07-12
[print] Refactor print middleware

alexander-yakushev:
Several changes are made:

- Proxies that represented a writer that throws QuotaExceeded errors and the writer that calls underlying transport were rewritten to Java classes `QuotaWriter` and `CallbackWriter` respectively. They are bigger in LOC in Java compared to Clojure, but the handling of different arities is more explicit. Also, proxies are meh.
- Simplified and made more apparent the interaction between request and response print parameters, added more comments about what is happening.

The dynamic variables are kept in the working state, they are not going anywhere. @cichli, you may want to take a look.

---

### DIFF
--- a/src/clojure/nrepl/middleware/print.clj
+++ b/src/clojure/nrepl/middleware/print.clj
@@ -10,23 +10,15 @@
    [nrepl.transport :as transport])
   (:import
    (java.io BufferedWriter PrintWriter StringWriter Writer)
-   (nrepl QuotaExceeded)
+   (nrepl CallbackWriter QuotaBoundWriter QuotaExceeded)
    (nrepl.transport Transport)))
-
-;; private in clojure.core
-(defn- pr-on
-  [x w]
-  (if *print-dup*
-    (print-dup x w)
-    (print-method x w))
-  nil)
 
 (def ^:dynamic *print-fn*
   "Function to use for printing. Takes two arguments: `value`, the value to print,
   and `writer`, the `java.io.PrintWriter` to print on.
 
   Defaults to the equivalent of `clojure.core/pr`."
-  pr-on)
+  @#'clojure.core/pr-on) ;; Private in clojure.core
 
 (def ^:dynamic *stream?*
   "If logical true, the result of printing each value will be streamed to the
@@ -60,41 +52,14 @@
 (def configuration-keys
   [::print-fn ::stream? ::buffer-size ::quota ::keys])
 
-(defn- to-char-array
-  ^chars
-  [x]
-  (cond
-    (string? x) (.toCharArray ^String x)
-    (integer? x) (char-array [(char x)])
-    :else x))
-
-(defn with-quota-writer
+(defn with-quota-bound-writer
   "Returns a `java.io.Writer` that wraps `writer` and throws `QuotaExceeded` once
   it has written more than `quota` bytes."
   ^java.io.Writer
   [^Writer writer quota]
-  (if-not quota
-    writer
-    (let [total (volatile! 0)]
-      (proxy [Writer] []
-        (toString []
-          (.toString writer))
-        (write
-          ([x]
-           (let [cbuf (to-char-array x)]
-             (.write ^Writer this cbuf (int 0) (count cbuf))))
-          ([x off len]
-           (locking total
-             (let [cbuf (to-char-array x)
-                   rem (- quota @total)]
-               (vswap! total + len)
-               (.write writer cbuf ^int off ^int (min len rem))
-               (when (neg? (- rem len))
-                 (throw (QuotaExceeded.)))))))
-        (flush []
-          (.flush writer))
-        (close []
-          (.close writer))))))
+  (if quota
+    (QuotaBoundWriter. writer quota)
+    writer))
 
 (defn replying-PrintWriter
   "Returns a `java.io.PrintWriter` suitable for binding as `*out*` or `*err*`. All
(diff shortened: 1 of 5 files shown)


alexander-yakushev:
Also, the stack when an exception happens during printing has been shortened as discussed in #331.

```
user=> (map #(/ 1 %) (range)) *e
Error printing return value (ArithmeticException) at clojure.lang.Numbers/divide (Numbers.java:190).
Divide by zero
user=> #error {
 :cause "Divide by zero"
 :via
 [{:type clojure.lang.ExceptionInfo
   :message nil
   :data #:clojure.error{:phase :print-eval-result}
   :at [nrepl.middleware.interruptible_eval$evaluator$run__1445$fn__1456 invoke "interruptible_eval.clj" 118]}
  {:type java.lang.ArithmeticException
   :message "Divide by zero"
   :at [clojure.lang.Numbers divide "Numbers.java" 190]}]
 :trace
 [[clojure.lang.Numbers divide "Numbers.java" 190]
  [clojure.lang.Numbers divide "Numbers.java" 3891]
  [user$eval2508$fn__2509 invoke "NO_SOURCE_FILE" 1]
  [clojure.core$map$fn__5952 invoke "core.clj" 2772]
  [clojure.lang.LazySeq force "LazySeq.java" 50]
  [clojure.lang.LazySeq realize "LazySeq.java" 89]
  [clojure.lang.LazySeq seq "LazySeq.java" 106]
  [clojure.lang.RT seq "RT.java" 555]
  [clojure.core$seq__5484 invokeStatic "core.clj" 139]
  [clojure.core$print_sequential invokeStatic "core_print.clj" 53]
  [clojure.core$fn__7412 invokeStatic "core_print.clj" 174]
  [clojure.core$fn__7412 invoke "core_print.clj" 174]
  [clojure.lang.MultiFn invoke "MultiFn.java" 234]
  [clojure.core$pr_on invokeStatic "core.clj" 3700]
  [clojure.core$pr_on invoke "core.clj" 3694]
<-- nrepl ends here -->
  [nrepl.middleware.print$send_nonstreamed$fn__1366 invoke "print.clj" 114]
  [nrepl.middleware.print$send_nonstreamed invokeStatic "print.clj" 114]
  [nrepl.middleware.print$send_nonstreamed invoke "print.clj" 101]
  [nrepl.middleware.print$printing_transport$reify__1376 send "print.clj" 141]
  [nrepl.middleware.caught$caught_transport$reify__1412 send "caught.clj" 58]
  [nrepl.middleware.interruptible_eval$evaluator$run__1445$fn__1456 invoke "interruptible_eval.clj" 114]
  [nrepl.middleware.interruptible_eval$evaluator$run__1445 invoke "interruptible_eval.clj" 101]
  [nrepl.middleware.session$session_exec$session_loop__1538 invoke "session.clj" 281]
  [nrepl.SessionThread run "SessionThread.java" 21]]}
```

src/clojure/nrepl/middleware/print.clj
```diff
@@ -184,6 +143,12 @@
           (transport/send transport (misc/response-for msg resp))))
       print-var)))
 
+(defn- booleanize-bencode-val [m key]
+  (if (contains? m key)
+    ;; In bencode, empty list is treated as logical false.
```
bbatsov, line 148: I think this is more of a convention that we've established (because of Elisp) than something common with Bencode in general. I'm not sure that CIDER or any client actually does this, though - I'm fairly certain we simply don't send the keys when something's false on the client-side. Anyways, I understand your desire to preserve the current behavior, I'm just commenting, because this part might be a bit misleading for someone.

bbatsov:
The proposed refactoring looks good to me.

src/clojure/nrepl/middleware/print.clj
```diff
@@ -60,116 +52,83 @@
...
-        rf (completing
-            (fn [resp [key printed-value truncated?]]
-              (cond-> (assoc resp key printed-value)
-                truncated? (update ::truncated-keys (fnil conj []) key))))
-        resp (transduce (map print-key) rf resp keys)]
-    (transport/send transport (cond-> resp
-                                (::truncated-keys resp)
-                                (update :status #(set (conj % ::truncated)))))))
+   {:keys [::print-fn ::quota ::keys] :or {keys []}}]
+  ;; Iterator is used instead of reduce for cleaner stacktrace if an exceptions
```
bbatsov, line 97: exceptions -> exception

src/java/nrepl/QuotaWriter.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private final Writer wrappedWriter;
+    private final int quota;
+    private int remaining;
+    private final ReentrantLock lock = new ReentrantLock();
+
+    public QuotaWriter(Writer writer, int quotaBytes) {
+        super();
+        if (quotaBytes <= 0) {
+            throw new IllegalArgumentException("Bad quota: " + quotaBytes);
```
bbatsov, line 20: Perhaps "negative" or "invalid" would be a better term here.

src/java/nrepl/QuotaWriter.java
```diff
@@ -0,0 +1,94 @@
+package nrepl;
+
+import java.io.*;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Wrapper around <code>java.io.Writer</code> that throws
+ * <code>nrepl.QuotaExceeded</code> once it has written more than quota bytes.
+ */
+public class QuotaWriter extends Writer {
```
bbatsov, line 10: Or perhaps WriterWithQuota. Hard to find the optimal name for this one, but the current name will also do.

alexander-yakushev: QuotaBoundWriter?

bbatsov: Love this name!

src/clojure/nrepl/middleware/print.clj
alexander-yakushev: We could drop the current behavior if you are certain it's not used.

However, we have no way of representing `false` for cases when something is true by default, and the client wants to override that to false. Not sending the key then means preserving the default. I ran into this in cide-nrepl inspector middleware.

Alternatively, we can make it a convention to not use booleans in any APIs. That way, instead of `stream?` we would have something like `print-mode` that can be either "batch" or "stream".

bbatsov: Well, that's fairly big topic and we can follow-up on it separately. I remembered I wrote down some thoughts here https://github.com/nrepl/nrepl/issues/196 a long time ago.

alexander-yakushev:
All done.

PR #352, opened 2024-07-18
[test] Fix print middleware tests for Clojure 1.7-1.8

alexander-yakushev:
Not sure what's going on yet, but the logs point at a NPE when doing (.iterator keys). Maybe older Clojures have some problems with `:or` in destructuring.


alexander-yakushev:
Flaky season is upon us 😒.

PR #354, opened 2024-07-22
[session] Restore compatibility with middleware that passes 3 args to :exec

alexander-yakushev:
My refactoring of interruptible-eval and session middleware broke cider-nrepl middleware that relied on calling session's `:exec` function. While it is possible to fix things in cider-nrepl, then it would be trickier to keep it compatible with older and never nREPL versions. I rethought the implementation to preserve the backward compatibility here.

See https://github.com/clojure-emacs/cider/issues/3727.

We'll need a beta3 after merging this.

### DIFF
--- a/src/clojure/nrepl/middleware/session.clj
+++ b/src/clojure/nrepl/middleware/session.clj
@@ -7,7 +7,7 @@
    [nrepl.middleware.interruptible-eval :refer [*msg*]]
    [nrepl.middleware.print :as print]
    [nrepl.middleware.caught :as caught]
-   [nrepl.misc :as misc :refer [uuid response-for]]
+   [nrepl.misc :as misc :refer [uuid response-for log]]
    [nrepl.transport :as t]
    [nrepl.util.classloader :as classloader]
    [nrepl.util.threading :as threading])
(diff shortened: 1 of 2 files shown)


src/clojure/nrepl/middleware/session.clj
```diff
@@ -351,7 +350,8 @@
             "close" (close-session msg)
             "ls-sessions" (t/send transport (response-for msg :status :done
                                                           :sessions (or (keys @sessions) [])))
-            (h msg)))))))
+            (binding [*msg* msg]
```
bbatsov, line 353: I'd add a comment about this `binding`.

src/clojure/nrepl/middleware/session.clj
```diff
@@ -271,8 +266,12 @@
                         (reset-state)
                         running))))
      :close #(threading/interrupt-stop (:thread @state))
-     :exec (fn [exec-id r ack & [msg]]
-             (.put ^LinkedBlockingQueue (:queue @state) [exec-id r ack msg]))}))
+     :exec (fn [exec-id r ack]
+             ;; Here, *msg* is bound by session middleware on the server/handler
+             ;; thread. We have to convey it to the executor thread.
+             (if *msg*
```
bbatsov, line 272: Why do check that `*msg*` might be unbound?

bbatsov:
Overall the changes look fine apart from my small remarks. I preferred the previous implementation, but I agree that backward compatibility is more important in our case.

src/clojure/nrepl/middleware/session.clj
alexander-yakushev: Defensive programming here. It can't be unbound (I hope) in the current implementation, but if something changes in the future and inadvertently breaks this invariant, at least we'll get a reasonable error message instead of what happens now.

alexander-yakushev:
Added the comment. I agree that the new solution feels uglier and is somewhat less reliable (e.g., it will break if some middleware invokes `exec` lazily), but I really was too optimistic about changing this part of the API and hoping nothing breaks.

src/clojure/nrepl/middleware/session.clj
bbatsov: Got it.

bbatsov:
1.3.0-beta3 is out!

PR #355, opened 2024-07-24
[ci] Refactor CI config and improve caching

alexander-yakushev:
Various maintenance chores.

- Clean up project.clj a bit
- Refactor CI config, make caching more reliable, rename some jobs for shorter job names and easier visual parsing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,5 @@
 version: 2.1
 
-######################################################################
-#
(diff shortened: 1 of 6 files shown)


alexander-yakushev:
I also changed Clojure 1.12 from SNAPSHOT to LATEST because the snapshot repository is quite unreliable lately and causes tests to fail.

PR #356, opened 2024-07-29
Print all output in nrepl client

SevereOverfl0w:

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/clojure/nrepl/cmdline.clj
+++ b/src/clojure/nrepl/cmdline.clj
@@ -4,12 +4,13 @@
   https://github.com/trptcolin/reply/"
   {:author "Chas Emerick"}
   (:require
-   [clojure.java.io :as io]
    [clojure.edn :as edn]
+   [clojure.java.io :as io]
    [clojure.string :as str]
+   [nrepl.ack :refer [send-ack]]
    [nrepl.config :as config]
    [nrepl.core :as nrepl]
-   [nrepl.ack :refer [send-ack]]
+   [nrepl.misc :refer [uuid]]
    [nrepl.server :as nrepl-server]
    [nrepl.socket :as socket]
    [nrepl.transport :as transport]
@@ -93,15 +94,23 @@ Exit:      Control+D or (exit) or (quit)"
                      err    print
                      out    print
                      value  println}}]
-  (let [client (nrepl/client transport Long/MAX_VALUE)]
+  (let [client (nrepl/client transport Long/MAX_VALUE)
+        awaiting (atom {})]
     (println (repl-intro))
-    ;; We take 50ms to listen to any greeting messages, and display the value
-    ;; in the `:out` slot.
     (threading/run-with @threading/transport-executor
       ;; Doesn't matter which executor we hijack here, just don't use `future`.
-      (->> (client)
-           (take-while #(nil? (:id %)))
-           (run! #(when-let [msg (:out %)] (print msg)))))
+      (doseq [res (client)]
+        (some-> (:out res) out)
+        (some-> (:err res) err)
+        (when-let [p (and (some #{:done "done"} (:status res))
+                          (get @awaiting (:id res)))]
+          (deliver p true))
+        (when (some-> ^String (:id res)
+                      (.startsWith "nrepl.cmdline-"))
+          (some-> (:value res) value))
+        (flush)))
+    ;; We take 50ms to listen to any greeting messages, and display the value
+    ;; in the `:out` slot.
     (Thread/sleep 50)
     (let [session (nrepl/client-session client)]
       (swap! running-repl assoc :transport transport)
@@ -116,12 +125,13 @@ Exit:      Control+D or (exit) or (quit)"
               ;; Make sure the metadata read from *in* is preserved when we feed
               ;; the form to nREPL.
               (let [code-str (binding [*print-meta* true]
-                               (pr-str input))]
-                (doseq [res (nrepl/message session {:op "eval" :code code-str})]
-                  (some-> (:value res) value)
-                  (some-> (:out res) out)
-                  (some-> (:err res) err)
+                               (pr-str input))
+                    id (str "nrepl.cmdline-" (uuid))]
+                (swap! awaiting assoc id (promise))
+                (doseq [res (nrepl/message session {:op "eval" :code code-str :id id})]
                   (when (:ns res) (set! *ns* (create-ns (symbol (:ns res))))))
+                @(get @awaiting id)
+                (swap! awaiting dissoc id)
                 (recur)))))))))
 
 (defn- run-repl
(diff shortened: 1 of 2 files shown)


alexander-yakushev:
Hi, thanks for the fix! Overall, it seems reasonable. Do you think it is possible to combine the two `threading/run-with` into one? Just have a thread that prints stuff that arrives. Doesn't even have to look at the ID, there can be theoretically messages arriving that are not tied to any ID.

I'm also wondering what to do with the races. It seems like the `out` and the `value` can be printed in different order each time.

SevereOverfl0w:
I was pondering this yesterday as well.  I think we can't combine the two, because other clients could be issuing commands which produce output which shouldn't be mirrored to the command line client.  That `nil` id check is a bit of a hack for startup.

Regarding the racing, I was thinking that maybe instead of doing a `doseq` over the responses, instead everything is handled by the thread I start.  The tricky bit is that the REPL loop needs a signal that the message is "done" being printed.  That's a bit more difficult but should be solid in terms of races.

I'll have a noodle at that now.

alexander-yakushev:
> I was pondering this yesterday as well. I think we can't combine the two, because other clients could be issuing commands which produce output which shouldn't be mirrored to the command line client. That nil id check is a bit of a hack for startup.

I was under the impression that each `nrepl.core/client` will only see the messages from its own communication. Isn't it the case, @bbatsov? On the other hand, there doesn't seem to be anything but the message `id` to distinguish such things.

*CONT:*

I've just checked by starting to cmdline clients for a single server, and their output (and the messages received from the `(client)` don't intermingle. I suppose that's because each session gets its own `*out*` that forwards the output only to the session's own transport.

So, I think we should be quite safe listening and reacting to all messages on the client. What do you think, @bbatsov?

alexander-yakushev:
> Regarding the racing, I was thinking that maybe instead of doing a doseq over the responses, instead everything is handled by the thread I start.

~You wouldn't be able to do that directly because at some point the thread has to block on listening to user's input (the "read") but at the same time be ready to accept and display the asynchronously arriving print messages. It's possible to do that somehow with NIO selectors or something, but that would mean overcomplicating a quite simple function.~

You'll still need two threads but one will only `(read)` and send the messages to the server, and the other will dispatch all messages and communicate the changed `*ns*` back to the first one (it's probably the only reason they have to talk).

SevereOverfl0w:
Great, that simplifies things then.  I'll push something up now which does the coordinating threads which I think should be comprehensive in terms of printing _everything_ and coordinating the REPL loop with the thread.

src/clojure/nrepl/cmdline.clj
```diff
@@ -93,16 +95,22 @@ Exit:      Control+D or (exit) or (quit)"
...
-      ;; Doesn't matter which executor we hijack here, just don't use `future`.
-      (->> (client)
-           (take-while #(nil? (:id %)))
-           (run! #(when-let [msg (:out %)] (print msg)))))
-    (Thread/sleep 50)
+      (run!
+       (fn [res]
+         (some-> (:out res) out)
+         (some-> (:err res) err)
+         (when-let [p (and (seq (set/intersection #{:done "done"} (set (:status res))))
```
alexander-yakushev, line 106: Something like `(some #{:done "done} (:status res))` will be more compact.

src/clojure/nrepl/cmdline.clj
```diff
@@ -93,16 +95,22 @@ Exit:      Control+D or (exit) or (quit)"
...
     (println (repl-intro))
-    ;; We take 50ms to listen to any greeting messages, and display the value
-    ;; in the `:out` slot.
     (threading/run-with @threading/transport-executor
-      ;; Doesn't matter which executor we hijack here, just don't use `future`.
-      (->> (client)
-           (take-while #(nil? (:id %)))
-           (run! #(when-let [msg (:out %)] (print msg)))))
-    (Thread/sleep 50)
+      (run!
```
alexander-yakushev, line 102: `doseq` would save a few LOC over `run!` but that's up to you.

src/clojure/nrepl/cmdline.clj
```diff
@@ -93,16 +95,22 @@ Exit:      Control+D or (exit) or (quit)"
...
                      out    print
                      value  println}}]
-  (let [client (nrepl/client transport Long/MAX_VALUE)]
+  (let [client (nrepl/client transport Long/MAX_VALUE)
+        awaiting (atom {})]
     (println (repl-intro))
-    ;; We take 50ms to listen to any greeting messages, and display the value
-    ;; in the `:out` slot.
     (threading/run-with @threading/transport-executor
-      ;; Doesn't matter which executor we hijack here, just don't use `future`.
```
alexander-yakushev, line 101: I would prefer to preserve this comment.

alexander-yakushev (commented):
Looks good to me, left a few nits.

src/clojure/nrepl/cmdline.clj
```diff
@@ -93,16 +95,22 @@ Exit:      Control+D or (exit) or (quit)"
...
-    (Thread/sleep 50)
+      (run!
+       (fn [res]
+         (some-> (:out res) out)
+         (some-> (:err res) err)
+         (when-let [p (and (seq (set/intersection #{:done "done"} (set (:status res))))
+                           (get @awaiting (:id res)))]
+           (deliver p true))
+         (when (some-> (:id res)
+                       (str/starts-with? "nrepl.cmdline-"))
```
alexander-yakushev, line 110: No `starts-with?` in Clojure 1.7, gotta use interop.

src/clojure/nrepl/cmdline.clj
```diff
@@ -93,16 +94,21 @@ Exit:      Control+D or (exit) or (quit)"
...
+        awaiting (atom {})]
     (println (repl-intro))
-    ;; We take 50ms to listen to any greeting messages, and display the value
-    ;; in the `:out` slot.
     (threading/run-with @threading/transport-executor
       ;; Doesn't matter which executor we hijack here, just don't use `future`.
-      (->> (client)
-           (take-while #(nil? (:id %)))
-           (run! #(when-let [msg (:out %)] (print msg)))))
-    (Thread/sleep 50)
```
alexander-yakushev, line 105: I think this is still needed. We want the greeting to be printed before the `user=>` prompt, this delay increases the chances of it happening.

src/clojure/nrepl/cmdline.clj
```diff
@@ -93,16 +94,21 @@ Exit:      Control+D or (exit) or (quit)"
                      err    print
                      out    print
                      value  println}}]
-  (let [client (nrepl/client transport Long/MAX_VALUE)]
+  (let [client (nrepl/client transport Long/MAX_VALUE)
+        awaiting (atom {})]
     (println (repl-intro))
-    ;; We take 50ms to listen to any greeting messages, and display the value
-    ;; in the `:out` slot.
```
alexander-yakushev, line 99: Ditto.

src/clojure/nrepl/cmdline.clj
SevereOverfl0w: Ah, I missed that subtly.  Cool.

alexander-yakushev:
Awesome, thanks!

SevereOverfl0w:
Thank you for merging!  I'm running this locally quite actively, I'll keep an eye out for any oddities.

PR #357, opened 2024-08-06
doc: Remove non-nrepl compatible implementation

kristianlm:
Chicken Scheme's nrepl egg is not an nREPL protocol implementation, and is just an acronym for "networked repl".

I'm the author, sorry about the confusion.


bbatsov:
Today I learned! 😁

PR #359, opened 2024-09-06
Update beyond_clojure.adoc

nicolas-graves:
Changed link of recommended Python NREPL based on https://codeberg.org/sasanidas/nrepl-python/issues/1

Added Scilab NREPL server


alexander-yakushev:
Thank you!

PR #364, opened 2024-12-30
[session] Allow passing explicit msg to persistent session :exec

alexander-yakushev:
Resurrect an `:exec` arity that takes an explicit `msg`. It was briefly like that on the road to 1.3, then I scrapped it because it introduced incompatibility for middleware that still called `:exec` with three arguments. Ultimately, the correct way is to pass an explicit `msg`. But the PR keeps the old 3-arg version for compatibility and tries to do "good enough" by taking the initial `*msg*` in such case.

The other commit partly restores the old session management behavior that Piggieback relies on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#364](https://github.com/nrepl/nrepl/pull/364): Retain explicitly added session values.
+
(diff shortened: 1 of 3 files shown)


bbatsov (approved):
Both changes look good to me. You might want to mention them as bug-fixes in the changelog and I guess we can cut a bugfix release after they are merged.

PR #370, opened 2025-04-09
Add client info to clone op

ericdallo:
As discussed [here](https://clojurians.slack.com/archives/C17JYSA3H/p1744123473420459), this adds client info to clone op.

- [x] All tests are passing

Thanks!

manual][1] extremely useful.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## master (unreleased)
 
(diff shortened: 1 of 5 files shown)


bbatsov:
I guessing we should also consider whether to always copy this from existing sessions (even if it's not specified), but that's probably not an issue right now.

PR #371, opened 2025-04-13
Add missing fields to clone on new-session

ericdallo:

- [X] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,7 @@
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


bbatsov (approved):

bbatsov:
I'll give you a few more days to test this and then I'll cut a release when you feel confident that we're good to go.

ericdallo:
awesome, I opened this one because I realized cursive didn't call directly the op, but via nrepl, I was meaning to ask when we could release indeed, but I'm ok waiting a few more days, thank you!

PR #374, opened 2025-08-06
[config] Add support for dynamic var defaults

alexander-yakushev:
This solves several issues:

1. Finally gives a way to enforce `*warn-on-reflection*` in all projects by default. Only on the server side, not the client side, but I suppose clients can invent something for this too if they want. But this covers my personal needs by 95%.
2. Offers an opt-in solution for users to open certain dynvars for changing via `set!`. This is something similar to what is requested in https://github.com/nrepl/nrepl/discussions/373, albeit less flexible. Perhaps, we will need something in the future that can evaluate code too (the current implementation only accepts constants as defaults), but still this solution can go a long way.

Thoughts, suggestions.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Changes
 
-* [#370](https://github.com/nrepl/nrepl/pull/3370): Accept `:client-name` and `:client-version` in `clone` op.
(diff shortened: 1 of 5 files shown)


alexander-yakushev:
Been using this for a bit, works fine, gonna merge.

PR #375, opened 2025-08-27
[load-file] Refactor load-file middleware

alexander-yakushev:
Here's my take on simplifying take `load-file` op.

Currently, `load-file` worked by mashing together the code (source file content) into the call to `Compiler/load`. This caused issues when the code is huge, and JVM tried to embed the file contents as a string into the compiler method, and ran into its 64K method size limitation. So, `load-file` mw heroically solved that by having a global atom to save file content to, and later read content from it, and overall it looked ugly.

This is no longer necessary because the underlying `interruptible-eval` op now has a lot of knobs that allow us to pass code for it to run directly, not perform the Read-Eval loop, which we don't need here.

I think the only place where backward compatibility may be broken after this is if somebody supplies their own `eval` op that don't know how to deal with a sequence being passed as `:code` or doesn't handle custom `:eval` functions. But `eval` is really fundamental middleware to be messed with, and I also checked that `eval` knew how to work with sequences inside `
:code` forever, since Chas days, so I'd say it is obligatory for a substitiute to implement that too.

---


bbatsov (approved):
Solid changes as far as I'm concerned. I had totally forgotten about this, but I'm happy to see you decided to clean up the house a bit. (and I'm really curious what prompted you to tackle this :D )

PR #377, opened 2025-08-31
[middleware] Resolve dynamic variables from a user session instead of server context

alexander-yakushev:
This is a bug we had for a while. Setting things like `*print-fn*` from within the user session never worked. It worked for other dynvars for printing, e.g. `*stream?*` because that one is resolved during the execution of user code so that happens inside `interruptible-eval` after `session` has already established proper dynvar values. It is confusing indeed, but this is the solution.

---


src/clojure/nrepl/middleware.clj
```diff
@@ -178,3 +178,18 @@
...
+
+;; Registering dynvars that are used to configure middleware. This lives here
+;; and not in nrepl.middleware.session to avoid circular dependencies.
+
+(def explicit-per-session-dynvars
+  "Mapping of dynamic variables that always have to be established for a new
+  session to their initial values."
+  (atom {}))
+
+(defn register-per-session-dynvars
```
bbatsov, line 190: I'm wondering if it won't be a bit nice if we moved the dynvars registration into `set-descriptor`, so that we can just list the dynvars in the middleware descriptors and include them automatically in the middleware docs.

src/clojure/nrepl/middleware.clj
```diff
@@ -178,3 +178,18 @@
        extend-deps
        (topologically-sort comparator)
        (map :implemented-by)))
+
+;; Registering dynvars that are used to configure middleware. This lives here
+;; and not in nrepl.middleware.session to avoid circular dependencies.
+
+(def explicit-per-session-dynvars
```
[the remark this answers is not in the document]
alexander-yakushev: It's public because `nrepl.middleware.session` looks at it. Will making a public getter instead be any better?

src/clojure/nrepl/middleware.clj
alexander-yakushev: Would it be OK for you that `set-descriptor!` performs a mutable side-effect unrelated to the middleware object itself?

src/clojure/nrepl/middleware/session.clj
```diff
@@ -337,13 +335,18 @@
                            :session session-id})
         (t/send transport (response-for msg :status #{:done}))))))
 
-(defn- close-session
-  "Drops the session associated with the given message."
-  [{:keys [session transport] :as msg}]
+(defn close-session
+  "Drop the given session."
```
bbatsov, line 339: I know you kept this from the older docstring, but "Close" seems like a better work than "Drop" here.

PR #383, opened 2025-09-03
[middleware] Introduce safe-handle

alexander-yakushev:
This is analogous to `with-safe-transport` that we talked about, with some differences.
- I never liked the name `with-safe-transport` because it implies the transport itself is somehow made safe where it is just a glorified try-catch. Meaning that asynchronous errors on the transport are not really caught. I think `safe-handle` name is more appropriate.
- Obviously, we cannot do the stacktrace analysis magic inside nrepl because that is a cider-nrepl functionality. So all `safe-handle` does is it sends an error response over the transport containing the exception message and the status of the form `:<op>-error`. I didn't do any customization of status keywords, these errors should be exceedingly rare, and the bikeshedding here is not important.

There just a few spots where I managed to use this new macro, but there might be more in the future, so I think it is useful.

### DIFF
--- a/src/clojure/nrepl/transport.clj
+++ b/src/clojure/nrepl/transport.clj
@@ -252,3 +252,27 @@
   "Send a response for `msg` with `response-data` using message's transport."
   [msg & response-data]
   (send (:transport msg) (apply response-for msg response-data)))
+
+(defmacro safe-handle
+  "When given a message and a number of <op handler> pairs, invoke the handler
+  with the op that matches message's op. If an exception is raised during
+  handling, send an automatic error response through the message's transport
+  with `:<op>-error` status. Special keyword `:else` can be used for an op to
+  define a catch-all handler. Handlers should functions of 1 argument `msg`."
+  {:style/indent 1}
+  [msg & body] ;; `body` is used, otherwise CIDER acts up and misindents
+  (assert (even? (count body)))
+  (let [msg-sym (gensym "msg")
+        op-sym (gensym "op")]
+    `(let [~msg-sym ~msg
+           ~op-sym (:op ~msg-sym)]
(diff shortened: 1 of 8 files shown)


src/clojure/nrepl/transport.clj
```diff
@@ -252,3 +252,27 @@
   "Send a response for `msg` with `response-data` using message's transport."
   [msg & response-data]
   (send (:transport msg) (apply response-for msg response-data)))
+
+(defmacro safe-handle
```
bbatsov, line 256: Perhaps `with-safe-handlers` would be slightly better name for this macro, but I'm not totally certain about this.

bbatsov:
I'm fine with your proposal and I agree with the point that `with-safe-transport` is a misleading name.

I think it'd be good to mention this macro here https://nrepl.org/nrepl/building_middleware.html#best-practices, so that more people would discover it (and potentially other recent improvements, we have forgotten to document). That's also a reminder for me to spend a bit of time on the "Building X" docs.

src/clojure/nrepl/transport.clj
alexander-yakushev: I don't like `with-safe-handlers` here (or any `with-` for that matter) because it not only establishes some context, like `with-` macros do, but actually performs an action – invokes the appropriate handler.

alexander-yakushev:
It looks like they aren't stuck, but rather I did something wrong. I will redo everything again here, looks like it's not as trivial as it seemed.

alexander-yakushev:
@bbatsov Fixed. I was confused where the `t/respond-to` should go in this new scheme.

bbatsov (approved):
Seems solid to me!

PR #386, opened 2025-09-08
Add support for XDG_CONFIG_HOME

### LINKED ISSUE #384: Add support for XDG_CONFIG_HOME
The XDG Base directory layout is popular on Unix-like OSes and it'd be nice if we supported it, alongside our custom `NREPL_CONFIG_DIR` env variable. 

I'm thinking we should probably check in that order for:

- `NREPL_CONFIG_DIR`
- `XDG_CONFIG_HOME` (and append `nrepl` to it)

Otherwise we stick with the default `~/.nrepl`. Pretty easy change overall.

avitkauskas:
Closes #384.

The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest) says:
> If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.
so I also checked if `~/.config` exists (therefore the user is already using other tools adhering to XDG) and then use the `~/.config/nrepl` as well before defaulting to \~/.nrepl`.

Put it into the TIP in the documentation as well.

No tests :( as there were no tests for this code before as well.
Any chances for the first commit to nrepl? :)

### DIFF
--- a/src/clojure/nrepl/config.clj
+++ b/src/clojure/nrepl/config.clj
@@ -29,10 +29,9 @@
   or `nil` otherwise."
   [env-var]
   (let [value (System/getenv env-var)]
-    (when (and value (not (str/blank? value)))
-      value)))
+    (when (not (str/blank? value)) value)))
 
-(def config-file-name "nrepl.edn")
+(def ^:private config-file-name "nrepl.edn")
 
 (def config-file
   "nREPL's global configuration file.
--- a/src/clojure/nrepl/config.clj
+++ b/src/clojure/nrepl/config.clj
@@ -23,7 +23,7 @@
   "Default XDG config directory (~/.config)."
   (io/file home-dir ".config"))
 
-(defn- non-empty-env 
+(defn- non-empty-env
   "Get the value of the environment variable.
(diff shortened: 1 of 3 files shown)


bbatsov:
The PR looks good to me. Just add a changelog and I think we're good to go. We can add some tests verifying the resolution order down the road, but for something as simple as this we can live without them.

src/clojure/nrepl/config.clj
```diff
@@ -12,22 +12,52 @@
...
+
+(def ^:private config-file-name "nrepl.edn")
 
 (def config-file
-  "nREPL's config file."
-  (io/file config-dir "nrepl.edn"))
+  "nREPL's global configuration file.
+
+  The location is determined with the following precedence:
+   $NREPL_CONFIG_DIR/nrepl.edn
```
bbatsov, line 40: I'd probably prefix each option with `-` or `*` for readability.

alexander-yakushev:
Looks perfect now! Thanks and congrats on the first contribution!

avitkauskas:
Thank you!

Issue #182, opened 2020-04-19
Throw Invalid token when read ::auto/resolved keyword.


defclass:
## Expected behavior

should read ::auto/resolved keyword in tty transport.

## Actual behavior

Will throw  java.lang.RuntimeException: Invalid token on server side.

## Steps to reproduce the problem

Strart nrepl server with using tty.

```clj
(require '[nrepl.transport :as nrepl-t])

(nrepl/start-server :bind "127.0.0.1" :port 31416 :transport-fn nrepl-t/tty)
```

repl (`nc 127.0.0.1 31416`)

```clj
user=> (ns foo  (:require [clojure.java.io :as io]))
nil
foo=> ::io/abc
```

The server will throw a  "java.lang.RuntimeException: Invalid token" caused by  `clojure.core/read` in https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/transport.clj#L159.

By the way, `transport bencode` works fine when using `lein repl`.

also refer: 
https://clojure.atlassian.net/browse/CLJ-2359

## Environment & Version information

0.7.0

### Clojure version

1.10.1

### Java version

1.8

### Operating system

macos 10.13.6

bbatsov:
Thanks for the great issue report. @shen-tian Would you mind taking a look into this one?

shen-tian:
Can reproduce and explain the bug. A simpler error could be just to type `::abc` instead of `::io/abc`. We should expect `:foo/abc`, but get `:user/abc` in my case. Not too sure how to fix though:

https://github.com/nrepl/nrepl/blob/f97a23370ecebc976498b25da46cb69918b88b0b/src/clojure/nrepl/transport.clj#L147-L179

- Line 159 uses `read` to grab the next valid expression. This will auto resolve namespaced keywords. However, this is not ran in the correct namespace, it won't be done correctly. (i.e. `::io/abc` won't find the `io` alias in the `foo` ns).
- Line 165 updates the correct ns. So it fees like if we could just wrap the `read` call in the correct binding for `*ns*` to `@cns`, then this should work...
- Except `read` gets called right after the last expression being received, and before its execution and subsequent ns change.

A few thoughts:

- We can avoid calling `read` here, and just read till end of line or similar, and pass the string through to be read later. But that might be breaking some behaviours for clients that were counting on the "wait till a full expression" behaviour. 
[...]

PR #387, opened 2025-09-12
Resolve namespace aliased keywords in tty transport

frankleonrose:
- [x] All tests are passing
- [x] You've updated the [changelog](../blob/master/CHANGELOG.md) (that only applies to user-visible changes)

Addresses #182.

This may appear overwrought, but tests showed both changes were necessary.
- The `dummy-resolver` + `read+string` allows the fn to read the next form from input without being in a complete context.
  - I changed the `:read-cond` option to `:preserve` - in order to evaluate reader conditionals in the eval context.
- There is a race condition in `handle`, because the handle-executor is not serialized. The test added for keyword resolution would sometimes indicate out-of-order evaluation. The `tty` transport expects to have each message evaluated before the next one is returned from `read-msg` in order to correctly set `:ns` and `:session`. Adding the `read-sync` semaphore blocks `read-msg` until after the prior message is evaluated.
  - The done check to release the `read-sync` is modeled after [this code](https://github.com/nrepl/nrepl/blob/be6d82de1e9c2c67fc7513d471c5930d11458801/src/clojure/nrepl/cmdline.clj#L105) in `run-repl-with-transport`. The fact that that code _relies_ on receiving a `:done` status suggests that it is good enough here, too.

### DIFF
--- a/src/clojure/nrepl/transport.clj
+++ b/src/clojure/nrepl/transport.clj
@@ -176,10 +176,11 @@
 
 (defmacro read-form
   "Read a form from `in` stream."
-  [in]
+  [in sync-fn]
   (if clojure<1-10
     ;; Remains broken - remove when support for 1.8 & 1.9 is dropped
-    `[(read {:read-cond :allow} ~in)]
+    `(do (~sync-fn)  ;; Sync *prior* to read - read relies on eval
+         [(read {:read-cond :allow} ~in)])
     `(let [dummy-resolver# (reify clojure.lang.LispReader$Resolver
                              (currentNS    [_]             (gensym))
                              (resolveAlias [_ _alias-sym#] (gensym))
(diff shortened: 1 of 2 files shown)


frankleonrose:
Noticed this [comment](https://clojure.atlassian.net/browse/CLJ-2359?focusedCommentId=50541) in the linked Clojure issue. It is a problem with this implementation. Adding a test and need to resolve.

frankleonrose:
I note that the design docs say
> If some request generated multiple response messages only the final one would have the status attached to it.

That suggests that rather than looking for `{:status :done}`, I should instead look for `(contains? response :status)`.

In which case, is [this code](https://github.com/nrepl/nrepl/blob/be6d82de1e9c2c67fc7513d471c5930d11458801/src/clojure/nrepl/cmdline.clj#L105) in `run-repl-with-transport` also incorrectly looking for the last response to a message?

@alexander-yakushev, should I follow the docs or the code?

bbatsov (approved):

bbatsov:
Looks good to me. Just clean up a bit the commits, so only the relevant ones stay in the history and we're good to go.

alexander-yakushev:
Let's just squash, all commits relate to a single feature anyway.

PR #388, opened 2025-09-15
[ci] Implement autodeploys

alexander-yakushev:


bbatsov:
Guess my memory mislead me here. :D 

> "Hacking on nREPL" mentions that master SNAPSHOTs are autodeployed which is not true:).

Might have been true at some point. 😅

alexander-yakushev:
Updated, and implemented actual autodeploys because I just copied the chunk from Orchard first which uses Make instead of bb. Now it should be good. We'll see the next time we do the release.

PR #393, opened 2025-09-27
Implement output capturing for stdout/stderr

alexander-yakushev:
This PR is semi-ready – it needs more thought about the facade/API.

Premise: nREPL captures and forwards `*out*`/`*err*` streams to the client, but doesn't do the same for `System/out` and `System/err`. We have this forwarding functionality for the latter in cider-nrepl, in the form of [out](https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/out.clj) middleware. It works alright; however, I feel like such functionality should exist in base nREPL.

This PR implements such functionality in the form of `nrepl.util.out` namespace. Overall, it is inspired by cider-nrepl's out middleware, and is able to do the following:

- Capture original System/out and err.
- Wrap them into teeing streams that are able to both pipe the text to the original stream and also send it to the client using regular nREPL messages.
- Replace original streams with these wrapped streams.
- Replace root bindings for `*out*` and `*err*` with update ones that use the new wrapped streams. Hence, any output to the root `*out*` will be streamed to the client as well. This is unimportant within session context, on the nREPL-managed threads (where we take care of `*out*` and `*err*` already) but relevant in cases where e.g. a raw Java thread is started.

The question now is – how/when do we enable this behavior?

1. Out middleware in cider-nrepl does its work only after the client sends `out-subscribe` message to the server. In nREPL, we can also make a similar middleware (maybe even make it react to the same `out-subscribe` message). This will answer the question when do we perform the stream injection/replacement. However, it will also mean that the client has the power to perform such significant changes to the server (and replacing standard streams is a quite important change).
2. I'm not sure if the client really wants/needs this optionality in listening to stdout. Are there cases when it wouldn't want that? Why is it fine for the client to automatically listen to `*out*` but not to System/out? In any case, `cider-redirect-server-output-to-repl` is enabled by default in CIDER, and I don't know if anybody disables it.
3. If we want to enable this behavior on the server side, what is the best way to opt in/out of it? Like usual, an argument to the server, assisted by a config key?

After answering these questions, I can finish up the PR.

---

- [x] You've updated the docs

### DIFF
--- a/doc/modules/ROOT/pages/ops.adoc
+++ b/doc/modules/ROOT/pages/ops.adoc
@@ -109,7 +109,7 @@ Optional parameters::
 * `:column` The column number in [file] at which [code] starts.
 * `:eval` A fully-qualified symbol naming a var whose function value will be used to evaluate [code], instead of ``clojure.core/eval`` (the default).
 * `:file` The path to the file containing [code]. ``clojure.core/\*file*`` will be bound to this.
-* `:file-name` Name of source file, e.g. io.clj
+* `:file-name` Name of source file, e.g. io.clj. Do not confuse with ``file`` param which contains full path while ``file-name`` is only the short file name. Both parameters are needed to properly set the bytecode location metadata.
 * `:id` An opaque message ID that will be included in responses related to the evaluation, and which may be used to restrict the scope of a later "interrupt" operation.
 * `:line` The line number in [file] at which [code] starts.
 * `:ns` The namespace in which to perform the evaluation. The supplied namespace must exist already (e.g. be loaded). If no namespace is specified the evaluation falls back to ``\*ns*`` for the session in question.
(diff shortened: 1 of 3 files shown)


bbatsov:
> I'm not sure if the client really wants/needs this optionality in listening to stdout. Are there cases when it wouldn't want that? Why is it fine for the client to automatically listen to *out* but not to System/out? In any case, cider-redirect-server-output-to-repl is enabled by default in CIDER, and I don't know if anybody disables it.


There are 2 reasons why this is optional in cider-nrepl:

- one is that we weren't certain how this would work at first, so we wanted to be careful
- some people were used to logging output going in their `*nrepl-server*` buffer in CIDER and didn't like the logs appearing in their REPL buffers; I think that's kind of accidental and mostly alleviated by the newer log viewing functionality in CIDER, but something to be kept in mind

Personally, I think this should just be the default behavior as all output is the same IMO. 

> If we want to enable this behavior on the server side, what is the best way to opt in/out of it? Like usual, an argument to the server, assisted by a config key?

Yeah, probably this should be opt-out only on the server config level and we shouldn't bother with middleware-level options for it. I think this will add complexity with little value. 

It seems to me we have a pretty similar vision about the optimal interface for this.

PR #397, opened 2025-10-15
Update CHANGELOG

alexander-yakushev:


alexander-yakushev:
@bbatsov FYI – I forgot to update the changelog in the feature PR and this slipped past my announcement too.

PR #398, opened 2025-10-17
[print] Fix unicode handling in CallbackBufferedOutputStream

alexander-yakushev:
Addresses https://github.com/clojure-emacs/cider/issues/3844.

Of course, it was sloppy unicode handling.

---


bbatsov (approved):

alexander-yakushev:
Released 1.5.1 and pushed update to CIDER.

PR #400, opened 2025-11-14
Add support for naming the port file.

awkay:
Adds the ability to configure the port file name (default `.nrepl-port`) via CLI arguments, environment variables, config files, and programmatic API.

I find with new AI agents I often want a test REPL (which starts/stops things that would break my dev REPL) AND dev REPL, and I have MCP servers start both. I want to be able to connect to both of them via port files so I can fix problems the AI agents have, but the current library overwrites the file on the second REPL start. I'd actually like to go both directions, but in any case I need alternate port file names. I've written my own deps entry point, but I'd like to be able to configure this in Cursive, but when I talked with Colin Fleming of Cursive, he said he can't add support for that without nrepl library support because of middleware constraints, so I'm submitting this patch.

## Examples:

```bash
# CLI
clojure -M -m nrepl.cmdline --port-file my-port.txt

# Environment variable
export NREPL_PORT_FILE=my-port.txt

# Config file (.nrepl.edn)
{:port-file "my-port.txt"}

# Programmatic API
(server/start-server :port 7888 :port-file ".nrepl-port")
```

## Configuration Precedence

1. CLI argument (`--port-file`)
2. Environment variable (`NREPL_PORT_FILE`)
3. Config file
4. Default (`.nrepl-port` for CLI, `nil` for programmatic API)

## Backward Compatibility

✅ **Fully backward compatible** - No breaking changes:
- CLI: Creates `.nrepl-port` by default (unchanged)
- Programmatic API: No port file unless explicitly requested (unchanged)
- All changes are additive (optional parameters only)

## Bug Fix

Fixes bug where filesystem socket servers incorrectly wrote port files with `nil` content. Port file creation now explicitly skipped for filesystem sockets.

## Testing

All 19 tests passed covering:
- CLI default behavior and custom `--port-file` option
- Environment variable and config file support
- Configuration precedence (CLI > Env > Config)
- Programmatic API with various `:port-file` values
- Filesystem socket edge case
- Backward compatibility

### DIFF
--- a/src/clojure/nrepl/server.clj
+++ b/src/clojure/nrepl/server.clj
@@ -2,6 +2,7 @@
   "Default server implementations"
   {:author "Chas Emerick"}
   (:require
+   [clojure.java.io :as io]
    [nrepl.ack :as ack]
    [nrepl.middleware :as middleware]
    nrepl.middleware.completion
@@ -203,13 +204,18 @@
        would provide this greeting upon connecting to the server, but telnet et
        al. isn't that. See `nrepl.transport/tty-greeting` for an example of such
        a function.
+   * :port-file — if specified, the path to a file where the server port will be
+       written. The file will be deleted on JVM exit. Defaults to nil (no file
+       written). Commonly set to \".nrepl-port\" to support tooling that
+       discovers the server port by reading this file. Ignored when using
+       :socket (filesystem sockets).
 
    Returns a (record) handle to the server that is started, which may be stopped
    either via `stop-server`, (.close server), or automatically via `with-open`.
    The port that the server is open on is available in the :port slot of the
    server map (useful if the :port option is 0 or was left unspecified."
   ^nrepl.server.Server
-  [& {:keys [port bind socket tls? tls-keys-str tls-keys-file transport-fn handler ack-port greeting-fn consume-exception]
+  [& {:keys [port bind socket tls? tls-keys-str tls-keys-file transport-fn handler ack-port greeting-fn port-file consume-exception]
       :or {consume-exception (fn [_] nil)}}]
   (when (and socket (or port bind tls?))
     (let [msg "Cannot listen on both port and filesystem socket"]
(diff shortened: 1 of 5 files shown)


bbatsov:
> Fixes bug where filesystem socket servers incorrectly wrote port files with nil content. Port file creation now explicitly skipped for filesystem sockets.

Probably this should be a separate commit to be clearer it's separate from the main addition. 

I'm wondering how this will be used by clients like CIDER and Cursive, though, as it's not clear to make how will they be expected to find multiple port files in the same folder if they have different names.

awkay:
Ah, ok, so here's the specific Cursive/AI integration use-case: It's useful the *start* the REPL in IntelliJ, and in my projects I often start two (tests and dev as I mentioned, since integration tests can muck of the dev env). I want to *start* those in Cursive so that I have complete control on them, I can pre-seed them, etc. Whatever. Maybe I'm not even *using* an AI at the moment, but when I start the MCP, I want it to be able to connect to whichever REPL I have already running. I have my MCP configured to know about both kinds of REPLs "run your tests on the nREPL whose port is in .nrepl-test-port". But in Cursive it always overwrites the .nrepl-port file. So running two REPLs means it is impossible for the (later-started) AI agent to discover the right one...it could do some ps/lsof/bash madness to derive it, but all that mess seems really silly if Cursive could just have an input field "nREPL port file name".

Cursive already has a "Remote" nREPL connector that can be told the file to look for. So even if I reverse who starts the REPL I'd still want the AI to specify the different file names (which is how I actually currently do it, by writing a custom main entry point for nREPL, but then the AI is in control of starting/stopping the REPL). 

Screenshot 2025-11-25 at 12 14 58

All I'm saying is that if it can *consume* a custom port filename, why not let it *write* one? Then it is symmetric and covers real additional use cases.

src/clojure/nrepl/server.clj
```diff
@@ -203,13 +204,18 @@
        would provide this greeting upon connecting to the server, but telnet et
        al. isn't that. See `nrepl.transport/tty-greeting` for an example of such
        a function.
+   * :port-file — if specified, the path to a file where the server port will be
+       written. The file will be deleted on JVM exit. Defaults to nil (no file
```
bbatsov: Hmm, shouldn't the default be `.nrepl-port`? Otherwise we'd be changing the default behavior unless I'm missing something.

PR #420, opened 2026-02-07
Clean up project.clj

alexander-yakushev:
- Remove mentions of Fastlane in the code – nobody uses it. Let's reconsider adding Transit transport to nREPL itself if desirable.
- Remove `lein test-all` because it wasn't doing what it claimed (it doesn't test against all Clojure versions). Let CI deal with that.
- Bump a bunch of deps.
- Rename `maint` profile/command to `docs` as that name has always been confusing me (obviously it stands for "maintenance" but I parsed it as "main" all the time).


bbatsov:
I would have probably done 4 commits (1 for each change), but I'm fine with all the proposals in general.

bbatsov:
> Rename maint profile/command to docs as that name has always been confusing me (obviously it stands for "maintenance" but I parsed it as "main" all the time

Yeah, totally get it. Back in the day the idea was that there would be multiple tasks available in this profile, but this never happened. 

Thanks for the cleanup!

PR #423, opened 2026-02-23
Use actual bind address in server startup message

### LINKED ISSUE #302: nrepl wrongly prefers ip4 on dual stack systems

bbatsov:
On dual-stack systems (IPv4 + IPv6), nREPL binds to `127.0.0.1` but prints `nrepl://localhost:PORT` in the startup banner because `as-nrepl-uri` calls `.getHostName` on the server socket. Clients like Calva resolve `localhost` to `::1` (IPv6), fail to connect, and report "connection refused".

This threads the explicit bind address through the Server record so the URI uses exactly what was specified (defaulting to `127.0.0.1`), rather than reverse-resolving the socket address.

Fixes #302, based on #340.

---

### DIFF
--- a/src/clojure/nrepl/cmdline.clj
+++ b/src/clojure/nrepl/cmdline.clj
@@ -16,8 +16,7 @@
    [nrepl.util.threading :as threading]
    [nrepl.version :as version])
   (:import
-   [java.io FileNotFoundException]
-   [java.net URI]))
+   [java.io FileNotFoundException]))
 
 (defn- clean-up-and-exit
   "Performs any necessary clean up and calls `(System/exit status)`."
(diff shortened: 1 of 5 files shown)


src/clojure/nrepl/cmdline.clj
```diff
@@ -454,16 +453,15 @@ Exit:      Control+D or (exit) or (quit)"
   connection details from.
   Takes nREPL server map and processed CLI options map.
   Returns connection header string."
-  [server options]
+  [{:keys [host port socket] :as server} options]
   (let [transport (:transport options)
-        ^java.net.ServerSocket ssocket (:server-socket server)
-        ^URI uri (socket/as-nrepl-uri ssocket (transport/uri-scheme transport))]
```
alexander-yakushev: Why remove the type hint from here and move it to the callsite? I think it was fine as is.

alexander-yakushev (approved):
Looks reasonable, with a single nitpick.

PR #426, opened 2026-03-11
Drop support for Clojure 1.8

alexander-yakushev:
Let's pull the plug? 1.8 was released precisely 10 years ago. Some things become simpler operationally if we do this.

---


bbatsov (approved):
I think we can probably drop support for Clojure 1.9 as well at this point, but let's take things a step at a time.

🚀

PR #435, opened 2026-05-12
Re-enable testing on Windows

alexander-yakushev:


bbatsov:
Still broken it seems. 😅

alexander-yakushev:
Yeah, consider it WIP for testing purposes for now. I'm doing some test refactoring too.

alexander-yakushev:
@bbatsov Tests pass now. What I did is I disabled all TLS tests on Windows. Maybe someday somebody will figure it out, but for now, it is better to test Windows without them than not to test anything at all.

bbatsov (approved):

PR #436, opened 2026-05-12
Fix site build command

Wilfred:
There isn't an antora flag called --pull as far as I can see.


bbatsov:
Good catch. Seems the option was renamed a while ago https://gitlab.com/antora/antora/-/blob/main/CHANGELOG.adoc#L1318

PR #441, opened 2026-05-25
Follow-up fixes for the tools.deps migration

bbatsov:
Small follow-ups to the tools.deps migration commit:

- Fix typos and broken grammar in `deps.edn` and the hacking guide.
- Update the middleware doc and the `version.clj` docstring that still referenced the old Leiningen-based doc generation.
- Drop the dead `error-msg` fn left behind when the CLI `-main` entrypoint was removed from `build/docs.clj`.
- Fix the `deploy` Make target, which passed a mangled version (Make was expanding `$PROJECT_VERSION` and the `export` ran in a separate shell from the `clojure` call).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ jobs:
           command: |
             java -version; if(-not $?){exit 9}
             clojure -T:build javac :with-tests true; if(-not $?){exit 9}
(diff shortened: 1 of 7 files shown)


alexander-yakushev:
Thanks!

PR #451, opened 2026-07-09
Allow configuring the default printer via config file

### LINKED ISSUE #93: Make the default printer configurable

bbatsov:
The print function could only be set per request or with `set!` inside a session, so there was no way to set a default for a whole server, and a client that doesn't expose the print options couldn't reach it at all.

The configured printer becomes the root value of `*print-fn*` when the server starts, rather than being consulted per request. That way the existing precedence holds by construction: request beats session `set!` beats configured default, with no extra bookkeeping in `wrap-print`.

Config keys are `:nrepl.middleware.print/print` and `:nrepl.middleware.print/options`, matching the request keys, which also keeps them clear of third-party middleware config.

Fixes #93.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#93](https://github.com/nrepl/nrepl/issues/93): Allow configuring the default print function via the `:printer` and `:printer-options` configuration keys.
+
 ### Changes
 
 * [#284](https://github.com/nrepl/nrepl/issues/284): Include the `-f`/`--repl-fn` option in the command-line help output.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### New features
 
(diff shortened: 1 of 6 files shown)


bbatsov:
@alexander-yakushev What you're take on this? 

I'm on the fence as it was discussed for a while in the past, but it's also not something particularly important.

alexander-yakushev:
I'd rather hold this up. We have so many ways to configure the damn printer that we drown in it already, and adding another one is something you can't walk back in the future.
I think it should be possible to configure this via default dynamic variables in a config, but It need to check. For now, let's wait.

bbatsov:
Fine by me.

PR #459, opened 2026-07-13
Send built-in client input to the server as raw text

### LINKED ISSUE #182: Throw Invalid token when read ::auto/resolved keyword.

bbatsov:
The interactive built-in client read each form with its own reader and re-printed it before sending it on, which made the *client's* reader the first to see your code. Anything it couldn't handle took the whole REPL down: a stray `)`, a custom tagged literal, or an auto-resolved keyword like `::io/foo` whose alias only exists in the server's session (#182 - deceptively worked under `--interactive` since client and server share a JVM there).

Now the client reads input only to find where each top-level form ends, captures the exact source text with `read+string` and sends it to the server verbatim, so the server's reader is the only one that interprets the code. The boundary read is defused: `*suppress-read*` holds record and tagged literals inert, a permissive `*reader-resolver*` accepts unknown aliases and classes, and reader conditionals are preserved. `*read-eval*` is off too - `*suppress-read*` doesn't cover `#=`, so otherwise the reader evaluates it client-side and the server then evaluates the same text again. Reader errors print and discard the rest of the line like `clojure.main` instead of crashing, and a closed input stream ends the loop cleanly.

Fixes #182.

### DIFF
--- a/src/java/nrepl/in/CapturingPushbackReader.java
+++ b/src/java/nrepl/in/CapturingPushbackReader.java
@@ -1,0 +1,76 @@
+package nrepl.in;
+
+import java.io.IOException;
+import java.io.PushbackReader;
+import java.io.Reader;
+
+/**
+ * A PushbackReader that records every character it reads into an internal
+ * buffer (removing it again on unread), so that the exact text consumed by a
+ * read can be recovered afterwards via {@link #getCaptured()}. Used by the
+ * built-in command-line client to find a form's boundary while keeping the
+ * original input text to send to the server verbatim.
+ *
+ * <p>Only single-character {@code read()}/{@code unread(int)} are capture-aware
+ * (that is all Clojure's LispReader uses). The bulk {@code read(char[], ...)}
+ * methods route through {@code read()} so they stay consistent, but the bulk
+ * {@code unread(char[])} methods are unsupported rather than silently
+ * corrupting the capture.
+ */
+public class CapturingPushbackReader extends PushbackReader {
+    private final StringBuilder captured = new StringBuilder();
+
+    public CapturingPushbackReader(Reader in) {
+        super(in);
+    }
+
+    /** Returns the text consumed since the last {@link #clearCaptured()}. */
+    public String getCaptured() {
+        return captured.toString();
+    }
+
+    /** Discards the captured text, e.g. before reading the next form. */
+    public void clearCaptured() {
+        captured.setLength(0);
+    }
+
+    @Override
+    public int read() throws IOException {
+        int c = super.read();
+        if (c != -1) {
+            captured.append((char) c);
+        }
+        return c;
+    }
+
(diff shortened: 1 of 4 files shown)


src/java/nrepl/in/CapturingPushbackReader.java
```diff
@@ -0,0 +1,76 @@
...
+ * built-in command-line client to find a form's boundary while keeping the
+ * original input text to send to the server verbatim.
+ *
+ * <p>Only single-character {@code read()}/{@code unread(int)} are capture-aware
+ * (that is all Clojure's LispReader uses). The bulk {@code read(char[], ...)}
+ * methods route through {@code read()} so they stay consistent, but the bulk
+ * {@code unread(char[])} methods are unsupported rather than silently
+ * corrupting the capture.
+ */
+public class CapturingPushbackReader extends PushbackReader {
```
alexander-yakushev, line 20: Can't we use `LineNumberingPushbackReader`? It can already capture.

bbatsov: `LineNumberingPushbackReader.captureString()` only landed in Clojure 1.10 (it's what backs `read+string`) - it's not there in 1.9, which we still support. That's the same reason I didn't just reach for `read+string`. Perhaps we should start considering targeting Clojure 1.10?

alexander-yakushev: Ah, right.

I see two options – either drop 1.9 completely (it's a narrow slice anyway), or version-gate the functionality for 1.10+ only. I guess the former is reasonable.
image

alexander-yakushev: So, LineNumberingPushbackReader is back on the menu now.

bbatsov: Indeed.

PR #461, opened 2026-07-14
Consume stdin responses explicitly in the request-*in* tests

bbatsov:
Following #460, the rest of the `request-*in*` tests still had the same shape that made `request-multiple-read-newline` flaky: the stdin `session` send was buried inside a lazy `for` over the response seq that `response-values` only realized later, so the send fired at a nondeterministic point during realization.

This rewrites the remaining four the same way #460 did the first: force the first response, assert it's `need-input`, send stdin, then force and assert the value. No behavior change, just deterministic consumption with the side effect out of the lazy seq.

Verified locally across both transports, including full-namespace runs.

- [ ] You've updated the [changelog](../CHANGELOG.md) (only applies to user-visible changes)

### DIFF
--- a/test/clojure/nrepl/core_test.clj
+++ b/test/clojure/nrepl/core_test.clj
@@ -643,22 +643,20 @@
     (is (thrown? SocketException (repl-eval session "(+ 1 1)")))))
 
 (def-repl-test request-*in*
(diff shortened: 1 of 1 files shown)


alexander-yakushev:
Makes sense! But the windows part is acting out, hm.

alexander-yakushev (approved):

PR #464, opened 2026-08-24
Use contextClassLoader in session middleware

plexus:
Partially reverts commit d3e1d1a2acafe5423599485f8dd2310a3687bbed "Minor code aesthetics", because it breaks use of custom classloaders like `lambdaisland.classpath/priority-classloader`.

Given the commit message I suspect this semantic change was not intentional. As a result of that change hot reloading of dependencies in lambdaisland.classpath and Launchpad has been broken since nrepl 1.6.0-alpha3.

- [ ] You've added tests to cover your changes.


alexander-yakushev:
Discussed in https://clojurians.slack.com/archives/C17JYSA3H/p1787554459814919

bbatsov:
I guess we can also do a bug-fix release to ship this, right?

plexus:
that would be great, then I can bump the dependency in launchpad and we're full circle

PR #466, opened 2026-08-26
Retire nrepl.misc/requiring-resolve

bbatsov:
Ours predates the one Clojure grew in 1.10, and the only thing it added was a `log?` arity nothing ever used - paid for by shadowing core's in every namespace that needed it. Call sites now use core's directly.

Worth knowing they aren't drop-in equivalents: ours returned nil for anything it couldn't resolve, core's throws on a missing namespace or an unqualified symbol. Where the symbol is one of ours a failure is a bug and should surface, so those call it bare. Where it comes from the client (`::print`, `::caught`, `lookup-fn`, `complete-fn`) the existing error and fallback paths depend on nil, so those keep it - `print` and `caught` already had tests pinning that, `lookup-fn`/`complete-fn` didn't and do now.

Side benefit: loading goes through `serialized-require`, so sessions resolving vars concurrently can't race on `require`.

`building_middleware.adoc` used it in an example, so that's updated too. Also drops a `clojure.stacktrace` require in `nrepl.tls` that nothing used.

Based on #465, since the changelog entries belong in the sections that adds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changes
 
 - [#462](https://github.com/nrepl/nrepl/pull/462): **(Breaking)** Raise minimal supported Clojure version to 1.10.
+- [#466](https://github.com/nrepl/nrepl/pull/466): **(Breaking)** Remove `nrepl.misc/requiring-resolve` in favour of `clojure.core/requiring-resolve`.
 - [#144](https://github.com/nrepl/nrepl/issues/144): Document more middleware development best practices.
 - [#257](https://github.com/nrepl/nrepl/issues/257): Document the `lookup` op's return values.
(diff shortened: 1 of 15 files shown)


alexander-yakushev:
I say just drop it altogether and replace callsites with core's requiring-resolve. The `log?` arity was never used anywhere.

Maybe leave `nrepl.misc/requiring-resolve` as an anemic passthrough shim, but that is also optional, I presume. Whoever happened to use that (= nobody) can jump to core's version just the same.

bbatsov:
Yeah, I think that's fair and probably here we shouldn't care about strict backwards compatibility. Probably at some point we should draw a stricter line between internal utility functions and actual public APIs. I'll update this.

alexander-yakushev (approved):

PR #469, opened 2026-08-28
Don't ask for stdin that is already arriving

bbatsov:
Patched twice on the test side (#452, #460), but the bug is in the server.

`add-stdin` enqueues stdin one character at a time via `LinkedBlockingQueue.addAll`, which isn't atomic, so a reader can find the queue empty mid-payload and ask for input it's already been sent. Clients answer that, and the unrequested input gets picked up by the next read - the stray `:ohai` from #452. The enqueue now takes a lock and `request-input` re-checks under it.

Couldn't get it to repro end-to-end on demand; driving the queue and `request-input` directly gives a spurious `need-input` in ~2% of runs, none with the lock. No test changes on purpose - the existing tests assert exactly one `need-input`, which is the invariant that was broken.


alexander-yakushev:
Yeah, I'm separately researching this at the moment. Going to come up with a solution too.

alexander-yakushev:
Overruled by #470.

PR #470, opened 2026-08-28
[session] Make adding stdin input to queue atomic

alexander-yakushev:
Alternative solution to #469.

The issue is indeed with a race between the producer adding characters one-by-one with non-atomic `.addAll` and the consumer that blocks halfway into reading the input.

I don't like using a lock as a solution. Even if it won't deadlock right now, the shape of the code may change sometime in the future where it might. It's also not immediately obvious to me what the scope of the lock should be. Also, it is important to make sure that the consumer holding the lock doesn't block the producer from proceeding.

The proposed solution reverts to using a queue as a single source of truth (and which already handles concurrency correctly). Whole strings are put onto the queue. Then, a string is fed out char-by-char to the consumer. A small added benefit is that we don't copy the string into sequences of chars or even char arrays, very minor, but still.

Spent a lot of time trying to simplify `*skipping-eol*` thing, didn't work the way I wanted because of the damn outer PushbackReader. Tried making QPR extend PushbackReader – bleh, need to override many unrelated methods, not too pretty.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - [#458](https://github.com/nrepl/nrepl/pull/458): Report descriptive errors for invalid TLS key material (missing, legacy-format or encrypted private keys, missing certificates) instead of NPEs and generic JSSE messages.
 - [#182](https://github.com/nrepl/nrepl/issues/182): The built-in client now sends input to the server as raw text instead of reading it client-side, so reader typos, auto-resolved keywords relying on session aliases (e.g. `::io/foo`) and custom tagged literals no longer crash the REPL.
 - [#468](https://github.com/nrepl/nrepl/pull/468): Fix `nrepl.spec` key types for the `describe` and `ls-sessions` responses, which never matched what the server sends.
+- [#470](https://github.com/nrepl/nrepl/pull/470): [session] Fix race condition between stdin consumer and producer.
 
 ## 1.7.0 (2026-04-14)
 
--- a/src/clojure/nrepl/middleware/interruptible_eval.clj
+++ b/src/clojure/nrepl/middleware/interruptible_eval.clj
@@ -54,6 +54,18 @@
   (try (.getName (io/file source-path))
        (catch Exception _)))
 
+(defn- skip-stdin-newline []
+  ;; NB: confusing hack. When `(read)` is issued to an input stream, it
+  ;; returns a Lisp form, ending with the last character of the form,
+  ;; naturally (e.g. a closing paren). However, it is expected that any
+  ;; trailing newline after such form is thrown away in order for a
+  ;; subsequent `(read-line)` call to not just return that empty newline but
+  ;; the actual following content. To simulate this behavior, we run
+  ;; this newline-skipping function after every `eval` request.
+  ;; Note that we check if stdin is ready in order not to block if it is empty.
+  (when (.ready ^PushbackReader *in*)
+    (clojure.main/skip-if-eol *in*)))
+
 (defn evaluator
   "Return a closure that evaluates msg's `:code` (either a string or a seq of
   forms to be evaluated) within the dynamic context of its session. `:ns` can be
@@ -155,6 +167,7 @@
           (caught e))
         (finally
           (flush)
+          (skip-stdin-newline)
           (pop-thread-bindings))))))
 
 (defn interruptible-eval
--- a/src/clojure/nrepl/middleware/session.clj
+++ b/src/clojure/nrepl/middleware/session.clj
@@ -2,7 +2,6 @@
   "Support for persistent, cross-connection REPL sessions."
   {:author "Chas Emerick"}
   (:require
-   clojure.main
    [nrepl.config :refer [config]]
    [nrepl.middleware :refer [set-descriptor!]]
    [nrepl.middleware.interruptible-eval :refer [*msg*]]
@@ -37,32 +36,20 @@
 ;; depending upon the expectations of the client/user.  I'm not sure at the moment
 ;; how best to make it configurable though...
 
-(def ^:dynamic ^:private *skipping-eol* false)
-
 (def ephemeral-executor
   "Executor for running eval requests in ephemeral sessions."
   (delay (Executors/newCachedThreadPool
           (threading/thread-factory "nREPL-ephemeral-session-%d"))))
 
 (defn- session-in
-  "Returns a LineNumberingPushbackReader suitable for binding to *in*.
-   When something attempts to read from it, it will (if empty) send a
-   {:status :need-input} message on the provided transport so the client/user
-   can provide content to be read."
+  "Returns a QueuePollingReader (that is suitable for binding to *in* when
+  wrapped into LineNumberingPushbackReader). When something attempts to read
+  from it, it will (if empty) send a {:status :need-input} message on the
+  provided transport so the client/user can provide content to be read."
   [session-id transport]
-  (let [input-queue (LinkedBlockingQueue.)
-        request-input #(or (.poll input-queue)
-                           ;; Skipping newlines shouldn't trigger need-input
-                           ;; from the client if queue is empty.
-                           (when-not *skipping-eol*
-                             (t/send transport
-                                     (response-for *msg* :session session-id
-                                                   :status :need-input))
-                             (.take input-queue)))
-        reader (LineNumberingPushbackReader.
-                (QueuePollingReader. input-queue request-input))]
-    {:input-queue input-queue
-     :stdin-reader reader}))
+  (QueuePollingReader. #(t/send transport
+                                (response-for *msg* :session session-id
+                                              :status :need-input))))
 
 (let [state (atom {})]
   (defn- dynvar-defaults-from-config []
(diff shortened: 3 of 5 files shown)


bbatsov:
This hangs the test suite for me. `nrepl.core-test/ensure-no-*msg*-capture` never finishes (I killed it at 90s, 420s and 600s); master and #469 both pass it.

jstack, on `nREPL-handler-2`:

```
LinkedBlockingQueue.take
nrepl.in.QueuePollingReader.readInputChar(QueuePollingReader.java:45)
nrepl.in.QueuePollingReader.read(QueuePollingReader.java:100)
BufferedReader.fill / LineNumberReader.read / PushbackReader.read
clojure.main$skip_if_eol(main.clj:115)
nrepl.middleware.session$add_stdin$fn(session.clj:403)
```

The newline skip blocks in `take()`, and `readInputChar` fires `sendNeedInputRequest` before it blocks, which is exactly what `*skipping-eol*` was there to prevent.

It stops working because the flag went from a dynamic var to an instance field, and the skip doesn't run on the thread doing the reading:

```
skip: nREPL-handler-1
eval: nREPL-session-e9ec1604-...
```

So `newlineSkippingMode`, `currentInput` and `currentInputIndex` are now non-volatile state shared across two threads. Aside from the hang, a session thread that reads while the handler holds the flag gets `-1`, a silent EOF in the middle of live input.

I do like the queue-of-strings part better than the lock in #469, it removes the partially visible state instead of guarding it. It's the skip flag that needs to stay thread-local.

A couple of the Copilot comments are off, by the way: `withNewlineSkipping` already has the `finally`, and `read` already checks `len == 0` first.

bbatsov:
Here's what I ended up with locally. Fixes the hang, and the suite is green over seven full runs.

The hang is the skipping flag. `ensure-no-*msg*-capture` fires two evals at one session, so two handler threads sit in `withNewlineSkipping` at once, and the first one's `finally` clears the flag while the second is still inside `skip-if-eol`. That one then takes the blocking branch, sends `need-input` and parks in `take()`. Making it a `ThreadLocal` gets back what `*skipping-eol*` had as a dynamic var.

The rest follows from that: `pollInputChar` never blocks and is synchronized, `readInputChar` does the send and the `take()` outside the monitor, so `currentInput`/`currentInputIndex` are guarded and nobody holds the lock while waiting for input. The skip and the drain loop each become a single synchronized operation so they can't interleave on the index.

Separate bug I hit while in there: the drain loop swallowed the EOF marker, so `"abc"` followed by EOF lost the EOF and the next read blocked instead of returning -1. The marker is now only consumed when `read` actually reports EOF, which also lets the restore in `skipLeadingNewlines` go away.

The monitor here is private to the reader, guards its own two fields and is never held across a blocking call, so I don't think it has the scope problem you flagged on #469.

```diff
diff --git a/src/java/nrepl/in/QueuePollingReader.java b/src/java/nrepl/in/QueuePollingReader.java
index 32e1301..8236ae5 100644
--- a/src/java/nrepl/in/QueuePollingReader.java
+++ b/src/java/nrepl/in/QueuePollingReader.java
@@ -2,75 +2,118 @@ package nrepl.in;
 
 import clojure.lang.IFn;
 import java.io.*;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.LinkedBlockingDeque;
 
 /**
- * An implementation of Reader which, when read is attempted, pulls characters
- * out of a wrapped LinkedBlockingQueue. Not thread-safe! Should only be
- * interacted with by a single thread.
+ * An implementation of Reader which, when read is attempted, pulls strings out
+ * of a wrapped blocking queue. Input is added by the thread that handles
+ * "stdin" requests and consumed by the thread that runs the evaluation, so all
+ * access to the buffered input is guarded by the reader's monitor.
  */
 public class QueuePollingReader extends Reader {
 
     private static final String EOF_MARKER = new String();
 
     private final IFn sendNeedInputRequest;
-    private final LinkedBlockingQueue<String> queue;
+    private final LinkedBlockingDeque<String> queue;
 
+    // Guarded by this.
     private String currentInput = null;
     private int currentInputIndex = 0;
-    private boolean newlineSkippingMode = false;
+
+    // Newline skipping runs on the thread that handles the "eval" request,
+    // while the evaluation itself reads *in* on the session thread, and two
+    // requests for the same session can be handled at once. The mode must not
+    // leak between those threads, hence a thread-local and not a plain field.
+    private final ThreadLocal<Boolean> newlineSkippingMode =
+        ThreadLocal.withInitial(() -> Boolean.FALSE);
 
     public QueuePollingReader(IFn sendNeedInputRequest) {
-        this.queue = new LinkedBlockingQueue<>();
+        this.queue = new LinkedBlockingDeque<>();
         this.sendNeedInputRequest = sendNeedInputRequest;
     }
 
-    private int readInputChar(boolean blockOnEmpty) {
+    /**
+     * Return the next character of buffered input, -1 if there is none right
+     * now, or -2 if the end of input has been reached. The end-of-input signal
+     * stays put until readEof() takes it, so that peeking at it (while skipping
+     * newlines, or while draining what has already arrived) doesn't drop it.
+     */
+    private synchronized int pollInputChar() {
         while (true) {
-            if (currentInput == EOF_MARKER) {
-                currentInput = null;
-                return -2;
-            }
+            if (currentInput == EOF_MARKER) return -2;
+
+            if (currentInput != null && currentInputIndex < currentInput.length())
+                return currentInput.charAt(currentInputIndex++);
 
-            if (currentInput != null && currentInputIndex < currentInput.length()) {
-                char c = currentInput.charAt(currentInputIndex++);
-                return c;
-            }
             currentInputIndex = 0;
             currentInput = queue.poll();
-            if (currentInput != null) continue;
-            if (blockOnEmpty) {
-                sendNeedInputRequest.invoke();
-                try {
-                    currentInput = queue.take();
-                } catch (InterruptedException ex) {
-                    return -1;
-                }
-            } else return -1;
+            if (currentInput == null) return -1;
+        }
+    }
+
+    /** Consume the end-of-input signal, so that later input can be read again. */
+    private synchronized void readEof() {
+        if (currentInput == EOF_MARKER) currentInput = null;
+    }
+
+    /** Make input taken off the queue the current buffer, if it is exhausted. */
+    private synchronized boolean bufferInput(String input) {
+        if (currentInput == null
+            || (currentInput != EOF_MARKER && currentInputIndex >= currentInput.length())) {
+            currentInput = input;
+            currentInputIndex = 0;
+            return true;
         }
+        return false;
     }
 
-    private void skipLeadingNewlines() {
-        // Only skip newlines for the input that we have already obtained
+    /** Read one character, asking the client for input and waiting if needed. */
+    private int readInputChar() {
         while (true) {
-            int c = readInputChar(false);
-            if (c == -1)
-                return;
-            else if (c == -2) {
-                // Restore EOF marker for future read.
-                currentInput = EOF_MARKER;
-                return;
-            } else if (c != '\n') {
+            int c = pollInputChar();
+            if (c != -1) return c;
+
+            sendNeedInputRequest.invoke();
+            String input;
+            try {
+                input = queue.take();
+            } catch (InterruptedException ex) {
+                return -1;
+            }
+            // Another thread may have refilled the buffer while we waited;
+            // put what we took back at the front and let the next round of
+            // polling drain the buffer before reaching it again.
+            if (!bufferInput(input)) queue.addFirst(input);
+        }
+    }
+
+    private synchronized void skipLeadingNewlines() {
+        // Only skip newlines for the input that we have already obtained.
+        while (true) {
+            int c = pollInputChar();
+            if (c < 0) return;
+            if (c != '\n') {
                 currentInputIndex--; // Unread non-newline character
                 return;
             }
         }
     }
 
+    private synchronized int readBufferedChars(char[] buf, int off, int len, int from) {
+        int i = from;
+        while (i < len) {
+            int c = pollInputChar();
+            if (c < 0) break;
+            buf[off + i++] = (char)c;
+        }
+        return i;
+    }
+
     public void withNewlineSkipping(Runnable f) {
-        newlineSkippingMode = true;
+        newlineSkippingMode.set(Boolean.TRUE);
         try { f.run(); }
-        finally { newlineSkippingMode = false; }
+        finally { newlineSkippingMode.remove(); }
     }
 
     public void addInput(String input) {
@@ -90,25 +133,23 @@ public class QueuePollingReader extends Reader {
     public int read(char[] buf, int off, int len) throws IOException {
         if (len == 0) return 0;
 
-        if (newlineSkippingMode) {
+        if (newlineSkippingMode.get()) {
             skipLeadingNewlines();
             return -1;
         }
 
         // First char taken from an empty queue will cause `needs-input` message
         // to be sent to the client.
-        int firstChar = readInputChar(true);
+        int firstChar = readInputChar();
+        if (firstChar == -2) {
+            readEof();
+            return -1;
+        }
         if (firstChar < 0) return -1;
         buf[off] = (char)firstChar;
 
-        // For subsuquent read chars, only poll the queue for data already
-        // there, and when queue becomes empty, return the number of chars read.
-        int i = 1;
-        while (i < len) {
-            int c = readInputChar(false);
-            if (c < 0) break;
-            buf[off + i++] = (char)c;
-        }
-        return i;
+        // For subsequent read chars, only take the data that is already there,
+        // and when it runs out, return the number of chars read.
+        return readBufferedChars(buf, off, len, 1);
     }
 }
```

alexander-yakushev:
Here's another attempt.

There indeed was another race between the actual consumer thread the the newline-cleanup code that was run by the handler thread. It is possible to protect from races between them but I'm reluctant to do so because, in the end, there really should only be a single consumer thread for the reader – you can't be reading a stream with multiple consumers and expect anything coherent, regardless of how much "race-safe" the code is.

- So, I've rewritten the skipping code again. Now, it relies on the `ready()` method to make sure we don't block on reading the next character. `ready()` is implemented by QueuePollingReader and is already supported by the outer PushbackReader.
- The call to skip the newline has moved to `interruptible-eval` middleware. While it muddles the separation of concerns a bit, it ensures that the reader and the skipper are the same thread. Skipping now happens on the "session" (eval) thread instead of the handler thread. It also now happens after the eval like it should, not before the next eval which was quite confusing.
- Implementing `ready()` allowed to get rid of newline-skipping dynvars and flags altogether.

PR #474, opened 2026-09-09
Keep the session alive when evaluated code closes *in* or *out*

bbatsov:
After #470 moved the newline-skipping hack into the evaluator's `finally`, `(slurp *in*)` or a `set!` of `*in*` to a plain `BufferedReader` made that finally throw, which killed the session thread: no `done`, and every later eval on the session hung. A closed `*out*` did the same via `flush`, and that one predates 1.7.0.

The cleanup is now best-effort, and the newline skip only runs on the `LineNumberingPushbackReader` a session binds `*in*` to. Closing `*in*` still leaves it closed afterwards, same as in `clojure.main`; the point is that the session keeps answering.


alexander-yakushev (approved):
Makes sense, even though binding *in* to something else than LNPR will cause other issues, so it's a defense against PEBKAC, but other clauses are mor plausible.

PR #479, opened 2026-09-10
Pin the CI toolchain and fix the deploy tag filters

bbatsov:
Resolving the classpath dies now and then with a ClassCastException out of HashMap: Maven's model validator caches the coordinates it has validated in an unsynchronized HashSet, and tools.deps resolves on several threads at once. It hit the Clojure 1.11 job on JDK 26 on #451 and on JDK 8 on #481, both with a restored ~/.m2, so it's about .cpcache missing and the classpath being recomputed, not the JDK or the network. Dependencies are now resolved in a retried step before the tests run. While looking at that: the cache key never included build/main.clj, because `find -name` doesn't match paths, so that's fixed too.

Clojars deploys skipped the -beta and -rc tags that the GitHub release workflow treats as prereleases (v1.6.0-beta1 never reached Clojars), and snapshot tags deployed without running any tests. One deploy job now handles all of them, gated on the full matrix, so a vX.Y.Z-SNAPSHOT tag is a real dry run of the release path.

The Windows job ran on whatever JDK the orb image shipped, with no dependency cache. It now installs Temurin 21, caches ~/.m2, and stores its test results. That's the part I couldn't try locally, so its run on this PR is the verification.


doc/modules/ROOT/pages/hacking_on_nrepl.adoc
```diff
@@ -9,9 +9,8 @@ interactive Lisp manner and just make some changes and test them directly from a
 
 == Building nREPL
 
-NOTE: Releases are available from Clojars, and SNAPSHOT builds from master's
-HEAD are automatically deployed there as well, so manually building
-nREPL shouldn't ever be necessary (unless you're hacking on it).
+NOTE: Releases are available from Clojars, so manually building nREPL
+shouldn't ever be necessary (unless you're hacking on it).
```
alexander-yakushev: This is such a weird paragraph that says absolutely nothing. I suggest deleting it altogether.

alexander-yakushev (changes requested):
I don't feel like pinning tools.deps version is necessary. It's just adding points for senseless bump churn.

The other changes are welcome.

doc/modules/ROOT/pages/hacking_on_nrepl.adoc
bbatsov: I guess it seemed useful at some point, but I have forgotten the reason for that. 😂

bbatsov:
PR updated.

alexander-yakushev (approved):

